PR #2, opened 2020-02-04
CI migration

dbkeitel:
This should replicate the gitlab CI on github. It will probably fail because `black` somehow is stricter here than it was over at AEI (probably a version question), but I'd prefer to fix those failures in a separate PR after.

However, this PR is just for testing; I suspect one probably cannot actually add a CI "action" through a PR and if nothing seems to be happening, I'll cancel it and add the action + config file manually once @GregoryAshton gives me access.


dbkeitel:
...yeah, doesn't do anything.

PR #3, opened 2020-02-05
CI migration

dbkeitel:
This should replicate the gitlab CI on github. It will probably fail because black somehow is stricter here than it was over at AEI (probably a version question), but I'd prefer to fix those failures in a separate PR after.


dbkeitel:
Ok good, after dropping python 3.5 (no black available) it runs almost through, just the expected style fails that I will fix separately.

3.8 also failed with some weird errors earlier on my test repo, will need to investigate that later.

PR #16, opened 2020-02-05
use os.path.join instead of manual string joining with "/"

dbkeitel:
Found this old branch lying around. I think I started this in the suspicion that it would fix some weird failures with ephemerides finding on a non-standard platform, which turned out to have another source in the end. But it's still a good idea for portability. So if black is ok with it, we should merge it in.


dbkeitel:
@GregoryAshton : black is now happy, but not all lines of the diffs are 100% trivial, so a quick look would be appreciated.

PR #17, opened 2020-02-06
add manual ephemerides options to search classes

### LINKED ISSUE #13: add manual ephemerides options to search classes
from https://gitlab.aei.uni-hannover.de/GregAshton/PyFstat/issues/21 (20191106):

> [...] the Readme was claiming that one can provide ephemerides manually through a search object's arguments. However this doesn't seem to be true: BaseSearchClass has a set_ephemeris_files() method, but it's called without arguments in init(), which itself doesn't take any ephemerides arguments.

dbkeitel:
-fixes #13

I've added the new input variables to ComputeFstat and all derived classes in core.py, grid_based_searches.py and mcmc_based_searches.py. But I'm somewhat confused about the different ways they do their inits, some explicitly setting self.xxx=xxx and some not, etc. So this might not all be sane, even though tests seem to all pass, but I know we don't have full coverage.

I should also add at least one test that explicitly uses the manual options.

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -203,10 +203,13 @@ def predict_fstat(
     cl_pfs.append("--maxStartTime={}".format(int(maxStartTime)))
     cl_pfs.append("--outputFstat={}".format(tempory_filename))
 
-    if earth_ephem is not None:
-        cl_pfs.append("--ephemEarth='{}'".format(earth_ephem))
-    if sun_ephem is not None:
-        cl_pfs.append("--ephemSun='{}'".format(sun_ephem))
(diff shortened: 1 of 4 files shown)


dbkeitel:
I've added one simple test case. @GregoryAshton please have a first look and let's decide if more are needed.

dbkeitel:
I now understand better the way that different classes inherit from each other and the updated PR should be more robust than previously.

dbkeitel:
Passes all tests and the relevant examples at least do not crash and produce reasonable output by eye (except for semi_coherent_directed_follow_up.py which seems broken for unrelated reasons), so I'll merge this now. Any potential untested breakages we'll need to address afterwards.

Issue #7, opened 2020-02-05
add header to output files


dbkeitel:
old issue from gitlab, see https://gitlab.aei.uni-hannover.de/GregAshton/PyFstat/issues/10

my last post from 2017122:

For the reference, Greg's example of current output from gitlab MR 9:
```
# date:2017-12-20 11:54:51.736768;user:greg;hostname:greg-len-ubuntu
# _ _ F0 F1 F2 Alpha Delta
1.000000000000000000e+09 1.000172800000000000e+09 2.999998991056105524e+01 -1.000000000000000036e-10 0.000000000000000000e+00 5.000000000000000000e-01 1.000000000000000000e+00 4.206904220581054688e+01 1.000097200000000000e+09 1.980000000000000000e+04
1.000000000000000000e+09 1.000172800000000000e+09 2.999999011234983470e+01 -1.000000000000000036e-10 0.000000000000000000e+00 5.000000000000000000e-01 1.000000000000000000e+00 4.204999923706054688e+01 1.000097200000000000e+09 1.980000000000000000e+04
```

Suggestions:
* [x]  separate date,user,host into extra lines
* [x] remove "_ _" from columns line (I think these are some internals from self.keys)
* [x] include PyFstat git hash
* [x] include lalpulsar git hash
* [ ] include some sort of configuration dump (like the CLAs in LALApps output files, here I guess we'd need to subselect some of the class attributes manually (to avoid dumping actual data), or maybe save a string to self.config at init time?

dbkeitel:
Working on this now, at least for grid searches, on [this branch](https://github.com/PyFstat/PyFstat/tree/output-file-headers).

dbkeitel:
>  remove "_ _" from columns line (I think these are some internals from self.keys)

Actually the corresponding output data are minStartTime and maxStartTime and I didn't see any clear reason why they needed to be replaced with "_", so I'm now naming them explicitly.

dbkeitel:
> > remove "_ _" from columns line (I think these are some internals from self.keys)
> 
> Actually the corresponding output data are minStartTime and maxStartTime and I didn't see any clear reason why they needed to be replaced with "_", so I'm now naming them explicitly.

No, actually this is more difficult than I thought, as some of the derived classes use their `keys` quite differently...

dbkeitel:
With #23 this should mostly be resolved satisfactorily for `GridSearch` and its derivatives. I left open
[...]

PR #23, opened 2020-03-05
improve grid search output file headers

dbkeitel:
refs #7 (but doesn't close it, as MCMC not yet addressed)

### DIFF
--- a/examples/grid_examples/grid_search_F0.py
+++ b/examples/grid_examples/grid_search_F0.py
@@ -10,14 +10,16 @@ Alpha = 1.0
 Delta = 1.5
 
 # Properties of the GW data
-sqrtSX = 1e-23
+depth = 70
+sqrtS = "1e-23"
+h0 = float(sqrtS) / depth
+IFOs = "H1"
+# IFOs = "H1,L1"
+sqrtSX = ",".join(np.repeat(sqrtS, len(IFOs.split(","))))
 tstart = 1000000000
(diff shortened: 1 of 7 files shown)


dbkeitel:
This now includes headers of the type
```
# date: 2020-03-09 16:40:28.693842
# user: dkeitel
# hostname: [...]
# PyFstat: UNCLEAN 2a8d0dc 2020-03-09 16:06:42 +0100
# LAL: 6.21.0 (CLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# LALPulsar: 1.18.2 (CLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# minStartTime maxStartTime F0 F1 F2 Alpha Delta twoF
```
for the standard GridSearch class output, and similar (just last line with the column names changed) for the extra F_mn and atoms outputs TransientGridSearch.

GridGlitchSearch and FrequencySlidingWindow have also been updated, at least as far as was necessary to make their unit tests pass.

Note the internal changes to search parameter storage (`self.keys` and related fields);  and that I've promoted the header construction function to `BaseSearchClass.get_output_file_header()` so it can be re-used for MCMC searches next.  
(At first I thought to put it at the `ComputeFstat` class level, but since `BaseSearchClass` already contains ephemerides-related stuff, it's not lal-independent either, and currently seems the logical level to start collecting version information and prepare it for later output.)

Issue #32, opened 2020-04-07
MCMCFollowUpSearch: do not try to reproduce walker plot when loading existing data


dbkeitel:
If called with `return_fig=True` as e.g. in `examples/followup_examples/semi_coherent_directed_follow_up.py` this leads to an empty plot.

PR #33, opened 2020-04-07
MCMC: improve handling of walker plots

dbkeitel:
@Rodrigo-Tenorio This should solve the problem in #32 that you had previously spotted when re-running the `examples/followup_examples/semi_coherent_directed_follow_up.py` case.

I decided to do a **backwards-incompatible** change to the arguments of both `MCMCFollowUpSearch.run()` and `MCMCSearch.run()`:

- rename `create_plots` option to `plot_walkers` for clarity
- replace `kwargs` pass-through to `_plot_walkers()` with explicit `walker_plot_args` dictionary
- in `MCMCFollowUpSearch.run()` don't return `fig`, `axes` but store as `self.walker_fig`, `self.walker_axes`
- introduce the same feature to the base `MCMCSearch.run()`
- adapt the followup example script to explicitly check if fig and axes were stored; if not, it's likely because old data was used, and it won't overwrite the figure. This would be the intended usage in general.

Please have a look and test.

cc @GregoryAshton

### DIFF
--- a/examples/followup_examples/semi_coherent_directed_follow_up.py
+++ b/examples/followup_examples/semi_coherent_directed_follow_up.py
@@ -79,19 +79,24 @@ fig, axes = plt.subplots(nrows=2, figsize=(3.4, 3.5))
 mcmc.run(
     NstarMax=NstarMax,
     Nsegs0=Nsegs0,
-    labelpad=0.01,
-    plot_det_stat=False,
-    return_fig=True,
-    fig=fig,
-    axes=axes,
+    walker_plot_args={
+        "labelpad": 0.01,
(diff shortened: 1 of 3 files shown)


Rodrigo-Tenorio:
Tested using follow up for isolated and binary CW. No changes other than changing empty dictionaries to None as default (note that X or Y **returns the first truthy value encountered** starting form left hand side).

Side note: One has to be careful to allocate as many axes as searched parameters; otherwise, the program will not stop but just warn.

dbkeitel:
Looks like you hit the "close" button instead of merge, I'll do that now.

PR #34, opened 2020-04-07
CI: try python 3.8 once more

dbkeitel:


dbkeitel:
Cool, this actually works without any changes! 3.8 gave weird errors when I first set up the CI, but not any more apparently.

PR #38, opened 2020-04-10
Prune injection_helper_functions

Rodrigo-Tenorio:
After 134ca45fd7c98e4c1904474f61004b7d9d56a099 I've noticed that injection_helper_functions contains sort of a duplication of helper_functions.py/get_covering_band.

The main difference is that the lalpulsar implementations assumes the worst sky position for this computation, while the one in PyFstat takes into account a particular sky position. This may become significant depending on the computational cost of the method and the frequency range being analyzed.

However, it also "duplicates" a function using non-lal tools and grep shows no usage of this module at all, which means that we could even get rid of it for the sake of simplicity and explicitly use astropy whenever a change of coordinates is needed.


dbkeitel:
Actually, `get_covering_band()` doesn't seem to be in use either...

@GregoryAshton Do you remember what was the intended use of the homebrew `get_frequency_range_of_signal()` and the whole `injection_helper_functions.py` module? I agree with @Rodrigo-Tenorio that this looks like it could just be removed at this point.

dbkeitel:
@GregoryAshton I think this really needs your decision. As Rodrigo points out, this mini-module is not used anywhere else in the package. But if you have/had a workflow where you used to call it directly to design your commandlines, it can also be kept. Please let us know.

dbkeitel:
@Rodrigo-Tenorio As per #40 I think we want to close this and keep the module after all?

GregoryAshton:
@Rodrigo-Tenorio and @dbkeitel my apologies for not having seen this before. I have no dependencies which rely on pyfstat so please feel free to make changes which clean up the module like this :)

dbkeitel:
@Rodrigo-Tenorio so do you think these functions would still be useful to you in some circumstances? If yes, let's keep them (close the PR). If not, we can prune them now (I merge the PR) and they're still recoverable from git history if ever needed.

Rodrigo-Tenorio:
@dbkeitel merge the PR!

Those functions are either a wraparound Astropy, which is simple enough to be stated clearly every time it's used, 
or a duplication of a lalpulsar function specifically taylored to isolated CW.

Both things made sense in the past before adapting PyFstat to Binary Modulated Signals, but now I'd say the best thing to do is to stick to the actual libraries without any wrappers to avoid unfortunate mistakes.

dbkeitel:
Ok, let's remove them, and if there is a need again, they're still there in the history.

Issue #37, opened 2020-04-08
improve binary support, add test cases / examples


dbkeitel:
@Rodrigo-Tenorio continuing from #29:

> Also, I see that binary parameters are not supported for injection using this code... Since injections using noise SFT seem not to be supported (am I right?), there is need to focus on this issue by now.

Can you please open separate issues/PRs for any parts of the code where explicit binary support needs to be added, after this one is merged?

_Originally posted by @dbkeitel in https://github.com/PyFstat/PyFstat/pull/29#issuecomment-610361725_

Just to keep better track, I'm opening this one for now, and you can add more details if necessary. It would also be great to have some test cases and/or examples for binaries. Maybe a simple (and tiny!) grid search as a test case, and a simple one-stage follow-up example?

Rodrigo-Tenorio:
Hi @dbkeitel , seems I made a mistake in comment, it should read "there is **no**need to focus on this issue by now". My bad.

What I meant is that PyFstat does not support the addition of noise via SFTs to the injections.
Hence, the only reason we would use this functionality would be to test a toy follow up for binary signals.

However, since PyFstat uses makefakedata, I guess it would not be that difficult to add this functionality and allow follow ups to be done using this tool only?

I'll start with this and think about it as I go.

dbkeitel:
The injections / SFT generation part is actually one that should ideally be redone, since it currently calls MFD executables, while it would be cleaner to call the underlying functions directly through SWIG. However, none of us have time for it, and as you say, it should also be easy to add support for missing features through the current interface.

Rodrigo-Tenorio:
Hi, started to work on this in abbd9f60a6f16b1ef36c5c6fd54c89af1b232b5d.
As you say, `make_sft.py` is screaming for a total rework (too many hard coded arguments in too many ways).

Bright side is injection parameters as given to mfd are read from a file rather than  formatted from object attributes, so that was the only part I had to accommodate for.

dbkeitel:
Just for the record, keeping this open until #39 is merged.

PR #39, opened 2020-04-10
Binary injections and examples

Rodrigo-Tenorio:
Work on #37 

To do:

- [x] Implement binary injections
- [x] Add binary search example
- [x] Add binary follow up example
- [x] Add binary writer test


dbkeitel:
@Rodrigo-Tenorio This looks good at first glance, but I'll have to take another closer look at the writer class before merging, just to be on the safe side. Two questions meanwhile:

1.  For the search/followup itself, as called from the new examples, the key binary support is already [here](https://github.com/PyFstat/PyFstat/blob/master/pyfstat/mcmc_based_searches.py#L261), right?
1. Could you add a simple test for `BinaryModulatedWriter`, similar to [the one for the base class](https://github.com/PyFstat/PyFstat/blob/master/tests.py#L61)? For the moment, it doesn't have to actually verify the result in much detail, just check that it runs in principle.

Rodrigo-Tenorio:
@dbkeitel 

1. Yes, those classes already supported binary parameters, so I just had to extend the isolated examples adding more parameters.

Rodrigo-Tenorio:
2. Just implemented it at 3a6b54e.

Since BinaryModulatedWriter inherits from Writer to add binary parameters into the format and nothing more, I though it would make sense to use the same tests through inheritance. 

I could understand that we wanted fully fleshed out tests even if that implied code duplication,
but inheritance seemed easier and clearer.

pyfstat/make_sfts.py
```diff
@@ -413,6 +413,170 @@ def predict_fstat(self):
         return twoF_expected
 
 
+class BinaryModulatedWriter(Writer):
+    """ Instance object for generating SFTs containing binary continuous signals """
```
dbkeitel, line 417:
```suggestion
    """ Instance object for generating SFTs containing a continuous wave signal for a source in a binary system """
```

pyfstat/make_sfts.py
```diff
@@ -413,6 +413,170 @@ def predict_fstat(self):
...
+            maxStartTime,
+            add_noise,
+            transientWindowType,
+        )
+
+        self.parse_args_consistent_with_mfd()
+
+    def parse_args_consistent_with_mfd(self):
+        """ This will allow us to get rid of the get_single_config_* family
+            Future rework may use a dictionary as the default input method,
```
dbkeitel, line 498: This interface looks indeed a bit nicer than the one on the original Writer class! So eventually one could merge the two based on this new one. But for now I'm fine adding this one and leaving the original as is.

dbkeitel:
>     1. Just implemented it at [3a6b54e](https://github.com/PyFstat/PyFstat/commit/3a6b54eb47ae4e1db13d0f4501f9fbdb3ef084f5).
> 
> 
> Since BinaryModulatedWriter inherits from Writer to add binary parameters into the format and nothing more, I though it would make sense to use the same tests through inheritance.
> 
> I could understand that we wanted fully fleshed out tests even if that implied code duplication,
> but inheritance seemed easier and clearer.

That's in fact a very nice solution. :+1:

dbkeitel:
@Rodrigo-Tenorio apologies for the extreme delay. The changes all look sane. I've only improved one doc string. I'm now running the examples locally, then will merge.

dbkeitel:
@Rodrigo-Tenorio I found one issue while running `examples/binary_examples/semi_coherent_directed_binary_follow_up.py`: the separate CFSv2 run used to produce a `.loudest` file only recovers
```
twoF     =  860.017;
```
against the `Max twoF: 829758.125` reported by the MCMC run.

It simply does not seem to have passed the binary parameters to that call at all. Could you have a look at modifying the `generate_loudest` method please?

Rodrigo-Tenorio:
Hi @dbkeitel , 

Yes, you are right, isolated parameters where hard-coded in the loudest call.
Plus, I realized I made a mistake when coding the binary parameters into the writer class:
This package uses F0, F1, F2,..., asini, period, tp, ... while lal uses Freq, f1dot, f2dot, ... orbitasini, orbitPeriod, orbitTp... (as you can see, lal is not consistent either, since orbitasini is the only one not following the camelCase convention). 

I fixed both things: 
1) I wrote a function to convert dictionary keys from one convention to the other one (bca5877)
Additionally this makes BinaryModulatedWriter cleaner: https://github.com/PyFstat/PyFstat/blob/bca5877bc590dd013ec7c09cf050f9a56beac0df/pyfstat/make_sfts.py#L537)

2) theta_prior labels are used to define CFS_v2's input arguments (6e3e249).

dbkeitel:
The translation function is great!

dbkeitel:
Hmm, something's still off with the "loudest" run. It's only getting `twoF     =  22186.6`. Comparing 
[semi_coherent_directed_binary_follow_up.loudest](https://github.com/PyFstat/PyFstat/files/4700412/semi_coherent_directed_binary_follow_up.loudest.txt) with the following log output, it looks like it may be using the median of each parameter instead of the max2F parameter space point...? 

```
09:50 INFO    : Max twoF: 974152.6875 with parameters:
  Alpha      = 1.472514790e-01
  Delta      = 1.542256910e-01
  F0         = 2.999999754e+01
  argp       = 2.783693603e+00
  asini      = 9.974532501e+00
  ecc        = 3.104122926e-03
  period     = 3.889161762e+06
  tp         = 1.009140515e+09
09:50 INFO    : Median +/- std for production values
09:50 INFO    :   Alpha      = 1.477695848e-01 +/- 5.237046311e-03
09:50 INFO    :   Delta      = 1.520400306e-01 +/- 7.615368763e-03
09:50 INFO    :   F0         = 2.999999999e+01 +/- 1.808604016e-06
09:50 INFO    :   argp       = 3.106845665e+00 +/- 5.834129194e-01
09:50 INFO    :   asini      = 9.995133632e+00 +/- 2.756873938e-01
09:50 INFO    :   ecc        = 4.678664477e-03 +/- 1.089193492e-03
09:50 INFO    :   period     = 3.893789793e+06 +/- 4.403461183e+04
09:50 INFO    :   tp         = 1.001720195e+09 +/- 1.715429328e+07
```

Rodrigo-Tenorio:
Hi, 

After running 770d2c7 I get `twoF     =  8.1816e+06` for the loudest candidate, not far away from the output:
```
15:49 INFO    : Max twoF: 11509770.75 with parameters:
  asini      = 9.999993564e+00
  period     = 3.887999885e+06
  tp         = 1.003536001e+09
15:49 INFO    : Median +/- std for production values
15:49 INFO    :   asini      = 9.999995640e+00 +/- 1.272776068e-02
15:49 INFO    :   period     = 3.888000084e+06 +/- 1.283432688e+02
15:49 INFO    :   tp         = 1.003536001e+09 +/- 4.445095798e+04
```
As you say, it seems it's taking the median value instead of the max 2F parameter space point; however, examples/mcmc_examples/semicoherent_search_using_MCMC.py does the same thing.

dbkeitel:
I've looked into the code and what is going on is simply that the `write_par()` method is called with method="med" by default when used as input for the "loudest" CFSv2 call. I'll merge this now and make a separate PR to store both the max and median parameters in separate files, then make the "loudest" call at the max instead.

Issue #40, opened 2020-04-10
Enhance injection options


Rodrigo-Tenorio:
Actually, `get_covering_band()` doesn't seem to be in use either...

@GregoryAshton Do you remember what was the intended use of the homebrew `get_frequency_range_of_signal()` and the whole `injection_helper_functions.py` module? I agree with @Rodrigo-Tenorio that this looks like it could just be removed at this point.

_Originally posted by @dbkeitel in https://github.com/PyFstat/PyFstat/pull/38#issuecomment-612122286_

I'd keep `get_frequency_range_of_signal()`. I myself have coded some similar functions whenever I had to do injections and this one will come handy.

I would let the user decide whether they want to use it or on the other hand they want to write a particular band by themselves (makefakedata_v5 will complain if the band is not wide enough).

On the other hand Makefakedata's `--noiseSFTs` could  be implemented to allow for injections using SFTs noise rather than just gaussian noise.

dbkeitel:
In #49 I now have repurposed `get_covering_band()` for internal use, so that one will definitely stay.

dbkeitel:
Remaining changes will be done in #41.

dbkeitel:
Effectively done with #41, just need to decide on #38, but no need for this issue any more.

PR #41, opened 2020-04-10
implement noiseSFTs option for Writer

Rodrigo-Tenorio:
Ref. #40 

- [x] Implement noise SFTs for Writer and sons
- [x] Implement test

P. S.
@dbkeitel I'm not sure if this is the right way to do this: This builds upon #39. After getting #39 merged into master, I'll properly rebase this one on top of master from the point of divergence.

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -45,10 +45,10 @@ class Writer(BaseSearchClass):
         phi=0,
         Tsft=1800,
         outdir=".",
-        sqrtSX=1,
+        sqrtSX=1.,
         noiseSFTs=None,
-        SFTWindowType="tukey",
-        SFTWindowBeta=0.01,
+        SFTWindowType="rectangular",
+        SFTWindowBeta=0.0,
         Band=4,
         detectors="H1",
         minStartTime=None,
@@ -374,14 +374,13 @@ transientTau = {:10.0f}\n"""
         cl_mfd.append('--outSFTdir="{}"'.format(self.outdir))
         cl_mfd.append('--outLabel="{}"'.format(self.label))
         if self.add_noise:
-            if self.noiseSFTs is not None:
-                logging.warning(
-                    "noiseSFTs option takes preference over sqrtSX (i.e. sqrtSX option is *ignored*)."
-                )
-                cl_mfd.append('--noiseSFTs="{}"'.format(self.noiseSFTs))
-            else:
+            cl_mfd.append('--noiseSFTs="{}"'.format(self.noiseSFTs))
+
+            if self.noiseSFTs is None:
                 cl_mfd.append("--IFOs={}".format(self.IFOs))
+            if self.sqrtSX is not None:
                 cl_mfd.append('--sqrtSX="{}"'.format(self.sqrtSX))
+
             cl_mfd.append('--SFTWindowType="{}"'.format(self.SFTWindowType))
             cl_mfd.append("--SFTWindowBeta={}".format(self.SFTWindowBeta))
         else:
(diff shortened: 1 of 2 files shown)


dbkeitel:
Will need a manual rebase now.

pyfstat/make_sfts.py
```diff
@@ -370,9 +373,33 @@ def run_makefakedata(self):
...
         cl_mfd.append('--outLabel="{}"'.format(self.label))
-        cl_mfd.append("--IFOs={}".format(self.IFOs))
+
         if self.add_noise:
+
+            if self.noiseSFTs is not None and self.SFTWindowType is None:
+                raise ValueError(
+                    "SFTWindowType is required when using noiseSFTs. "
+                    "Please, make sure you understand the window function used "
+                    "to produced noiseSFTs."
```
dbkeitel, line 383: minor typo: produced->produce

pyfstat/make_sfts.py
```diff
@@ -370,9 +373,33 @@ def run_makefakedata(self):
         cl_mfd.append("--outSingleSFT=TRUE")
         cl_mfd.append('--outSFTdir="{}"'.format(self.outdir))
         cl_mfd.append('--outLabel="{}"'.format(self.label))
-        cl_mfd.append("--IFOs={}".format(self.IFOs))
+
         if self.add_noise:
+
```
dbkeitel, line 378: could you remove this blank line and 389 too so the block hierarchy stays cleaner

pyfstat/make_sfts.py
```diff
@@ -370,9 +373,33 @@ def run_makefakedata(self):
...
+            elif self.noiseSFTs is not None:
+                cl_mfd.append('--noiseSFTs="{}"'.format(self.noiseSFTs))
+            else:
+                cl_mfd.append("--IFOs={}".format(self.IFOs))
+
+            logging.warning(
+                "Gaussian noise *is added* by default due to "
+                "consistency issues. Please, make sure sqrtSX value is "
+                "consistent with what you intend to do (0=No Gaussian noise). "
+                "Currently, sqrtSX={}".format(self.sqrtSX)
```
dbkeitel, line 394: That whole situation may require some further thought, I'm not so sure if the separate `add_noise` flag really makes sense either. But for now it's a good idea to put in that warning and if I end up unhappy with the sqrtSX situation, I'll make a separate issue/PR. Thanks.

dbkeitel:
For the record, the lalsuite fix is here: https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1354
But we decided to implement our own safety check on top of it, to be also safe with old versions. (New LALSuite release should be imminent, but can still be delayed.)

Rodrigo-Tenorio:
For the sake of completeness I'll give some clarification on 20acefb:

- Makefakedata_v5's kerfuffle
  - `SFTWindowType` was supposed to be required (and it actually was at _v4) but somehow such requirement flow over everyone's head and wasn't being enforced *at all*.
  - PyFstat has been updated to enforce this requirement. It is useful to make the user aware of the choices being done.

- The issue with sqrtSX
  - sqrtSX is being (wrongly) used as two different things at different points of this code: As a way of specifying Gaussian noise at MFD and as a PSD measure at ComputeFstat. 
  - This is problematic, since *no Gaussian noise* is represented as sqrtSX=0 in MFD, but as sqrtSX=1 in ComputeFstat.
  - Short term solution: Warn the user to set sqrtSX=0 when creating SFTs.

- Test
  - Improved naming conventions prevents name clashing between different stages.

pyfstat/make_sfts.py
Rodrigo-Tenorio: I'm not sure how many things will break down when removing `add_noise`. Agree with doing a separate issue/PR.

dbkeitel: see #50

PR #45, opened 2020-06-03
dump search class init arguments to a text file

dbkeitel:
-add `_dump_init_params()` method to `BaseSearchClass`
 -call it from the `__init__()` of all high-level classes
 -closes #35

The idea is to make runs more reproducible.

@GregoryAshton @ReinhardPrix @Rodrigo-Tenorio how does this look?

example output from the fully_coherent_search_using_MCMC.py example:
```
MCMCSearch
theta_prior = {'F0': {'type': 'unif', 'lower': 29.99999995, 'upper': 30.00000005}, 'F1': {'type': 'unif', 'lower': -1.0005e-10, 'upper': -9.995e-11}, 'F2': 0, 'Alpha': 1.4596048908088417, 'Delta': 0.38422376285103965}
tref = 1004320000.0
label = fully_coherent_search_using_MCMC
outdir = example_data/fully_coherent_search_using_MCMC
minStartTime = 1000000000
maxStartTime = 1008640000
sftfilepattern = example_data/fully_coherent_search_using_MCMC/*fully_coherent_search_using_MCMC*sft
detectors = None
nsteps = [300, 300]
nwalkers = 100
ntemps = 2
log10beta_min = -0.5
theta_initial = None
rhohatmax = 1000
binary = False
BSGL = False
SSBprec = None
minCoverFreq = None
maxCoverFreq = None
injectSources = None
assumeSqrtSX = None
transientWindowType = None
tCWFstatMapVersion = lal
earth_ephem = None
sun_ephem = None
```

Issue #35, opened 2020-04-07
dump configuration of search objects?


dbkeitel:
extracted as a more difficult feature from #7, hence optional for future improvement:

> include some sort of configuration dump (like the CLAs in LALApps output files, here I guess we'd need to subselect some of the class attributes manually (to avoid dumping actual data), or maybe save a string to self.config at init time?

This could go either into the ASCII data output files or into a separate file, but might be a tricky to do in a universal way.

For now, it's at least already possible to work with the pickles.

PR #46, opened 2020-06-03
output file headers: include search class name and init arguments dict

dbkeitel:
As a solution to #35, actually I think this is better than #45. Not a separate dump file, but adding the parameter dict to the existing header of each output file. Similar to the lalapps convention.

@GregoryAshton @ReinhardPrix @Rodrigo-Tenorio how does this look?

example output from the fully_coherent_search_using_MCMC.py example:
```
# date: 2020-06-03 11:48:41.727082
# user: dkeitel
# hostname: T490UIB
# PyFstat: UNCLEAN 16b9dda 2020-06-03 11:27:38 +0200
# search: MCMCSearch
# parameters: {'theta_prior': {'F0': {'type': 'unif', 'lower': 29.99999995, 'upper': 30.00000005}, 'F1': {'type': 'unif', 'lower': -1.0005e-10, 'upper': -9.995e-11}, 'F2': 0, 'Alpha': 1.4596048908088417, 'Delta': 0.38422376285103965}, 'tref': 1004320000.0, 'label': 'fully_coherent_search_using_MCMC', 'outdir': 'example_data/fully_coherent_search_using_MCMC', 'minStartTime': 1000000000, 'maxStartTime': 1008640000, 'sftfilepattern': 'example_data/fully_coherent_search_using_MCMC/*fully_coherent_search_using_MCMC*sft', 'detectors': None, 'nsteps': [10, 10], 'nwalkers': 100, 'ntemps': 2, 'log10beta_min': -0.5, 'theta_initial': None, 'rhohatmax': 1000, 'binary': False, 'BSGL': False, 'SSBprec': None, 'minCoverFreq': None, 'maxCoverFreq': None, 'injectSources': None, 'assumeSqrtSX': None, 'transientWindowType': None, 'tCWFstatMapVersion': 'lal', 'earth_ephem': None, 'sun_ephem': None}
# LAL: 6.21.0 (CLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# LALPulsar: 1.18.2 (CLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# F0 F1
30.00000000356636 -1.00003001670372e-10
[...]
```

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -431,6 +431,13 @@ class BaseSearchClass(object):
         pretty_init_parameters = pformat(
             self.init_params_dict, indent=2, width=74
         ).split("\n")
+        pretty_init_parameters = (
+            ["{"]
+            + [pretty_init_parameters[0].replace("{", " ")]
+            + pretty_init_parameters[1:-2]
+            + [pretty_init_parameters[-1].rstrip("}")]
+            + ["}"]
+        )
 
         header = [
             "date: {}".format(str(datetime.now())),
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -4,6 +4,7 @@
 import os
 import logging
 import copy
+from pprint import pformat
 
 import glob
 import numpy as np
@@ -426,14 +427,19 @@ class BaseSearchClass(object):
         self.init_params_dict = argsdict
 
     def get_output_file_header(self):
+
+        pretty_init_parameters = pformat(
+            self.init_params_dict, indent=2, sort_dicts=True, width=74
+        ).split("\n")
+
         header = [
             "date: {}".format(str(datetime.now())),
             "user: {}".format(getpass.getuser()),
             "hostname: {}".format(socket.gethostname()),
             "PyFstat: {}".format(helper_functions.get_version_string()),
             "search: {}".format(type(self).__name__),
-            "parameters: {}".format(self.init_params_dict),
-        ]
+            "parameters: ",
+        ] + pretty_init_parameters
         lalVCSinfo = lal.VCSInfoString(lalpulsar.PulsarVCSInfoList, 0, "")
         header += filter(None, lalVCSinfo.split("\n"))
         return header
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -437,11 +437,14 @@ class BaseSearchClass(object):
             "user: {}".format(getpass.getuser()),
             "hostname: {}".format(socket.gethostname()),
             "PyFstat: {}".format(helper_functions.get_version_string()),
-            "search: {}".format(type(self).__name__),
-            "parameters: ",
-        ] + pretty_init_parameters
+        ]
         lalVCSinfo = lal.VCSInfoString(lalpulsar.PulsarVCSInfoList, 0, "")
         header += filter(None, lalVCSinfo.split("\n"))
+        header += [
+            "search: {}".format(type(self).__name__),
+            "parameters: ",
+        ]
+        header += pretty_init_parameters
         return header
 
 
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -429,7 +429,7 @@ class BaseSearchClass(object):
     def get_output_file_header(self):
 
         pretty_init_parameters = pformat(
-            self.init_params_dict, indent=2, sort_dicts=True, width=74
+            self.init_params_dict, indent=2, width=74
         ).split("\n")
 
         header = [
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -421,12 +421,18 @@ class BaseSearchClass(object):
         else:
             self.sun_ephem = sun_ephem
 
+    def _set_init_params_dict(self, argsdict):
+        argsdict.pop("self")
+        self.init_params_dict = argsdict
+
     def get_output_file_header(self):
         header = [
             "date: {}".format(str(datetime.now())),
             "user: {}".format(getpass.getuser()),
             "hostname: {}".format(socket.gethostname()),
             "PyFstat: {}".format(helper_functions.get_version_string()),
+            "search: {}".format(type(self).__name__),
+            "parameters: {}".format(self.init_params_dict),
         ]
         lalVCSinfo = lal.VCSInfoString(lalpulsar.PulsarVCSInfoList, 0, "")
         header += filter(None, lalVCSinfo.split("\n"))
@@ -532,6 +538,7 @@ class ComputeFstat(BaseSearchClass):
 
         """
 
+        self._set_init_params_dict(locals())
         self.set_ephemeris_files(earth_ephem, sun_ephem)
         self.init_computefstatistic_single_point()
         self.output_file_header = self.get_output_file_header()
--- a/pyfstat/grid_based_searches.py
+++ b/pyfstat/grid_based_searches.py
@@ -95,6 +95,7 @@ class GridSearch(BaseSearchClass):
         with the `clean` option which uses a generator instead.
         """
 
+        self._set_init_params_dict(locals())
         if os.path.isdir(outdir) is False:
             os.mkdir(outdir)
         self.set_out_file()
@@ -796,6 +797,7 @@ class SliceGridSearch(GridSearch):
         For all other parameters, see `pyfstat.ComputeFStat` for details
         """
 
+        self._set_init_params_dict(locals())
         if os.path.isdir(outdir) is False:
             os.mkdir(outdir)
         self.set_out_file()
@@ -1019,6 +1021,7 @@ class GridGlitchSearch(GridSearch):
         For all other parameters, see pyfstat.ComputeFStat.
         """
 
+        self._set_init_params_dict(locals())
         self.BSGL = False
         self.input_arrays = False
         if tglitchs is None:
@@ -1114,6 +1117,7 @@ class SlidingWindow(GridSearch):
         For all other parameters, see `pyfstat.ComputeFStat` for details
         """
 
+        self._set_init_params_dict(locals())
         if os.path.isdir(outdir) is False:
             os.mkdir(outdir)
         self.set_out_file()
@@ -1251,6 +1255,7 @@ class FrequencySlidingWindow(GridSearch):
         For all other parameters, see `pyfstat.ComputeFStat` for details
         """
 
+        self._set_init_params_dict(locals())
         self.transientWindowType = "rect"
         self.nsegs = 1
         self.t0Band = None
@@ -1408,6 +1413,7 @@ class EarthTest(GridSearch):
 
         For all other parameters, see `pyfstat.ComputeFStat` for details
         """
+        self._set_init_params_dict(locals())
         self.transientWindowType = None
         self.t0Band = None
         self.tauBand = None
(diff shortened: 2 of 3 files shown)


dbkeitel:
@Rodrigo-Tenorio thanks for the pretty-print commit, that sounds like a good idea. However, the `sort_dicts` argument seems to be a python 3.8 novelty so we can't rely on that.

Also, will it still use proper comment prefixes?

Rodrigo-Tenorio:
@dbkeitel Nice, using `locals()`  

If I understand correctly, this is only a quality-of-live improvement to check the exact input producing the result (i.e. this is NOT intended to be parsed to reproduce calls, but just to keep track of input parameters).

I've improved the output format to something more readable (no lines longer than 80 characters!). Here an example:

```
# PyFstat: UNCLEAN ae6b064 2020-06-03 13:22:34 -0700
# search: MCMCSemiCoherentSearch
# parameters:
# { 'BSGL': False,
#   'SSBprec': None,
#   'assumeSqrtSX': None,
#   'binary': False,
#   'detectors': None,
#   'earth_ephem': None,
#   'injectSources': None,
#   'label': 'semi_coherent_search_using_MCMC',
#   'log10beta_min': -1,
#   'maxCoverFreq': None,
#   'maxStartTime': 1008640000,
#   'minCoverFreq': None,
#   'minStartTime': 1000000000,
#   'nsegs': 10,
#   'nsteps': [3, 3],
#   'ntemps': 1,
#   'nwalkers': 100,
#   'outdir': 'example_data/semi_coherent_search_using_MCMC',
#   'rhohatmax': 1000,
#   'sftfilepattern': 'example_data/semi_coherent_search_using_MCMC/*semi_coherent_search_using_MCMC*sft',
#   'sun_ephem': None,
#   'theta_initial': None,
#   'theta_prior': { 'Alpha': 1.4596048908088417,
#                    'Delta': 0.38422376285103965,
#                    'F0': { 'lower': 29.99999995,
#                            'type': 'unif',
#                            'upper': 30.00000005},
#                    'F1': { 'lower': -1.0005e-10,
#                            'type': 'unif',
#                            'upper': -9.995e-11},
#                    'F2': 0},
#   'tref': 1004320000.0}
# LAL: 6.21.0 (UNCLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# LALPulsar: 1.18.2 (UNCLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
```

dbkeitel:
Ok thanks, that example output answers my 2nd question. Looks good!

Do you know how strict it will be about the character limit - e.g. `sftfilepattern` can easily be much longer, and already is >100 chars in this example, so it seems for individual overlong strings it just leaves them as is? If so, that seems good.

dbkeitel:
> If I understand correctly, this is only a quality-of-live improvement to check the exact input producing the result (i.e. this is NOT intended to be parsed to reproduce calls, but just to keep track of input parameters).

Ideally that should be possible too, but one would need a custom parser to get rid of the comment markers etc I guess. The primary purpose is to make result files self-documenting so that years later one can still look at them and know what was run to produce them.

Rodrigo-Tenorio:
> Ok thanks, that example output answers my 2nd question. Looks good!
> 
> Do you know how strict it will be about the character limit - e.g. `sftfilepattern` can easily be much longer, and already is >100 chars in this example, so it seems for individual overlong strings it just leaves them as is? If so, that seems good.

Yes, `width` does not break lines if they can't be broken

dbkeitel:
Ok great. Do you have any more edits in the pipeline? If not, I'll resolve the conflict with master locally and also change the order of header lines (bringing the lal version info before this).

dbkeitel:
Updated example output:
```
# date: 2020-06-03 22:52:25.077915
# user: dkeitel
# hostname: T490UIB
# PyFstat: v1.3+115.621917db.dirty
# LAL: 6.21.0 (CLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# LALPulsar: 1.18.2 (CLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# search: MCMCSearch
# parameters: 
# { 'BSGL': False,
#   'SSBprec': None,
#   'assumeSqrtSX': None,
#   'binary': False,
#   'detectors': None,
#   'earth_ephem': None,
#   'injectSources': None,
#   'label': 'fully_coherent_search_using_MCMC',
#   'log10beta_min': -0.5,
#   'maxCoverFreq': None,
#   'maxStartTime': 1008640000,
#   'minCoverFreq': None,
#   'minStartTime': 1000000000,
#   'nsteps': [10, 10],
#   'ntemps': 2,
#   'nwalkers': 10,
#   'outdir': 'example_data/fully_coherent_search_using_MCMC',
#   'rhohatmax': 1000,
#   'sftfilepattern': 'example_data/fully_coherent_search_using_MCMC/*fully_coherent_search_using_MCMC*sft',
#   'sun_ephem': None,
#   'tCWFstatMapVersion': 'lal',
#   'theta_initial': None,
#   'theta_prior': { 'Alpha': 1.4596048908088417,
#                    'Delta': 0.38422376285103965,
#                    'F0': { 'lower': 29.99999995,
#                            'type': 'unif',
#                            'upper': 30.00000005},
#                    'F1': { 'lower': -1.0005e-10,
#                            'type': 'unif',
#                            'upper': -9.995e-11},
#                    'F2': 0},
#   'transientWindowType': None,
#   'tref': 1004320000.0}
# F0 F1
```

ReinhardPrix:
looks nice. Btw, just for curiosity, are there off-the-shelf tools to parse such commented dict-output back into python?

dbkeitel:
The pprint docs say the output is "in a form which can be used as input to the interpreter". The `#` comment markers are our own doing though, so I guess one has to read the header first and strip those, then pass the concatenated string to eval...? Haven't tried it.

Rodrigo-Tenorio:
We could come up with a hacky way of parsing those lines (and only those lines) from the output file (keep in mind the number of lines after `parameter` does not have to be fixed...). I guess the trick would be to reformat the output into
```
# paramters:
# {
# <key>: <value>
# }
```
and parse around the first and last curly bracket, but this approach misses the point somehow.

If we want this to be only a *stamp* in the sense of "Here are the input parameters used to run the search, in case you forgot after 6 months of not looking at this results", it's already done.

However, this information is "somehow" already encoded in the log file; hence I would propose 
to use a .json file with the parameters dictionary instead. Re-using input parameters is then as simple as reading that file and feeding it into the constructor using unpacking.

At first glance, the input parameter's dictionary only contains json serializable objects (meaning there are NO mandatory numpy arrays, everything is dict, list, or int, string, float)

ReinhardPrix:
sure, not to worry, I was just wondering on general grounds ...

dbkeitel:
Switching to .json or .h5 outputs would be a nice idea overall. However I'd prefer to keep changes minimal for now and draw up a roadmap for a PyFstat 2.0 with such big changes at some point - the biggest one would probably to wrap Bilby for flexible sampler choice and all its other more modern features.

The minor reformatting of putting the {} on separate lines might be a good idea however, I'd also find it slightly more human-readable.

dbkeitel:
```
# date: 2020-06-04 12:58:51.924307
# user: dkeitel
# hostname: T490UIB
# PyFstat: v1.3+116.a21cf2a9.dirty
# LAL: 6.21.0 (CLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# LALPulsar: 1.18.2 (CLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# search: MCMCSearch
# parameters: 
# {
#   'BSGL': False,
#   'SSBprec': None,
#   'assumeSqrtSX': None,
#   'binary': False,
#   'detectors': None,
#   'earth_ephem': None,
#   'injectSources': None,
#   'label': 'fully_coherent_search_using_MCMC',
#   'log10beta_min': -0.5,
#   'maxCoverFreq': None,
#   'maxStartTime': 1008640000,
#   'minCoverFreq': None,
#   'minStartTime': 1000000000,
#   'nsteps': [10, 10],
#   'ntemps': 2,
#   'nwalkers': 10,
#   'outdir': 'example_data/fully_coherent_search_using_MCMC',
#   'rhohatmax': 1000,
#   'sftfilepattern': 'example_data/fully_coherent_search_using_MCMC/*fully_coherent_search_using_MCMC*sft',
#   'sun_ephem': None,
#   'tCWFstatMapVersion': 'lal',
#   'theta_initial': None,
#   'theta_prior': { 'Alpha': 1.4596048908088417,
#                    'Delta': 0.38422376285103965,
#                    'F0': { 'lower': 29.99999995,
#                            'type': 'unif',
#                            'upper': 30.00000005},
#                    'F1': { 'lower': -1.0005e-10,
#                            'type': 'unif',
#                            'upper': -9.995e-11},
#                    'F2': 0},
#   'tref': 1004320000.0
# }
# F0 F1
```

@Rodrigo-Tenorio If you know of a built-in way to make pprint do this without manual postproc, let me know. But it works for now.

Rodrigo-Tenorio:
I just had an idea: If we use the json package we can achieve essentially the same behavior in a cleaner way. 

The only "problem" is that keywords follow the Java convention (true/false instead of True/False, null instead of None), but since this is just a print to inform the user and not a proper output to ease possible re-runs, that's fine (actually, even if it were such an output, it would be as easy as using the same package to read the output, so no problem at all).

Example log:

```
# LAL: 6.21.0 (UNCLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# LALPulsar: 1.18.2 (UNCLEAN c6afb0789517073ea7423d2422b1ba844a050ef9)
# search: MCMCSemiCoherentSearch
# parameters:
# {
#   "BSGL": false,
#   "SSBprec": null,
#   "assumeSqrtSX": null,
#   "binary": false,
#   "detectors": null,
#   "earth_ephem": null,
#   "injectSources": null,
#   "label": "semi_coherent_search_using_MCMC",
#   "log10beta_min": -1,
#   "maxCoverFreq": null,
#   "maxStartTime": 1008640000,
#   "minCoverFreq": null,
#   "minStartTime": 1000000000,
#   "nsegs": 10,
#   "nsteps": [
#     3,
#     3
#   ],
#   "ntemps": 1,
#   "nwalkers": 100,
#   "outdir": "example_data/semi_coherent_search_using_MCMC",
#   "rhohatmax": 1000,
#   "sftfilepattern": "example_data/semi_coherent_search_using_MCMC/*semi_coherent_search_using_MCMC*sft",
#   "sun_ephem": null,
#   "theta_initial": null,
#   "theta_prior": {
#     "Alpha": 1.4596048908088417,
#     "Delta": 0.38422376285103965,
#     "F0": {
#       "lower": 29.99999995,
#       "type": "unif",
#       "upper": 30.00000005
#     },
#     "F1": {
#       "lower": -1.0005e-10,
#       "type": "unif",
#       "upper": -9.995e-11
#     },
#     "F2": 0
#   },
#   "tref": 1004320000.0
# }
```

dbkeitel:
@Rodrigo-Tenorio do you understand this test error?
```
E       TypeError: Object of type 'int64' is not JSON serializable
```

Rodrigo-Tenorio:
So... the native integer type is `int`. I'd say `int64` actually refers to one of the integers types implemented by NumPy (which is not native to Python and hence json complains, for some reason). 

I was not expecting this, to be honest

dbkeitel:
Actually, though your solution is indeed quite elegant, I guess it's also a bit of a hack to use json functionality for this purpose; so unless you feel particularly motivated debugging this, I'd suggest going back to the previous solution.

Rodrigo-Tenorio:
Let's go back to your proposal. It's not that important as to spend an hor (given that a future relese will likely solve this problem in another way)

dbkeitel:
Ok thanks for your effort anyway! Merging this version now.

PR #49, opened 2020-06-08
Test against CFSv2 and improve compatibility

### LINKED ISSUE #48: CFSv2 compatibility test
I'd like to add a test that ensures exact reproducibility of CFSv2 results with PyFstat when using the same lalpulsar version. It seems this will require a few small fixes to the handling of default parameters.

dbkeitel:
This is a long branch that had the main goal of introducing a new test case directly comparing a small GridSearch against lalapps_CFSv2. To achieve full compatibility, I had to address 2 main issues:

- Figure out that the running median window defaults were different - there's a user option` RngMedWindow` for it now on the PyFstat side, but I'll likely also patch CFSv2 soon to bring it in line with the lalpulsar standard.
- Get our frequency band estimation in line with what CFSv2 does. I've extended the so-far unused `helper_functions.get_covering_band()` to deal with bands, and use it for this purpose. All ComputeFstat/GridSearch derived classes now have an `estimate_covering_band` flag which is false by default (use user freq ranges input or old default from SFTs), but could be changed to true in the future at least for GridSearches since it's known to be safe from the CFSv2 world, while our SFT-derived defaults were always a bit sketchy.

Other changes:
- validate SFTs in tests
- improve error handling for lal programs
- add read_txt_file_with_header() helper function
- add randSeed option to Writer (e.g. to make tests exactly reproducible when needed)

Closes #48.


dbkeitel:
Since CFSv2 is already updated on lalsuite master (https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1343) but not in the released version, I've changed the test so that it leaves the PyFstat default but manually tells CFSv2 to use 101 bins. That way it's both backwards- and forewards-compatible.

Issue #50, opened 2020-06-18
sqrtSX usage


dbkeitel:
From @Rodrigo-Tenorio in #41:
> - sqrtSX is being (wrongly) used as two different things at different points of this code: As a way of specifying Gaussian noise at MFD and as a PSD measure at ComputeFstat. 
>  - This is problematic, since *no Gaussian noise* is represented as sqrtSX=0 in MFD, but as sqrtSX=1 in ComputeFstat.
> - Short term solution: Warn the user to set sqrtSX=0 when creating SFTs.

The short-term solution is already implemented in #41 but this should be further investigated.

dbkeitel:
So the search classes explicitly call their option `assumeSqrtSX`, so I think there isn't actually a "wrong" double use of the name/option in the package.

However, for safety and for consistency with lalapps_MFD, `Writer` and its children should really default to `sqrtSX=0` unless requested otherwise. I'll look into whether I can adapt tests and examples to work with that.

dbkeitel:
Done in #51.

PR #51, opened 2020-06-23
Writer: default to sqrtSX=0

dbkeitel:
As pointed out by @Rodrigo-Tenorio in #50, the Writer default of `sqrtSX=1` is odd, and particularly dangerous now that we also allow for noiseSFTs. Historically it was probably introduced to match the `assumeSqrtSX=1` default of the F-stat codes (Compute & Predict), but as far as I can see that isn't really necessary, as one can always manually set `assumeSqrtSX=1` for those even if the SFTs actually contain 0 noise.

This PR:
-  changes the Writer default
-  adapts the warning message introduced in #41
-  adds the missing feature for setting assumeSqrtSX when wanting to predict F directly from the Writer
-  improves the test suite

In particular, the only test failing after the first commit was `test_get_semicoherent_twoF`, because it picked up spurious SFTs from the other test case in the same class. So I've now cleared up sftfilepatterns, separated two classes, and introduced joint setup methods where appropriate. Also repurposed the `no_noise` tests to actually do exactly that, and use the PFS trick as described above.

The test suite is still messy and could do with a full rewrite matching `pytest` best practices, but for now it works.

Will need to test some examples before merging, too.

Comments @Rodrigo-Tenorio @GregoryAshton @ReinhardPrix ?

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -428,20 +428,24 @@ transientTau = {:10.0f}\n"""
         if check_ok is False:
             helper_functions.run_commandline(cl_mfd)
 
-    def predict_fstat(self):
+    def predict_fstat(self, assumeSqrtSX=None):
         """ Wrapper to lalapps_PredictFstat """
+        if assumeSqrtSX:
+            sqrtSX = assumeSqrtSX
+        else:
+            sqrtSX = self.sqrtSX
         twoF_expected, twoF_sigma = predict_fstat(
-            self.h0,
-            self.cosi,
-            self.psi,
-            self.Alpha,
-            self.Delta,
-            self.F0,
-            self.sftfilepath,
-            self.minStartTime,
-            self.maxStartTime,
-            self.detectors,
-            self.sqrtSX,
+            h0=self.h0,
+            cosi=self.cosi,
+            psi=self.psi,
+            Alpha=self.Alpha,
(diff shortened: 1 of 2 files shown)


Rodrigo-Tenorio:
Hi @dbkeitel ,

Looks good to me overall.

Just a minor question: 
Do you think we should include a default value of assumeSqrtSX=1 in case one tries to run predict_fstat with sqrtSX=0 (since that would lead to $Fstat \to \infty$)?

I though it was a good idea so I modified the logic to make it easy to include as a trailing `or 1` (friendly reminder: `or` and `and` [return values, not booleans](https://docs.python.org/release/2.7.3/library/stdtypes.html#boolean-operations-and-or-not)). However, I don't like the idea of default values, so I'll let it like this.

dbkeitel:
PredictFstat isn't really a core function, so I don't think we need to hold the user's hands with a default here. When someone wants to use it for sanity checking or testing and gets huge values (usually not inf for me actually), they are probably savvy enough to get the idea what to try.

dbkeitel:
After a lot of rebases on master fixes, rerunning examples with this now.

dbkeitel:
All examples seem to at least run through again (after various unrelated fixes that I've rebased this branch onto). For a few where the output is easy to check, I've done so, but not for all. Anyway this seems robust enough to merge for now.

PR #58, opened 2020-06-26
Add docker workflow

Rodrigo-Tenorio:
Build a Docker container using Python 3.8 and the last commit in master branch.

Containers can be pulled using Docker/Singularity, requiring a Personal Access Token with "read:packages" privileges.

This is a nice-to-have feature, since allows PyFstat to be used out of the box in almost every feasible situation* without any painful dependency drama.

*Provided Singularity is available, which is sort of a current trend

- [x]  Check how to pull without PAT
- [x]  Check requirements.txt and setup.py
- [x] Add documentation on how to use a Personal Access Token

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,54 @@
+name: Docker
+
+on:
+  push:
+    # Publish `master` as Docker `latest` image.
+    branches:
+      - master
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+env:
+  # TODO: Change variable to your image's name.
+  IMAGE_NAME: pyfstat
+
+jobs:
+  # Push image to GitHub Packages.
+  # See also https://docs.docker.com/docker-hub/builds/
+  push:
+
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build image
+        run: docker build . --file Dockerfile --tag $IMAGE_NAME
+
+      - name: Log into registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
+
+      - name: Push image
+        run: |
+          IMAGE_ID=docker.pkg.github.com/${{ github.repository }}/$IMAGE_NAME
+          
+          # Change all uppercase to lowercase
+          IMAGE_ID=$(echo $IMAGE_ID | tr '[A-Z]' '[a-z]')
+
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+
+          # Strip "v" prefix from tag name
+          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+
+          # Use Docker `latest` tag convention
+          [ "$VERSION" == "master" ] && VERSION=latest
+
+          echo IMAGE_ID=$IMAGE_ID
+          echo VERSION=$VERSION
+
+          docker tag $IMAGE_NAME $IMAGE_ID:$VERSION
+          docker push $IMAGE_ID:$VERSION
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM python:3.8-buster
+
+WORKDIR /pyfstat
+COPY requirements.txt .
+RUN pip install -r requirements.txt
+COPY . .
+RUN pip install .
--- a/README.md
+++ b/README.md
@@ -15,6 +15,12 @@ demonstrating different use cases.
 
 ## Installation
 
+### Docker container
+Ready-to-use PyFstat containers are available at the [Packages](https://github.com/PyFstat/PyFstat/packages)
+page. A git-hub account together with a personal access token is required. See the available 
+[documentation](https://github.com/PyFstat/PyFstat/wiki/Containers) to learn how to pull them from the git-hub
+registry using Docker or Singularity.
+
 ### python installation
 This package requires `python3.6+`.
 While many systems come with a system wide python


Rodrigo-Tenorio:
@dbkeitel ,

1) I'll revert the last commit since this branch will be deleted after merging.

2) What do you think of this? 
Right now it only  supports Python 3.8. I think building a container for each python version should be easy (a bit of git hub's "matrix" command trickery?), although it may be pointless unless someone really wants to do a strict comparison.

dbkeitel:
I have no experience with building for docker (hardly any with passively using images, either), so I don't have much of an educated opinion. If you can get it to work to your own requirements, happy to activate it.

Two questions:

1. How are access rights, meta pages etc handled on dockerhub? Can you create some sort of group access linked to the github group?

1. Since it's used by the `Dockerfile`, you may want to double-check if the `requirements.txt` is actually still in sync with `setup.py` - I tend to forget about it.

Rodrigo-Tenorio:
This is kind of the only way I know to use my own python packages on remote machines.
 It may not be the best way of doing it, but it's simple enough not to be problematic.
Of course, any suggestions are welcome.

>     1. How are access rights, meta pages etc handled on dockerhub? Can you create some sort of group access linked to the github group?

The container registry is actually hosted in git-hub. You can notice a new section in the repository's main page ([Packages](https://github.com/PyFstat/PyFstat/packages/287876)). 

However, I realize one should be able to pull containers without access tokens  since this is a public repository (I'm used to work on private repos). I'll have a look into it.
> 
>     2. Since it's used by the `Dockerfile`, you may want to double-check if the `requirements.txt` is actually still in sync with `setup.py` - I tend to forget about it.

I'd say it's not... although now that you mention it, they fulfill different purposes, so I'm not really sure if they have to be in sync (or even if they have to exists at the same time) in the first place. I'll have a look as well.

dbkeitel:
You'll want to rebase this on current master after the lalsuite chaos... But don't worry about doing it over the weekend, I probably won't have time for a closer look until Tuesday.

> This is kind of the only way I know to use my own python packages on remote machines.
It may not be the best way of doing it, but it's simple enough not to be problematic.
Of course, any suggestions are welcome.

I think it's actually considered the best and recommended way by experts. :+1: 

> The container registry is actually hosted in git-hub. You can notice a new section in the repository's main page (Packages).

Nice!

>      I'd say it's not... although now that you mention it, they fulfill different purposes, so I'm not really sure if they have to be in sync (or even if they have to exists at the same time) in the first place. I'll have a look into this as well.

The test pipelines seem to first install packages from requirements.txt, then when installing the package itself they say "requirement already fulfilled" from those, so they would presumably add anything missing at that point. But I should sync them.

dbkeitel:
Requirements updated in #62.

Rodrigo-Tenorio:
Looks like it's not possible (as of now, although it's been already a while) to pull containers from GitHub's registry without a Personal Access Token (PAT) even if it belongs to a public repository. This means one has to have a GitHub account to generate a PAT with "read:packages" privileges.

I propose to merge so we can have access to this feature and not worry about the PAT stuff by now. In case we need containers to be freely pullable (which would be nice but it's not required as of now), we could look into setting up a free dockerhub organization/repository.

dbkeitel:
Could you include instructions in the README please, or if they get too complicated with the PAT stuff, on a wiki page and put just a link in the README?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,54 @@
...
+    # Publish `master` as Docker `latest` image.
+    branches:
+      - master
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+env:
+  # TODO: Change variable to your image's name.
```
dbkeitel, line 14: looks like you've done that already ;)

dbkeitel:
Some more questions:

- [ ] Should the version number show up somewhere on https://github.com/PyFstat/PyFstat/packages/287876 automatically? I see there's some magic in the yml do deal with its formatting and I'm not sure where to see if it dealt correctly with our versioneered numbers?

- [ ] Can the "Let’s write down a good description for this version!" on that same page be done automatically from one of these files, or does it need to be edited there?

- [ ] `docker-publish.yml` says `runs-on: ubuntu-latest` while `Dockerfile` says `FROM python:3.8-buster`; is this is a contradiction or is it about completely different things?

Rodrigo-Tenorio:
>     * [ ]  `docker-publish.yml` says `runs-on: ubuntu-latest` while `Dockerfile` says `FROM python:3.8-buster`; is this is a contradiction or is it about completely different things?

Completely different things. `runs-on` refers to where the git-hub job is executed (like gitlab runners, somehow), while the `Dockerfile` refers to the container we are building.

dbkeitel:
Thanks for the documentation. But there's one point I'm still missing: to actually use the PAT and pull, do I have to first do the `docker login` step as described [here](https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-docker-for-use-with-github-packages#authenticating-to-github-packages) or is there a one-line with `docker pull` only? If the former, can you add that line (and link to the official docs) to https://github.com/PyFstat/PyFstat/wiki/Containers ?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,54 @@
+name: Docker
+
+on:
+  push:
+    # Publish `master` as Docker `latest` image.
+    branches:
+      - master
```
dbkeitel, line 7: Sorry, one more question: since this seems to explicitly say to only publish an image on pushes to master, why is there already an example image from this branch? Did you upload that one manually, or will it in future actually do it automatically for all branches?

Rodrigo-Tenorio: No worries, it's useful to know what parts of this whole kerfuffle are unclear.

That was an unfortunate timing mistake. I added this branch to check whether git-hub pipelines were properly working, and then I reverted the commit to build containers only for master (which ought to be the only branch used for production, hence the only branch for which a container is required).

dbkeitel: That's actually a fortunate timing mistake then, since it told us it's already working. :)

dbkeitel:
@Rodrigo-Tenorio if you can answer the two questions about https://github.com/PyFstat/PyFstat/packages/287876  above, then I'll merge.

Rodrigo-Tenorio:
@dbkeitel 

Git tags are taken into account by the docker pipeline. Hence, we'll always have a "latest" container using the top master commit, plus tagged containers related to tagged commits.

dbkeitel:
From chat discussion with Rodrigo, this is how we understand the versioning and houskeeping of the images will work:

> We'll be having two "families" of containers
>   1.  "latest", created from the top of master (regardless of what that means)
>  1.  "v*", which should be like pyfstat:1.5, created from the last tagged commit

dbkeitel:
> Thanks for the documentation. But there's one point I'm still missing: to actually use the PAT and pull, do I have to first do the `docker login` step as described [here](https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-docker-for-use-with-github-packages#authenticating-to-github-packages) or is there a one-liner with `docker pull` only? If the former, can you add that line (and link to the official docs) to https://github.com/PyFstat/PyFstat/wiki/Containers ?

Wiki updated accordingly, plus `docker run` instructions.

Issue #52, opened 2020-06-26
tests fail with lalsuite 6.71: "not a code object"


dbkeitel:
Since today, even after reverting master to last known working state, with
```
 INTERNALERROR> TypeError: not a code object: <class 'pyfstat.make_sfts.Writer'>
```
I suspect some changes in the github python versions or package dependencies correlated with the design changes, and that somehow makes the syntax checker more strict...?

dbkeitel:
Comparing the last working pipeline outputs with today's, at first I though it was that the test time dependencies had been updated, but [this](https://github.com/PyFstat/PyFstat/pull/55/files#diff-7327122c763912b429300af1b6158a12) didn't help. Instead, it's the lalsuite 6.71 relese that somehow messed things upm and #56 has fixed things for now.

Need to dig into this...

dbkeitel:
Actually, doing a `pip install --upgrade lalsuite` in my local venv, I can now reproduce this! That will make things easier to debug.

dbkeitel:
Fixed by lalsuite-6.72, see #59.

PR #59, opened 2020-06-26
lalsuite-6.72 compatibility

dbkeitel:
- [lalsuite-6.72](https://pypi.org/project/lalsuite/6.72/) has been released on pypi and should solve our scary-looking problems with 6.71 (#52)
 - see https://git.ligo.org/lscsoft/lalsuite/-/issues/340 and https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1369 for the underlying problem and fix
 - first commit reverts commit 2380a6c4bd84660c7daa8dc2194ce3460faa2162 (from #56)
 - additional test errors resolved in further commits

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -294,6 +294,10 @@ def run_commandline(cl, log_level=20, raise_error=True, return_output=True):
     """
 
     logging.log(log_level, "Now executing: " + cl)
+    if "|" in cl:
+        logging.warning(
+            "Pipe ('|') found in commandline, errors may not be" " properly caught!"
+        )
     try:
         if return_output:
             out = subprocess.check_output(
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -309,7 +309,7 @@ transientTau = {:10.0f}\n"""
                 )
 
         logging.info("Checking new commandline against existing SFT header...")
-        cl_dump = "lalapps_SFTdumpheader {}".format(self.sftfilepath)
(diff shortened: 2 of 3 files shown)


dbkeitel:
The scary errors from #52 are gone, but it now fails an actual test:
```
 ______________ BinaryModulatedWriter.test_makefakedata_usecached _______________

self = <tests.BinaryModulatedWriter testMethod=test_makefakedata_usecached>

    def test_makefakedata_usecached(self):
        Writer = self.tested_class(self.label, outdir=self.outdir, duration=3600)
        if os.path.isfile(Writer.sftfilepath):
            os.remove(Writer.sftfilepath)
        # first run: make everything from scratch
        Writer.make_cff()
        Writer.run_makefakedata()
        time_first = os.path.getmtime(Writer.sftfilepath)
        # second run: should re-use .cff and .sft
        Writer.make_cff()
        Writer.run_makefakedata()
        time_second = os.path.getmtime(Writer.sftfilepath)
>       self.assertTrue(time_first == time_second)
E       AssertionError: False is not true

PyFstat/tests.py:107: AssertionError
```

I do **not** see this locally, and once again I don't really see why the new lalsuite could cause this...

dbkeitel:
> I do not see this locally, and once again I don't really see why the new lalsuite could cause this...

The problem was simply that `lalapps_SFTdumpheader` does no longer exist (#60). The reason I didn't see it locally was a dirty environment, still picking it up from an older system install. And the reason the error wasn't obvious was improper checking of failures for piped commandline (#61).

For now it seems we can use current (6.72) lalsuite with this, merging!

PR #66, opened 2020-06-30
add Rodrigo to authors in bibtex (and later Zenodo)

dbkeitel:
@Rodrigo-Tenorio congratulations, I think your contributions have already been sufficient to add you to the author list in the bibtex citation. Thanks a lot for everything so far!

Do you have an ORCID? If yes, please note it here and then we (Greg) can enter that in the next Zenodo release.

@GregoryAshton @ReinhardPrix any objections?


dbkeitel:
Forgot the metadata in setup.py - now incuded. Merging so I can get started on #73.

Issue #74, opened 2020-07-03
no ephemerides in docker container


dbkeitel:
@Rodrigo-Tenorio it looks like the docker container is not actually very useful yet as it doesn't contain ephemerides files. Do you think there's an easy way to do a download like this: https://github.com/PyFstat/PyFstat#ephemerides-installation from the Dockerfile that builds the image, or does that build action not have internet access and we need local copies of them?

Alternatively, we could have a helper function that downloads them on-the-fly, but I think there will likely be use-cases on non-LDG clusters where you want to use Docker but don't have outgoing internet access from worker nodes.

Rodrigo-Tenorio:
@dbkeitel Yes, builds are (or should be) able to get stuff from the internet (e.g. the `pip install -r requirements.txt` needs internet acces). I'll probably add those instructions you just linked to the Dockerfile and that's it.

Question: How difficult is it to include those ephemeris in the package itself or to tell pip to download them? Do you know about the pythonic way of doing this?

dbkeitel:
> I'll probably add those instructions you just linked to the Dockerfile and that's it.

If it can just run wget, that's the easiest way to get them. The tricky part is knowing what's a good local location inside the image to put them into and point the env variable to - but maybe if you know more about their internal structure than me (i.e. more than 0) it's not actually tricky. ;)

> How difficult is it to include those ephemeris in the package itself

I'd like to avoid that (data duplication).

>  or to tell pip to download them

Probably not pip, but may be possible in setup.py.

See also https://git.ligo.org/lscsoft/lalsuite/-/issues/326 for earlier discussion of how to resolve this on the LALSuite side.

Follow-up question for you: do you ever need ephemerides other than DE405?

dbkeitel:
Indeed after #75 was merged to master and the c7bc7fc6 image built by github, I can successfully run examples with it. Thanks Rodrigo!

PR #75, opened 2020-07-03
Add Ephemerides to Dockerfile

Rodrigo-Tenorio:
Solve #74 including steps to download ephemerides during docker build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,25 @@
 FROM python:3.8-buster
 
 WORKDIR /pyfstat
+
+# Ephemerides arguments
+## Put ephemerides in this folder
+ARG EPHEMERIDES_FOLDER=ephemerides/
+## URL to wget ephemerides from
+ARG EPHEMERIDES_URL=https://git.ligo.org/lscsoft/lalsuite/raw/master/lalpulsar/lib/
+## Earth epehemerides file name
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,25 @@
 FROM python:3.8-buster
 
 WORKDIR /pyfstat
+
+# Ephemerides arguments
+## Put ephemerides in this folder
+ARG EPHEMERIDES_FOLDER=ephemerides/
```
Rodrigo-Tenorio: Actually, I'm not sure of this relative path:

WORKDIR ensures that we are referring to /pyfstat/ephemerides, but once spawned, LALPULSAR_DATADIR will be "ephemerides/" only (relative path)

dbkeitel:
I guess you could temporarily add the branch to the .yml, let it build the image once, then when it looks correct, drop that change again from the commit before merging...?

Rodrigo-Tenorio:
GitHub does not let you remove packages from a public repository (e.g. look at the vDummyTag), so I'd prefer to keep the docker-clutter low. 

I'm building locally, using get_ephemeris_files to check whether ephemeris are there.

Rodrigo-Tenorio:
@dbkeitel that should be it.  

- `ARG` values can be set at build time (in case anyone wants to try some fancy ephemerides). 
- `ENV` will set environmental variables at run time (i.e where are the ephemerides?)

Issue #76, opened 2020-07-03
docker images don't pick up full version(eer) numbers


dbkeitel:
To be looked at next week, just so it doesn't get forgotten:

Inside a pyfstat:latest docker container one gets `pyfstat._version_` as something like `0+untagged.1.c7bc7fc6.clean` whereas directly from source it should be `v1.3+190.c7bc7fc6.clean`. So it gets the right git hash, but doesn't get the latest previous tag (first bit) and the second bit where versioneer counts the number of commits since.  ("bits" here always meaning entries of version.split("."))

Probably this is versioneer trying to dynamically figure out the first two bits while running in the container, but not having access to the git. Maybe we'd need to teach the dockerfile to make a fake tag that matches the expected parts, and then hard-write it into the image, and/or teach versioneer to notice it's in a container and do things differently. To be investigated further.

Rodrigo-Tenorio:
@dbkeitel 

It seems to be a [known issue](https://github.com/opsdroid/opsdroid/issues/950). I found [this example from git.ligo](https://git.ligo.org/emfollow/gwcelery/commit/220adecd356a3e71953967bb5d55c8823c143826) and [this git-hub issue](https://github.com/warner/python-versioneer/issues/170) where a solution seems to be explained and implemented, somehow.

versioneer seems to be unable to pick version information from wheels (compiled python packages...?), so they build the source distribution first (which is understood by versioneer) and then the wheel.

This is kind of alien to me right now (0% experience working with wheel/source distributions; I'm happy running pip install), but it looks like we should do that inside the Dockerhub instead of directly running `pip install .`?

dbkeitel:
Yes I also read it this way: the key from Leo's fix seems to be to build a "sdist" first, then build both the binary "wheels" and the docker image after that, _in the same job_. That also seems to match what the other discussion implies.

No idea if this will be a possible workaround path for docker-only jobs, but yes, building a distribution before installing should be possible there too. Feel free to try it if you have an idea how to start; I hope it's as easy as following Leo's logic in the github-instead-of-gitlab yml flavour. Otherwise you can wait for me to try setting up the pypi uploads first and then maybe a combined job can be constructed.

Rodrigo-Tenorio:
[actions/checkout@v2](https://github.com/actions/checkout#Fetch-all-history-for-all-tags-and-branches) only fetches the last commit. If we want the whole history, we need to add the following option to the action:
- uses: actions/checkout@v2
  with:
[...]

PR #79, opened 2020-07-06
docker build: Fetch all history using checkout@v2

Rodrigo-Tenorio:
checkout@v2 only picks up the last commit by default [see the official docs](https://github.com/actions/checkout#Fetch-all-history-for-all-tags-and-branches). Add the suitable argument to the action file and #76 is done!

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -11,7 +11,6 @@ on:
       - v*
 
 env:
-  # TODO: Change variable to your image's name.
   IMAGE_NAME: pyfstat
 
 jobs:
@@ -24,6 +23,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+            fetch-depth: 0
 
       - name: Build image
         run: docker build . --file Dockerfile --tag $IMAGE_NAME


Rodrigo-Tenorio:
This is what I get after pulling the docker image:
```
[rodrigo@spinning-quadrupole PyFstat]$ docker run -it --entrypoint /bin/bash  --rm docker.pkg.github.com/pyfstat/pyfstat/pyfstat:fix-docker-versioning
root@183326110d00:/pyfstat-repo# python
Python 3.8.3 (default, Jun  9 2020, 17:39:39)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyfstat
>>> pyfstat.__version__
'v1.3+196.6a3734fa.clean'
```
Version is properly parsed!

.github/workflows/docker-publish.yml
```diff
@@ -5,13 +5,13 @@ on:
     # Publish `master` as Docker `latest` image.
     branches:
       - master
+      - fix-docker-versioning
```
dbkeitel, line 8: Wait, actually this should be removed again now that the "test job" finished successfully.

PR #80, opened 2020-07-07
update packaging for PyPI

dbkeitel:
So far just updated/added some metadata in setup.py:
 -maintainer
 -description
 -url
 -platforms
 -classifiers

Test upload is here: https://test.pypi.org/project/PyFstat/1.4/

Will see if anything else needs to be cleaned up.


dbkeitel:
Packaging extra files is weird. I found many contradictory instructions online; in particular, the statements from https://setuptools.readthedocs.io/en/latest/setuptools.html that "data_files is deprecated. It does not work with wheels, so it should be avoided." seems to contradict my experiments: in fact I **need** to use `data_files` **specifically for wheels**. For a source distribution tarball (sdist), `include_package_data` plus `MANIFEST.in` worked well, but not for wheels.

For now I'm trying with the `MANIFEST.in` completely removed (as some sources say it's mostly a python2 backwards compatibility feature, and pure setup.py parameters to be preferred).

Inside the wheels (`.whl` file), the examples now appear in an ugly path `PyFstat-1.3+200.2efe6483.dirty.data/data/examples`; will check in the next step how those turn out when installed from testpypi...

dbkeitel:
> Inside the wheels (.whl file), the examples now appear in an ugly path PyFstat-1.4.1.data/data/examples; will check in the next step how those turn out when installed from testpypi...

Installing into a venv,
```
python3 -m venv testvenv
source testvenv/bin/activate
pip install --upgrade pip
pip install -r git/sources/PyFstat/requirements.txt
pip install --index-url https://test.pypi.org/simple/ pyfstat
```
they do end up in the root of that environment:
```
ls /home/dkeitel/testvenv/
bin/        etc/        examples/   include/    lib/        lib64/      pyvenv.cfg  share/      var/
```
Not ideal. Either I'll figure out a way to put them under "opt" or something, or we need to move them into the package directory itself.

dbkeitel:
I've changed the examples to be included through the `scripts` argument of setup; now they end up under `bin`. Since either like this or with the previous version, the wheels would never preserve the subdirectories, I also renamed them all to have consistent and unique `PyFstat_example_*` names.

However, having examples as installed `bin` environment-wide accessible seems a bit out of scope, so I'm not so sure if this is really the right solution.

dbkeitel:
I have now removed the examples from the wheels, but they're still included in the source distribution (sdist).

@Rodrigo-Tenorio while I test upload to test.pypi.org again, could you have a look over the [updated README](https://github.com/PyFstat/PyFstat/blob/9f36a3ab7fcf21d63584186310aa778e50d3fae7/README.md) please?

Rodrigo-Tenorio:
The most important change is "documentation" to "wiki page".

dbkeitel:
Thanks!

In the end I went back to a hybrid MANIFEST.in / setup.py approach which seems to work best to build clean versions of both sdist and wheels.

Merging now, tagging as 1.4.0 and trying the real upload...

PR #85, opened 2020-07-14
integration.yml: put black first, build distributions, upload releases to TestPyPi

dbkeitel:
1. put black before pytest
1. (still to do): try automated sdist/wheels builds and maybe even PyPi upload

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -45,6 +45,7 @@ jobs:
         python setup.py sdist bdist_wheel
         check-wheel-contents dist/*.whl
     - name: Publish package to TestPyPI
+      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@master
       with:
         user: __token__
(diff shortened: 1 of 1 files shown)


dbkeitel:
Current state:

- On every commit, will build sdist and wheel, and test the wheel with [check-wheel-contents](https://github.com/jwodder/check-wheel-contents).
- On every commit that also has a tag attached, will try to upload the wheels to https://test.pypi.org/project/PyFstat/ using a "pyfstatdev" account I'll share with @GregoryAshton and with an access token stored as a secret in this repo.
- Not the real PyPi index for now until after I've seen it in action on a real tag/release.
- Can't do the upload on every commit, since our versioneer version numbers ("local" by PEP440 logic) aren't accepted by PyPi.
- The one part I'm not sure about is whether the "on push" action will actually trigger at all when I add a tag (either from my local machine via `git push --tags` or from the github web interface) after the commit is already on master. Will have to test that once the next release is ready, and use a separate action if this doesn't work.

For now it passes while skipping the publish job, and in earlier commits I tested that it fails "as expected" due to the version number, and only runs for a single python version (3.8 currently) to avoid race conditions. Seems good to merge for now and see what happens later.

@Rodrigo-Tenorio since you seem quite yml-fluent, could you please give this a quick review? I think I've also found the setting to force myself to play by the rules, so you'll actually have to use the review feature and approve this. (_If_ someone with lower privileges can approve my PR... will also have to test that.)

PR #89, opened 2020-07-14
publish releases to real PyPI

dbkeitel:
First to TestPyPI, then the real one, with same logic (only on tagged releases and only for newest python version).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -51,3 +51,9 @@ jobs:
         user: __token__
         password: ${{ secrets.test_pypi_password }}
         repository_url: https://test.pypi.org/legacy/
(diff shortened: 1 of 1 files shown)


dbkeitel:
The 1.4.2 auto-release [failed](https://github.com/PyFstat/PyFstat/runs/870757532?check_suite_focus=true) at the TestPyPI stage with "This filename has already been used, use a different version." presumably because I had done some test uploads including that version number before, even though I've since deleted them.
Manual upload to the real PyPI [worked](https://pypi.org/project/PyFstat/1.4.2/). Same failure may appear twice more because I also tried 1.4.3 and 1.4.4 on TestPyPI before.

PR #91, opened 2020-07-15
make_sfts: change options to better distinguish signal duration vs data duration

### LINKED ISSUE #90: make_sfts cleanup
Some things I noticed to improve in `make_sfts.py` while looking for something else:

- [x] `duration` gets needlessly converted to days and back
-> #91

- [x] `duration` vs `data_duration` are needlessly complicated for CWs, and also used a bit confusingly for transients, so should clean this up a little or even make a separate `TransientWriter` class.
-> #91

- [x] This doesn't look robust for general applications:
```
    def calculate_fmin_Band(self):
        self.fmin = self.F0 - 0.5 * self.Band
```

- [x] Several tests manipulate `Writer.tstart`, `Writer.duration` and `Writer.tend` manually after the object has been set up, to then re-use it for a second data generation, and this seems like something real users may also be tempted to do. But it is potentially dangerous since the 3 variables really only have 2 degrees of freedom. A joint `update_data_duration()`, or so, method should be provided.

dbkeitel:
This will partially address #90 but further PRs will be required. See commit messages and comments below for details.

### DIFF
--- a/examples/transient_examples/PyFstat_example_long_transient_MCMC_search.py
+++ b/examples/transient_examples/PyFstat_example_long_transient_MCMC_search.py
@@ -11,20 +11,23 @@ Alpha = 0.5
 Delta = 1
 
 minStartTime = 1000000000
-maxStartTime = minStartTime + 200 * 86400
+duration = 200 * 86400
+maxStartTime = minStartTime + duration
 Tspan = maxStartTime - minStartTime
 tref = minStartTime
 
 DeltaF0 = 6e-7
 DeltaF1 = 1e-13
 
+# to make the search cheaper, we exactly target the transientStartTime
+# to the injected value and only search over TransientTau
 theta_prior = {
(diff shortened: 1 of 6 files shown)


dbkeitel:
@GregoryAshton The first commit is fairly trivial, but the second "hat es in sich" if you remember that German phrase. ;) It _should_ change absolutely nothing for standard CWs (tests and examples all pass), but is backwards-incompatible in what arguments users need to pass for transients and I also needed to adapt the glitch test to make it pass.

Basically I'm fixing the logic regarding what is "signal lenght" and "data length" in the `Writer` classes, by making the default `tstart` and `duration` always refer to the **full data set**, deprecate the `minStartTime`/`maxStartTime` options completely which are IMHO only confusing in this context, and requiring more explicitly-named options for transients. This also brings our interface closer to lalapps_MFDv5 itself.

An actual review would be good on this one.

I also have a few further changes planned, will list those under #90 and make separate PRs, but I think they'll have less chance of breaking things.

GregoryAshton:
I ran through the changes @dbkeitel and it all looks good to me!

PR #95, opened 2020-07-17
Estimate coverage bands by default, extend feature to binaries

dbkeitel:
Another big-ish change. A while ago in a9e993dddae9424d4f01d3f24d42435228e348e8 I introduced the possibility to call an XLAL function to estimate the "covering band" to actually grab from the given SFTs. This is essential in being able to 1:1 reproduce results from CFSv2, and also should be more robust and general than the previous practice of estimating as `[min(SFTs)+0.5,max(SFTs)-0.5]` if the user didn't give manual options.

New behaviour now:

- use `estimate_min_max_CoverFreq()` by default,  based on search ranges (grid boundaries for grid searches / prior bounds for MCMC searches)
- deprecate `estimate_covering_band` user-facing option (has only been live for a month, so shouldn't break too much)
- old behaviour of SFT ranges +- 0.5 can now be reproduced by passing negative minCoverFreq,maxCoverFreq (or more flexible offsets from SFT boundaries)
- binary parameters are added in the second commit, please check @Rodrigo-Tenorio 
- on the other hand, ignoring transient parameters is conservative and hence should be safe

Once again I hacked the specialised search classes (glitches, slices, sliding windows) until tests and examples worked, but I don't understand all their details and can't guarantee they still work fully as intended.

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -999,6 +999,9 @@ class ComputeFstat(BaseSearchClass):
         Tsft = 1.0 / catalog.data[0].header.deltaF
         startTime = catalog.data[0].header.epoch
         endTime = catalog.data[-1].header.epoch + Tsft
+        # start by constructing a DopplerRegion structure
+        # which will be needed to conservatively account for sky-position dependent
+        # Doppler shifts of the frequency range to be covered
         searchRegion = lalpulsar.DopplerRegion()
         # sky region
         Alpha = self.search_ranges["Alpha"][0]
@@ -1029,6 +1032,7 @@ class ComputeFstat(BaseSearchClass):
                     if len(self.search_ranges[Fk]) >= 2
                     else 0.0
                 )
+        # now construct DopplerFullScan from searchRegion
         scanInit = lalpulsar.DopplerFullScanInit()
         scanInit.searchRegion = searchRegion
         scanInit.stepSizes = lalpulsar.PulsarDopplerParams()
@@ -1055,8 +1059,30 @@ class ComputeFstat(BaseSearchClass):
         scanInit.startTime = startTime
         scanInit.Tspan = float(endTime - startTime)
         scanState = lalpulsar.InitDopplerFullScan(scanInit)
+        # now obtain the PulsarSpinRange extended over all relevant Doppler shifts
         spinRangeRef = lalpulsar.PulsarSpinRange()
         lalpulsar.GetDopplerSpinRange(spinRangeRef, scanState)
+        # optional: binary parameters
+        if "asini" in range_keys:
+            if len(self.search_ranges["asini"]) >= 2:
+                maxOrbitAsini = self.search_ranges["asini"][1]
+            else:
+                maxOrbitAsini = self.search_ranges["asini"][0]
+        else:
+            maxOrbitAsini = 0.0
+        if "period" in range_keys:
+            minOrbitPeriod = self.search_ranges["period"][0]
+        else:
+            minOrbitPeriod = 0.0
+        if "ecc" in range_keys:
+            if len(self.search_ranges["ecc"]) >= 2:
+                maxOrbitEcc = self.search_ranges["ecc"][1]
+            else:
+                maxOrbitEcc = self.search_ranges["ecc"][0]
+        else:
+            maxOrbitEcc = 0.0
+        # finally call the wrapped lalpulsar estimation function with the
+        # extended PulsarSpinRange and optional binary parameters
         self.minCoverFreq, self.maxCoverFreq = helper_functions.get_covering_band(
             tref=self.tref,
             tstart=startTime,
@@ -1067,9 +1093,9 @@ class ComputeFstat(BaseSearchClass):
             F0band=spinRangeRef.fkdotBand[0],
             F1band=spinRangeRef.fkdotBand[1],
             F2band=spinRangeRef.fkdotBand[2],
-            orbitasini=0.0,  # FIXME
-            orbitPeriod=0.0,  # FIXME
-            orbitEcc=0.0,  # FIXME
+            maxOrbitAsini=maxOrbitAsini,
+            minOrbitPeriod=minOrbitPeriod,
+            maxOrbitEcc=maxOrbitEcc,
         )
 
     def get_fullycoherent_twoF(
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -496,6 +496,7 @@ class ComputeFstat(BaseSearchClass):
         detectors=None,
         minCoverFreq=None,
         maxCoverFreq=None,
+        search_ranges=None,
         injectSources=None,
         injectSqrtSX=None,
         assumeSqrtSX=None,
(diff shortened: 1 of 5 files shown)


dbkeitel:
Improved documentation a bit.

@Rodrigo-Tenorio have you had a chance to take a look? Especially the binary parameter handling [here](https://github.com/PyFstat/PyFstat/pull/95/files#diff-ed8f52637a7d535159fd46ef4d8a91a2R1054) and [here](https://github.com/PyFstat/PyFstat/pull/95/files#diff-a92bbecb209ef931c587d93a1dbc2a16) as well as the [prior range logic for MCMCSearch](https://github.com/PyFstat/PyFstat/pull/95/files#diff-622cde0a01ba6381361488f5987cb983R223).

pyfstat/helper_functions.py
```diff
@@ -354,18 +354,26 @@ def get_covering_band(
     F0band=0.0,
     F1band=0.0,
     F2band=0.0,
-    orbitasini=0.0,
-    orbitPeriod=0.0,
-    orbitEcc=0.0,
+    maxOrbitAsini=0.0,
+    minOrbitPeriod=0.0,
+    maxOrbitEcc=0.0,
```
Rodrigo-Tenorio, line 359: Maximum `asini` and minimum `orbital period` (which implies maximum `orbital frequency`). Again, this makes sense.

pyfstat/mcmc_based_searches.py
```diff
@@ -220,13 +220,32 @@ def _log_input(self):
         logging.info("ntemps = {}".format(self.ntemps))
         logging.info("log10beta_min = {}".format(self.log10beta_min))
 
+    def _get_search_ranges(self):
```
Rodrigo-Tenorio, line 223: This looks good and consistent with the indexing in the other bit of code.

pyfstat/mcmc_based_searches.py
```diff
@@ -1033,44 +1052,57 @@ def _add_prior_to_corner(self, axes, samples, add_prior):
             for k in range(0, i):
                 axes[i][k].set_ylim(xlim[0], xlim[1])
 
+    def _get_prior_bounds(self, normal_stds=2):
```
Rodrigo-Tenorio, line 1055: I wonder if we should be more agressive and set normal_stds=3.

Otherwise, I like it.

pyfstat/core.py
```diff
@@ -922,48 +911,167 @@ def init_computefstatistic(self):
...
+                    "If sftfilepattern==None, cannot use negative values for"
+                    " minCoverFreq or maxCoverFreq (interpreted as offsets from"
+                    " min/max SFT frequency)."
+                    " Please use actual frequency values for both,"
+                    " or set both to 'None' (automated estimation)."
+                )
+            if self.minCoverFreq < 0.0:
+                logging.info("minCoverFreq not provided, estimating from SFTs.")
+                fAs = [d.header.f0 for d in self.SFTCatalog.data]
+                self.minCoverFreq = np.min(fAs) + 0.5
```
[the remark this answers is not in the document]
dbkeitel: The -0.5 was the old default behaviour. But you're completely right that I forgot to overwrite it with the user input...

pyfstat/core.py
```diff
@@ -922,48 +911,167 @@ def init_computefstatistic(self):
...
         lalpulsar.GetDopplerSpinRange(spinRangeRef, scanState)
+        # optional: binary parameters
+        if "asini" in range_keys:
+            if len(self.search_ranges["asini"]) >= 2:
+                maxOrbitAsini = self.search_ranges["asini"][1]
+            else:
+                maxOrbitAsini = self.search_ranges["asini"][0]
+        else:
+            maxOrbitAsini = 0.0
+        if "period" in range_keys:
```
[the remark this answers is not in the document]
dbkeitel: Also, it's taken directly from lalpulsar as is..

(For the record, I always dislike using period instead of frequency, exactly because it turns the logic upside down. ;) )

pyfstat/mcmc_based_searches.py
dbkeitel: That was taken from the original `plot_prior_posterior()` where I stole most of this per-prior code. But since that one has its own default which can be different, and here we really want to be reasonably sure we are conservative, it's a good idea to go higher by default.

pyfstat/core.py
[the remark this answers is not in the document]
dbkeitel: Fixed.

pyfstat/mcmc_based_searches.py
dbkeitel: I've added another commit that changes the default to 3 (at the `_get_search_ranges()` level, though), but also throws a warning whenever this truncation happens, and instructs the user to manually set [minCoverFreq,maxCoverFreq] instead if there are problems. Because due to the Gaussian's infamous infinite tails and MCMC's small but nonvanishing chances to explore them, in principle no number of stds is sufficient to safely predict reaching outside the estimated coverage band.

dbkeitel: Changing the `MCMCSearch::test_fully_coherent` test case to a wide prior `"F0": {"type": "norm", "loc": F0, "scale": 1,` (and the Writer band from 4 to 10 tu support it), I can indeed produce the expected effect that a search can easily fail by trying to sample outside the `normal_stds=3`.

However I think there's no fundamental solution that can be easily provided here. The warning
```
WARNING  root:mcmc_based_searches.py:234 Gaussian priors (normal / half-normal) have been truncated at 3.000000 standard deviations for estimating the coverage frequency band. If sampling fails at any point, please consider manually setting [minCoverFreq,maxCoverFreq] to more generous values.
```
should be enough to guide the user to a workaround, even if it will require trial&error.

In most practical cases we'll use uniform, not normal, priors on frequency anyway, and this won't occur.

dbkeitel:
I though this may need a specialised test case, but in addition to the grid-based tests I've already adapted to cover the various input options, `MCMCSearch::test_fully_coherent` actually covers the estimate-from-priors case pretty well.

@Rodrigo-Tenorio if you're fine with the "resolution" of the Gaussian prior issue above, this is ready to merge I'd say.

dbkeitel:
...rebased to master to be on the safe side.

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: What about being more conservative (like 5 sigmas)? 
I guess the answer is "Well, that will just make the error less likely ;)"
Anyways:
1. Follow Ups require well-defined "small" prior volumes to be effective. Why would you use a Gaussian then?
2. In case you need to use a Gaussian (what do you want to prove?), you'd be smart enough to locate it "properly" so the MCMC does not bail away....

I'm not sure of 2. being effective at all (the more I think about it, the less convinced I am), so let's go with the warning!

dbkeitel: Going to  more sigmas has the problem of in turn raising the chance of the provided SFTs not being wide enough. In some sense, having that error up front would be better than after a long sampling run randomly falls outside, but not sure it's worth it.

PR #98, opened 2020-07-27
Call computetransientfstatmap once

Rodrigo-Tenorio:
Semicoherent Fstat is computed using LALSuite's TransientFstatMap (there's no native implementation otherwise!); such implementation was properly documented in [the paper](https://arxiv.org/abs/1802.05450) through a characteristic time tau_T ~ 1.5e-8 (s), which conforms the main time contribution for all our purposes (i.e. Nsft ~ O(1e3 - 1e4) and Nsegments ~ O(1e2 - 1e3) ).

This PR improves tau_T by a factor of ~5 (tau_T ~ 2.8e-9 s) introducing the following changes:

- Given *nsegs* segments, we set up *nsegs* rectangular transient windows and call  `lalpulsar.ComputeTransientFstatMap` **once** (previous approach used *one* rectangular window and called the function *nsegs* times).

- We introduce a new variable `self._semicoherent_window_trick` to carry the transient window around, effectively uncoupling semicoherent and transient functionalities.

- NaN checks are marginally improved (we have to explicitly handle the array of coherent Fstat values).

- SemiCoherent Fstat is computed using numpy.ndarray.sum

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1574,17 +1574,18 @@ class SemiCoherentSearch(ComputeFstat):
         self._semicoherent_window_trick = lalpulsar.transientWindowRange_t()
         self._semicoherent_window_trick.type = lalpulsar.TRANSIENT_RECTANGULAR
 
-        # Range [t0, t0+t0Band] step dt0 
+        # Range [t0, t0+t0Band] step dt0
         self._semicoherent_window_trick.t0 = int(self.tboundaries[0])
-        self._semicoherent_window_trick.t0Band = int(self.tboundaries[-1] - self.tboundaries[0] - 2 * self.Tcoh)
+        self._semicoherent_window_trick.t0Band = int(
+            self.tboundaries[-1] - self.tboundaries[0] - 2 * self.Tcoh
+        )
         self._semicoherent_window_trick.dt0 = int(self.Tcoh)
-        
+
         # Range [tau, tau + tauBand] step dtau
         # Watch out: dtau must be !=0, but tauBand==0 is allowed
         self._semicoherent_window_trick.tau = int(self.Tcoh)
         self._semicoherent_window_trick.tauBand = int(0)
-        self._semicoherent_window_trick.dtau = int(1) # Irrelevant
-
+        self._semicoherent_window_trick.dtau = int(1)  # Irrelevant
 
     def init_semicoherent_parameters(self):
         logging.info(
@@ -1658,7 +1659,6 @@ class SemiCoherentSearch(ComputeFstat):
         #                                       self.BSGLSetup)
         #    return log10_BSGL/np.log10(np.exp(1))
 
-
         det_stat_per_segment = self._get_per_segment_det_stat()
         detStat = det_stat_per_segment.sum()
         if record_segments:
@@ -1678,11 +1678,15 @@ class SemiCoherentSearch(ComputeFstat):
             FstatResults_single.lenth = 1
             FstatResults_single.data = self.FstatResults.multiFatoms[0].data[0]
             FS0 = lalpulsar.ComputeTransientFstatMap(
-                FstatResults_single.multiFatoms[0], self.windowRange, False
+                FstatResults_single.multiFatoms[0],
+                self._semicoherent_window_trick,
+                False,
             )
             FstatResults_single.data = self.FstatResults.multiFatoms[0].data[1]
             FS1 = lalpulsar.ComputeTransientFstatMap(
-                FstatResults_single.multiFatoms[0], self.windowRange, False
+                FstatResults_single.multiFatoms[0],
+                self._semicoherent_window_trick,
+                False,
             )
 
             self.twoFX[0] = 2 * FS0.F_mn.data[0][0]
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1568,8 +1568,8 @@ class SemiCoherentSearch(ComputeFstat):
     def _init_semicoherent_window_trick(self):
         """
         Use this window to compute the semicoherent Fstat using TransientFstatMaps.
-        This way we are able to decouple this clever shenanigan from the actual
-        usage of a transient window.
+        This way we are able to decouple the semicoherent computation from the 
+        actual usage of a transient window.
         """
         self._semicoherent_window_trick = lalpulsar.transientWindowRange_t()
         self._semicoherent_window_trick.type = lalpulsar.TRANSIENT_RECTANGULAR
@@ -1673,6 +1673,7 @@ class SemiCoherentSearch(ComputeFstat):
 
         if self.BSGL is False:
             d_detStat = 2 * FS.F_mn.data[:, 0]
+        # FIXME: BSGL should not be compute per segment, but for summed F!
         else:
             FstatResults_single = copy.copy(self.FstatResults)
             FstatResults_single.lenth = 1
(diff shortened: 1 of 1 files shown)


pyfstat/core.py
```diff
@@ -1565,6 +1565,28 @@ def __init__(
         self.init_computefstatistic()
         self.init_semicoherent_parameters()
 
+    def _init_semicoherent_window_trick(self):
+        """
+        Use this window to compute the semicoherent Fstat using TransientFstatMaps.
+        This way we are able to decouple this clever shenanigan from the actual
```
dbkeitel, line 1571: While I agree this is a "clever shenanigan", I don't think that's good documentation style. ;)

pyfstat/core.py
```diff
@@ -1565,6 +1565,28 @@ def __init__(
...
+        usage of a transient window.
+        """
+        self._semicoherent_window_trick = lalpulsar.transientWindowRange_t()
```
dbkeitel, line 1574: Is it "pythonic" to have variables (as opposed to methods) with `_`? I don't think we're doing that anywhere else, at least?

dbkeitel: Another, semantically more correct and possibly cleaner option, would be to rename this one `self.semicoherentWindowRange` and the existing `self.windowRange` into `self.transientWindowRange`.

pyfstat/core.py
```diff
@@ -1636,38 +1659,34 @@ def get_semicoherent_twoF(
...
         )
 
         if self.BSGL is False:
-            d_detStat = 2 * FS.F_mn.data[0][0]
+            d_detStat = 2 * FS.F_mn.data[:, 0]
         else:
             FstatResults_single = copy.copy(self.FstatResults)
             FstatResults_single.lenth = 1
             FstatResults_single.data = self.FstatResults.multiFatoms[0].data[0]
             FS0 = lalpulsar.ComputeTransientFstatMap(
```
dbkeitel, line 1680: It's unrelated to your change, but since you're touching this part anyway, could you add a one-liner comment above this line please:
```
# FIXME: this currently hard-codes BSGL for exactly 2 detectors.
```
Increases the chance that I will at some point fix it. ;)

pyfstat/core.py
```diff
@@ -1676,9 +1695,9 @@ def _get_per_segment_det_stat(self, tstart):
                 2 * FS.F_mn.data[0][0], self.twoFX, self.BSGLSetup
             )
             d_detStat = log10_BSGL / np.log10(np.exp(1))
-        if np.isnan(d_detStat):
+        if np.isnan(np.sum(d_detStat)):
             logging.debug("NaNs in semi-coherent twoF treated as zero")
```
dbkeitel, line 1699: Also unrelated to your change, but a quick fix: this should actually say `in semi-coherent detection statistic`.

pyfstat/core.py
```diff
@@ -1676,9 +1695,9 @@ def _get_per_segment_det_stat(self, tstart):
                 2 * FS.F_mn.data[0][0], self.twoFX, self.BSGLSetup
             )
             d_detStat = log10_BSGL / np.log10(np.exp(1))
-        if np.isnan(d_detStat):
+        if np.isnan(np.sum(d_detStat)):
```
dbkeitel, line 1698: I think you want instead `if np.any(np.isnan(d_detStat)):` - the result should always be the same as `[0,1,2,nan].sum()=nan` hopefully, but it makes clearer what is being tested. (This function doesn't know the caller will want to take a sum.)

dbkeitel:
@Rodrigo-Tenorio ignore my first two BSGL-related comments. I think the way it was implemented before wasn't right to start with: it should not be computed in each segment and then summed up. Instead, `_get_per_segment_det_stat` should just always compute `F` only, and then `get_semicoherent_twoF` should compute BSGL from the **summed** F-stats. (Coincidentally, the function names are the wrong way round!)

So just put in the one extra `#FIXME` please and then we can merge this, and I'll make an extra PR reorganising things for BSGL.

pyfstat/core.py
```diff
@@ -1636,38 +1659,34 @@ def get_semicoherent_twoF(
...
-    def _get_per_segment_det_stat(self, tstart):
-        self.windowRange.t0 = int(tstart)  # TYPE UINT4
-
+    def _get_per_segment_det_stat(self):
         FS = lalpulsar.ComputeTransientFstatMap(
-            self.FstatResults.multiFatoms[0], self.windowRange, False
+            self.FstatResults.multiFatoms[0], self._semicoherent_window_trick, False
         )
 
         if self.BSGL is False:
```
dbkeitel, line 1674: Instead of the one I suggested at first, please add
```
#FIXME:  BSGL should not be computed per segment, but from summed F!
```

pyfstat/core.py
Rodrigo-Tenorio: > Is it "pythonic" to have variables (as opposed to methods) with `_`? I don't think we're doing that anywhere else, at least?

[Somehow, yes](https://www.python.org/dev/peps/pep-0008/). The intention is to say "This variable is spookier than you think; don't mess around with it unless you know what you are doing".



> Another, semantically more correct and possibly cleaner option, would be to rename this one `self.semicoherentWindowRange` and the existing `self.windowRange` into `self.transientWindowRange`.

I'd say `self.windowRange` was not doing what it  looks like it's doing... mainly because it was only used to do the Transient trick and nothing else.

I'd prefer to hide this variable since it's not a transient window to compute semicoherent transient statistics, but a transient window to do some "look-at-me-I'm-using-polymorphism" magic

dbkeitel: Since we have a lot of other clearly "internal"/"spooky" variables that aren't name that way, I think I'd prefer to not introduce this convention just for this one. Calling it `self.semicoherentWindowRange` should make clear what it does.

pyfstat/core.py
Rodrigo-Tenorio: That was the first thing I implemented, and I agree with the "clearer".

However...

```
In [2]: import numpy as np

In [3]: a = np.random.random((10000, 10000))

In [4]: %timeit np.isnan(a)
51.8 ms ± 528 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [5]: %timeit np.sum(a)
37.3 ms ± 243 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %timeit np.any(np.isnan(a))
57.6 ms ± 151 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %imteit np.isnan(np.sum(a))
UsageError: Line magic function `%imteit` not found.

In [8]: %timeit np.isnan(np.sum(a))
37.3 ms ± 106 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

although for smaller arrays...
```
In [9]: a = np.random.random((100, 100))

In [10]: %timeit np.sum(a)
4.95 µs ± 33.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [11]: %timeit np.isnan(a)
2.19 µs ± 16.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [12]: %timeit np.any(np.isnan(a))
5.56 µs ± 55.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [13]: %timeit np.isnan(np.sum(a))
7.07 µs ± 253 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

dbkeitel: Alternative: since `_get_per_segment_det_stat()` is only ever called from `get_semicoherent_twoF()`, just move the nan check into that one, after the sum is already taken anyway.

pyfstat/core.py
Rodrigo-Tenorio: I feel we should constraint more the usage of BSGL since neither its implementation nor its results are tested... but that can be done in another PR.

pyfstat/core.py
Rodrigo-Tenorio: Ok, I like the internal consistency argument.

pyfstat/core.py
Rodrigo-Tenorio: Ok, I like that alternative

Issue #99, opened 2020-07-27
Test BSGL thoroughly


Rodrigo-Tenorio:
Lately, some developments may have broken BSGL-related features (see e.g. #98 ). 

Actual numerical checks on BSGL should be implemented,  since there's only one "sanity" test [here](https://github.com/PyFstat/PyFstat/blob/master/tests.py#L550) which only checks if the value is positive.

- [ ] Implement numerical tests for BSGL statistic

dbkeitel:
I'm adding one more simple test in #100, but to do this properly the best would be to reproduce part or all of lalpulsar's [LineRobustStatsTest.c](https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalpulsar/test/LineRobustStatsTest.c) and maybe some bits from [testHierarchSearchGCT.sh](https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalapps/src/pulsar/GCT/testHierarchSearchGCT.sh) too.

dbkeitel:
> I'm adding one more simple test in #100, but to do this properly the best would be to reproduce part or all of lalpulsar's [LineRobustStatsTest.c](https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalpulsar/test/LineRobustStatsTest.c) and maybe some bits from [testHierarchSearchGCT.sh](https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalapps/src/pulsar/GCT/testHierarchSearchGCT.sh) too.

Directly reproducing those lalsuite tests doesn't make sense, as we're just wrapping the lalpulsar function anyway. #245 adds some basic qualitative test cases instead.

dbkeitel:
Should be in a decent enough state for now. Any more detailed tests probably only make sense after/if implementing more direct tuning options in #260.

PR #100, opened 2020-07-27
fix semicoherent BSGL

dbkeitel:
Following up from #98:

 -should be computed from summed F-stats,
  not per segment and then summed itself
 -also fixed method names:
   get_semicoherent_twoF -> get_semicoherent_det_stat
  _get_per_segment_det_stat -> _get_per_segment_twoF
 -and made this one into a proper loop over arbitrary numDet,
  though rest of package doesn't fully support >2 yet

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1677,7 +1677,7 @@ class SemiCoherentSearch(ComputeFstat):
                 )
                 self.twoFX[X] = 2 * FSX.F_mn.data[0][0]
             log10_BSGL = lalpulsar.ComputeBSGL(twoF, self.twoFX, self.BSGLSetup)
-            ln_BSGL = log10_BSGL / np.log10(np.exp(1))
+            ln_BSGL = log10_BSGL * np.log(10.)
             if np.isnan(ln_BSGL):
                 logging.debug("NaNs in semi-coherent ln(BSGL) treated as zero")
                 ln_BSGL = 0.0
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -785,7 +785,7 @@ class ComputeFstat(BaseSearchClass):
             if len(self.injectSqrtSX) != len(self.detector_names):
                 raise ValueError(
                     "injectSqrtSX must be of same length as detector_names ({}!={})".format(
-                        len(self.injectSqrtSX), len(detector_names)
+                        len(self.injectSqrtSX), len(self.detector_names)
                     )
                 )
             FstatOAs.injectSqrtSX = lalpulsar.MultiNoiseFloor()
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1190,7 +1190,7 @@ class ComputeFstat(BaseSearchClass):
                 return twoF
 
         FstatResults_single = copy.copy(self.FstatResults)
-        FstatResults_single.lenth = 1
+        FstatResults_single.numDetectors = 1
         FstatResults_single.data = self.FstatResults.multiFatoms[0].data[0]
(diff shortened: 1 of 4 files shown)


dbkeitel:
One test added, black fixed.

@Rodrigo-Tenorio that's as far as I have time to take it for the moment, but #99 should stay open as a reminder for future test improvements. A quick look over these changes would be appreciated.

pyfstat/core.py
```diff
@@ -1649,57 +1649,46 @@ def get_semicoherent_twoF(
...
         if record_segments:
-            self.detStat_per_segment = det_stat_per_segment
-
-        return detStat
-
-    def _get_per_segment_det_stat(self):
-        FS = lalpulsar.ComputeTransientFstatMap(
-            self.FstatResults.multiFatoms[0], self.semicoherentWindowRange, False
-        )
+            self.detStat_per_segment = twoF_per_segment
```
Rodrigo-Tenorio, line 1664: Should we re-name this variable to self.twoF_per_segment since there's no BSGL per segment?

tests.py
```diff
@@ -424,6 +426,60 @@ def test_run_computefstatistic_single_point_injectSources(self):
...
+            tstart=self.minStartTime,
+            tend=self.maxStartTime,
+            F0=self.F0,
+            F1=self.F1,
+            F2=self.F2,
+            Alpha=self.Alpha,
+            Delta=self.Delta,
+        )
+        self.assertTrue(lnBSGL < 0)
+        # now with an added signal, expect lnBSGL<0
```
Rodrigo-Tenorio: Typo, should be lnBSGL>0

pyfstat/core.py
dbkeitel: I was thinking the same. It doesn't actually seem to be used anywhere in the package or examples, either, so I guess it won't break anything and is purely an undocumented debug feature anyway.

pyfstat/core.py
```diff
@@ -1649,57 +1649,46 @@ def get_semicoherent_twoF(
...
+            for X in range(self.FstatResults.numDetectors):
+                FstatResults_single = copy.copy(self.FstatResults)
+                FstatResults_single.numDetectors = 1
+                FstatResults_single.data = self.FstatResults.multiFatoms[0].data[X]
+                FSX = lalpulsar.ComputeTransientFstatMap(
+                    FstatResults_single.multiFatoms[0],
+                    self.semicoherentWindowRange,
+                    False,
+                )
+                self.twoFX[X] = 2 * FSX.F_mn.data[0][0]
```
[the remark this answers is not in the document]
dbkeitel: Yes, absolutely! That should have been the whole point of this PR... :facepalm:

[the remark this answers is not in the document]
dbkeitel: Please double-check latest version, I've also added another NaN check here similar to the main one because I'm not sure how ugly the `lalpulsar.ComputeBSGL()` failure would be on NaN inputs.

(It's actually a bit dirty to just set individual segments to 0 in both cases, because then the output will be a sufficiently sane-looking semicoherent stat; while if the whole stat were nan-to-0 converted then at least an MCMC sampler would always reject the point and in a grid output the points would be more suspicious. But I'm ok with it for the moment, since there is the debug message.)

pyfstat/core.py
dbkeitel: ...it was used in a test after all. Fixed now.

pyfstat/core.py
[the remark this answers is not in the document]
Rodrigo-Tenorio: Looks Ok to me

PR #103, opened 2020-07-29
MCMC: save twoF into ASCII .dat output file

dbkeitel:
Also introduces a _get_twoF_from_loglikelihood() helper method so the slightly arcane method of getting 2F values back from the log-likelihood doesn't need to be rewritten multiple times.


dbkeitel:
Seems to also work correctly for transient and binary examples, will merge.

Issue #106, opened 2020-07-29
More testing on covering frequency estimation required


Rodrigo-Tenorio:
Running the binary example with uniform priors in the sky position I get the following error.

It looks like none of the tests actually uses sky position priors.

Priors on use:
```
theta_prior = {
    "F0": signal_parameters["F0"],
    "F1": signal_parameters["F1"],
    "F2": signal_parameters["F2"],
    "Alpha": {
        "type": "unif",
        "lower": signal_parameters["Alpha"],
        "upper": signal_parameters["Alpha"] + 0.1,
    },
    "Delta": {
        "type": "unif",
        "lower": signal_parameters["Delta"],
        "upper": signal_parameters["Delta"] + 0.1,
    },
    "asini": {"type": "unif", "lower": 9.9, "upper": 10.1},
    "period": {
        "type": "unif",
        "lower": 44.99 * 24 * 3600.0,
        "upper": 45.01 * 24 * 3600.0,
    },
    "ecc": signal_parameters["ecc"],
    "tp": {"type": "unif", "lower": 0.999 * mid_time, "upper": 1.001 * mid_time,},
    "argp": signal_parameters["argp"],
}
```

```
17:14 INFO    : Initialising Fstat arguments
17:14 INFO    : [minCoverFreq,maxCoverFreq] not provided, trying to estimate from search ranges.
XLAL Error - buildFlatSkyGrid (DopplerScan.c:819): Check failed: (dAlpha > 0) && (dDelta > 0)
XLAL Error - buildFlatSkyGrid (DopplerScan.c:819): Invalid argument
XLAL Error - XLALInitDopplerSkyScan (DopplerScan.c:200): Check failed: (skyScan->skyGrid = buildFlatSkyGrid( &(skyScan->skyRegion), init->dAlpha, init->dDelta)) != ((void *)0)
XLAL Error - XLALInitDopplerSkyScan (DopplerScan.c:200): Internal function call failed: Invalid argument
XLAL Error - XLALInitFactoredGrid (DopplerFullScan.c:326): Check failed: XLALInitDopplerSkyScan ( &(fscan->skyScan), &skyScanInit) == XLAL_SUCCESS
XLAL Error - XLALInitFactoredGrid (DopplerFullScan.c:326): Internal function call failed: Invalid argument
[...]

PR #107, opened 2020-07-29
expose and fix bug in estimate_min_max_CoverFreq for non-fixed sky locations

dbkeitel:
First commit splits the MCMCSearch test case into two: 
 - test_fully_coherent_MCMC_fixedsky
 - test_fully_coherent_MCMC_skyprior
 - the latter exposes bug from https://github.com/PyFstat/PyFstat/issues/106

@Rodrigo-Tenorio will look into a fix

### DIFF
--- a/pyfstat/mcmc_based_searches.py
+++ b/pyfstat/mcmc_based_searches.py
@@ -235,8 +235,8 @@ class MCMCSearch(core.BaseSearchClass):
                 )
             # first start with parameters that have non-delta prior ranges
             search_ranges = {
-                key: [prior_bounds[key]["lower"], prior_bounds[key]["upper"], 0]
-                for key in prior_bounds.keys()
+                key: [bound["lower"], bound["upper"]]
+                for key, bound in prior_bounds.items()
             }
             # then add fixed-point (delta prior) parameters
             for key in self.theta_prior:
--- a/tests.py
+++ b/tests.py
@@ -784,8 +784,16 @@ class MCMCSearch(Test):
             "F0": {"type": "norm", "loc": self.F0, "scale": np.abs(1e-10 * self.F0)},
             "F1": {"type": "norm", "loc": self.F1, "scale": np.abs(1e-10 * self.F1)},
             "F2": self.F2,
-            "Alpha": {"type": "norm", "loc": self.Alpha, "scale": 0.01},
-            "Delta": {"type": "norm", "loc": self.Delta, "scale": 0.02},
+            "Alpha": {
+                "type": "unif",
+                "lower": self.Alpha - 0.01,
+                "upper": self.Alpha + 0.01,
+            },
+            "Delta": {
+                "type": "unif",
+                "lower": self.Delta - 0.01,
+                "upper": self.Delta + 0.01,
+            },
         }
 
         search = pyfstat.MCMCSearch(
--- a/tests.py
+++ b/tests.py
@@ -697,54 +697,105 @@ class SemiCoherentGlitchSearch(Test):
 
 class MCMCSearch(Test):
     label = "TestMCMCSearch"
+    h0 = 1
+    sqrtSX = 1
+    F0 = 30
+    F1 = -1e-10
+    F2 = 0
+    minStartTime = 700000000
+    duration = 1 * 86400
+    maxStartTime = minStartTime + duration
+    Alpha = 5e-3
+    Delta = 1.2
+    tref = minStartTime
 
-    def test_fully_coherent(self):
-        h0 = 1
-        sqrtSX = 1
-        F0 = 30
-        F1 = -1e-10
-        F2 = 0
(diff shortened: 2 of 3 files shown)


Rodrigo-Tenorio:
If I'm not mistaken, it looks like the problem was a step defined as 0 when it should have really been `upper - lower`.

dbkeitel:
Thanks, you've found the source of the problem, but I think the right solution is to just drop that third entry in the list - `ComputeFstat.estimate_min_max_CoverFreq()` will accept `either [single_value], [min,max] or [min,max,step]` and so you're now basically telling it the MCMC has a "step size" of the whole covered band. But anyway the step size is not needed for the coverage band estimation, so we can just fall back on the default.

I'll push a new version.

pyfstat/mcmc_based_searches.py
```diff
@@ -235,8 +235,8 @@ def _get_search_ranges(self):
                 )
             # first start with parameters that have non-delta prior ranges
             search_ranges = {
-                key: [prior_bounds[key]["lower"], prior_bounds[key]["upper"], 0]
```
dbkeitel, line 238: Thanks, you've found the source of the problem, but I think the right solution is to just drop that third entry in the list - `ComputeFstat.estimate_min_max_CoverFreq()` will accept `either [single_value], [min,max] or [min,max,step]` and so you're now basically telling it the MCMC has a "step size" of the whole covered band. But anyway the step size is not needed for the coverage band estimation, so we can just fall back on the default.

I'll push a new version.

pyfstat/mcmc_based_searches.py
```diff
@@ -235,8 +235,8 @@ def _get_search_ranges(self):
...
-                for key in prior_bounds.keys()
+                key: [bound["lower"], bound["upper"], bound["upper"] - bound["lower"]]
+                for key, bound in prior_bounds.items()
```
dbkeitel, line 239: Thanks, you've found the source of the problem, but I think the right solution is to just drop that third entry in the list - `ComputeFstat.estimate_min_max_CoverFreq()` will accept `either [single_value], [min,max] or [min,max,step]` and so you're now basically telling it the MCMC has a "step size" of the whole covered band. But anyway the step size is not needed for the coverage band estimation, so we can just fall back on the default.

I'll push a new version.

dbkeitel:
I've squashed my version into your commit (but as I had hoped in adding an extra in-line comment, the previous suggested solution is still preserved as a "resolved conversation" above).

Rerunning the test case a few times, I also realised Gaussian priors on the sky are a bad idea, since the walkers can jump out of the allowed delta range. Changed to uniform.

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: That actually makes sense.

tests.py
```diff
@@ -697,54 +697,113 @@ def test_get_semicoherent_nglitch_twoF(self):
...
         predicted_FS = Writer.predict_fstat()
 
         theta = {
-            "F0": {"type": "norm", "loc": F0, "scale": np.abs(1e-10 * F0)},
-            "F1": {"type": "norm", "loc": F1, "scale": np.abs(1e-10 * F1)},
-            "F2": F2,
-            "Alpha": Alpha,
-            "Delta": Delta,
+            "F0": {"type": "norm", "loc": self.F0, "scale": np.abs(1e-10 * self.F0)},
+            "F1": {"type": "norm", "loc": self.F1, "scale": np.abs(1e-10 * self.F1)},
```
Rodrigo-Tenorio, line 735: I thought you said uniform? Or do you do this in order to test the normalization warning as well?

tests.py
```diff
@@ -697,54 +697,113 @@ def test_get_semicoherent_nglitch_twoF(self):
...
+        )
+        Writer.make_data()
+
+        predicted_FS = Writer.predict_fstat()
+
+        theta = {
+            "F0": {"type": "norm", "loc": self.F0, "scale": np.abs(1e-10 * self.F0)},
+            "F1": {"type": "norm", "loc": self.F1, "scale": np.abs(1e-10 * self.F1)},
+            "F2": self.F2,
+            "Alpha": {
```
Rodrigo-Tenorio, line 787: This may put us in some trouble if Alpha or Delta are close to their edges (it already happened to me); however, I'm willing to believe 0.01 is small enough so this never happens _almost surely_.

dbkeitel: Yes doing it that way could cause trouble in general, but the test suite is set up with globally
```
        self.Alpha = 5e-3
        self.Delta = 1.2
```
so it's safe here.

tests.py
dbkeitel: I changed Alpha,Delta to uniform in the "skyprior" case below, but left F0,F1 alone in both.

PR #109, opened 2020-08-04
Create .zenodo.json

dbkeitel:
- Zenodo can automatically pull version-independent metadata from this file for each release

- documentation: https://developers.zenodo.org/ (most of the page is aimed at the interactive requests API though)

- I've tested this with a [test repo](https://github.com/dbkeitel/test-zenodo-json) which gets pulled to a [Zenodo sandbox record](https://sandbox.zenodo.org/record/655360)

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -34,5 +34,14 @@
   ],
   "access_right": "open",
   "license": "MIT",
-  "upload_type": "software"
+  "upload_type": "software",
+  "language": "eng",
+  "related_identifiers": [
+        {
+            "scheme": "doi", 
+            "identifier": "10.5281/zenodo.1243930", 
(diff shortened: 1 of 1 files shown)


dbkeitel:
I had also put in an entry
```
  "related_identifiers": [
        {
            "scheme": "doi", 
            "identifier": "10.5281/zenodo.1243930", 
            "relation": "isNewVersionOf", 
            "resource_type": "software"
        }
    ]
```
but, going by the test repo results, the problem with that is that it then no longer picks up other `related_identifiers` automatically, i.e. not the link back to github. And since that one needs to be versioned, I can't hardcode it in here - an annoying circular design flaw. So I'll just leave this one off, as the need to navigate back from the new to the old Zenodo listings should be rare at best.

PR #116, opened 2020-08-06
MCMC tests: check if inj params within mean+-2std and within 90%

### LINKED ISSUE #112: MCMC tests randomly fail my (vaguely guessed) tolerance thresholds
The new MCMC tests from #113 ran fine the first few times, but actually fail unreasonably often when the parameter recovery happens to be randomly off-tolerance. Since, to keep the test runtime bounded, the runs are kept very short and unlikely to be perfectly converged even for the tiny test boxes, this shouldn't have come as a surprise to me...

I'm afraid the clean solution will be to test if mean+-stdev (or 90% intervals) cover the injection value, instead of the maxF point being within some guessed tolerance.

dbkeitel:
resolves #112 (I hope)

Still cannot be **guaranteed** to always succeed, but should be much more reliable than fixed tolerances.

If it fails again more often than I'd hope, we can try fixing "nice" random seeds.

### DIFF
--- a/pyfstat/mcmc_based_searches.py
+++ b/pyfstat/mcmc_based_searches.py
@@ -1804,9 +1804,15 @@ class MCMCSearch(core.BaseSearchClass):
             d[k] = {}
             d[k]["mean"] = np.mean(s)
             d[k]["std"] = np.std(s)
-            d[k]["lower90"], d[k]["median"], d[k]["upper90"] = np.quantile(
-                s, [0.05, 0.5, 0.95]
-            )
+            (
+                d[k]["lower99"],
+                d[k]["lower90"],
+                d[k]["lower50"],
+                d[k]["median"],
+                d[k]["upper50"],
+                d[k]["upper90"],
+                d[k]["upper99"],
+            ) = np.quantile(s, [0.005, 0.05, 0.25, 0.5, 0.75, 0.95, 0.995])
 
         return d
 
--- a/tests.py
+++ b/tests.py
@@ -719,7 +719,54 @@ class SemiCoherentGlitchSearch(Test):
         self.assertTrue(np.abs((FS - predicted_FS)) / predicted_FS < 0.3)
 
 
-class MCMCSearch(Test):
+class MCMCSearchTest(Test):
+    def _check_mcmc_quantiles(self, search, max_dict, writer, transient=False):
+        summary_stats = search.get_summary_stats()
+        nsigmas = 3
+        conf = "99"
+
+        if not transient:
+            inj = {k: getattr(writer, k) for k in max_dict}
+        else:
+            inj = {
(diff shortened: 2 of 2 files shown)


dbkeitel:
Tried a few hacks to get specifically the most failing-prone cases safer, but it all wasn't enough.

Now:

- only check injection parameters are within mean+-3std and within 99%, which for the narrow test case priors isn't really impressive at all, but better than nothing
- use a fixed random seed for data generation; this way the only run-to-run randomness is in the MCMC itself, which is much less than that in noise realisations

So it still cannot be guaranteed to always work, but after a lot of false starts I hope it is now reasonably robust...

pyfstat/mcmc_based_searches.py
```diff
@@ -1804,9 +1804,15 @@ def get_summary_stats(self):
...
-            )
+            (
+                d[k]["lower99"],
+                d[k]["lower90"],
+                d[k]["lower50"],
+                d[k]["median"],
+                d[k]["upper50"],
+                d[k]["upper90"],
+                d[k]["upper99"],
+            ) = np.quantile(s, [0.005, 0.05, 0.25, 0.5, 0.75, 0.95, 0.995])
```
Rodrigo-Tenorio, line 1815: I'm not sure if I get this: Why e.g. `lower99` goes with 0.005?

tests.py
```diff
@@ -723,7 +723,6 @@ class MCMCSearch(Test):
     label = "TestMCMCSearch"
 
     def test_fully_coherent_MCMC(self):
-
```
Rodrigo-Tenorio, line 726: Let me have a quick look on this

pyfstat/mcmc_based_searches.py
dbkeitel: symmetric 99% interval = 99% contained around the median -> [0.005,0.995] with half a percent left out either side

Rodrigo-Tenorio: Of course, that makes sense.

tests.py
```diff
@@ -67,6 +67,53 @@ def tearDownClass(self):
                 logging.warning("{} not removed prior to tests".format(self.outdir))
 
 
+class MCMCSearchTest(Test):
```
Rodrigo-Tenorio, line 70: @dbkeitel opinions?

dbkeitel: Yes, good idea! The whole test suite could be cleaned up with more inheritance, but the again, it seems it should be migrated from an unholy unittest-pytest mix to pure pytest later, so it's not worth much more effort at this point I'd say.

Could you move this to further down though, just before the first child class (`class MCMCSearch(MCMCSearchTest)`)?

Rodrigo-Tenorio: Done

Issue #84, opened 2020-07-13
Improve handling of timestamps when noiseSFTs is present


Rodrigo-Tenorio:
Currently, [Writer](https://github.com/PyFstat/PyFstat/blob/master/pyfstat/make_sfts.py#L27) accepts `noiseSFTs` as a way of adding noise to the injections; however, timestamp information (such as time start, duration and so on) still has to be given by hand.

It would be nice to get that information from the noiseSFTs themselves no-questions-asked. This may remove a degree of freedom related to specify signals shorter than the actual data time span, but I'd say transient windows already cover that.

dbkeitel:
:+1: in principle, though the transient part needs a careful look, since in some places there's some interference between the general start/endtime and the actual windowing options.

I suggest you make a draft PR when you have time, ignoring transients to start with, and I can check the effect on transients.

dbkeitel:
After #90/#91 this should be easier since I'm disentangling the transient situation.

dbkeitel:
> +1 in principle, though the transient part needs a careful look, since in some places there's some interference between the general start/endtime and the actual windowing options.
> 
> I suggest you make a draft PR when you have time, ignoring transients to start with, and I can check the effect on transients.

After #91 the interference with transients should be gone, so this could be quite easy to do now.

Issue #71, opened 2020-07-02
tests: "not a code object" error when Writer.make_data() fails in generic setUpClass


dbkeitel:
The dread
```
INTERNALERROR> TypeError: not a code object: <class 'pyfstat.make_sfts.Writer'>
```
that happened due to a broken lalsuite release in #52 is actually very easy to reproduce: just have a permission issue on the MFDv5 executable, or introduce a typo in its commandline in `Writer.run_makefakedata()`, and the same happens again.

The problem is somehow with how the generic `class Test(unittest.TestCase)` at the start of `test.py` calls the `Writer` object. It doesn't seem to be a question of error catching per se, as adding additional try-excepts, replacing the `Writer.make_data()` in L54 with separate `make_cff()` and `run_makefakedata()`, etc do not solve it.  Rather it seems `pytest` specifically dislikes any subprocess error happening in the `setUpClass`.

So either a clever solution can be researched somehow, or we need to get rid of that generic `setUpClass` construct.

PR #117, opened 2020-08-07
make_sfts.py/Split setup depending on noiseSFTs

Rodrigo-Tenorio:
Solve #84 by splitting the input data handling in two cases depending on whether `noiseSFTs` was given or not.

If the former, use `lalpulsar` tools to infer as much information from them as possible; if the latter, do what's been always done.

I'll have to think about how to deal with `tstart` and so on: I could ignore them if `noiseSFTs` are give, or change the default value to some `None` in order to prevent them from being used/confused (although that would probably imply dealing with the test suite and the infamous #71)

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -31,8 +31,8 @@ class Writer(BaseSearchClass):
     def __init__(
         self,
         label="Test",
-        tstart=700000000,
-        duration=100 * 86400,
+        tstart=None,
+        duration=None,
         tref=None,
         F0=30,
         F1=1e-10,
@@ -73,10 +73,13 @@ class Writer(BaseSearchClass):
             frequency, sky-position, and amplitude parameters
         Tsft: float
             the sft duration
+        noiseSFTs: str
+            SFT on top of which signals will be injected. 
+            If not None, additional constraints can be applied using the arguments 
+            tstart and duration.
         minStartTime, maxStartTime: float
             DEPRECATED, use [tstart,duration] and/or
-            [transientWindowType,transientStartTime,transientTau] instead!
-
+            [transientWindowType,transientStartTime,transientTau] instead!    
         see `lalapps_Makefakedata_v5 --help` for help with the other paramaters
         """
 
@@ -92,31 +95,44 @@ class Writer(BaseSearchClass):
         self.tbounds = [self.tstart, self.tend]
         logging.info("Using segment boundaries {}".format(self.tbounds))
 
-    def basic_setup(self):
+    def _get_sft_constraints_from_tstart_duration(self):
+        """
+        Use start and duration to set up a lalpulsar.SFTConstraints
+        object. This method is only used if noiseSFTs is not None.
+        """
+        SFTConstraint = lalpulsar.SFTConstraints()
+
+        if self.tstart is None:
+            SFTConstraint.minStartTime = None
+            SFTConstraint.maxStartTime = None
+        elif self.duration is None:
+            SFTConstraint.maxStartTime = None
+        else:
+            SFTConstraint.minStartTime = lal.LIGOTimeGPS(self.tstart)
+            SFTConstraint.maxStartTime = SFTConstraint.minStartTime + self.duration
+
+        SFTConstraint.timestamps = None  # FIXME: not currently supported
+
+        logging.info(
+            "SFT Constraints: [minStartTime:{}, maxStartTime:{}]".format(
+                SFTConstraint.minStartTime, SFTConstraint.maxStartTime,
+            )
+        )
+
+        return SFTConstraint
+
+    def _get_setup_from_tstart_duration(self):
+        """
+        Default behavior: If no noiseSFTs are given, use the input parameters (tstart, 
+        duration, detectors and Tsft) to make fake data.
+        """
         self.tstart = int(self.tstart)
         self.duration = int(self.duration)
         self.tend = self.tstart + self.duration
 
         IFOs = self.detectors.split(",")
-        if self.noiseSFTs:
-            numSFTs = []
-            for ifo in IFOs:
-                constraint = lalpulsar.SFTConstraints()
-                constraint.detector = ifo
-                constraint.minStartTime = lal.LIGOTimeGPS(self.tstart)
-                constraint.maxStartTime = lal.LIGOTimeGPS(self.tend)
-                catalog = lalpulsar.SFTdataFind(self.noiseSFTs, constraint)
-                numSFTs.append(catalog.length)
-        else:
-            numSFTs = len(IFOs) * [int(float(self.duration) / self.Tsft)]
+        numSFTs = len(IFOs) * [int(float(self.duration) / self.Tsft)]
 
-        self.theta = np.array([self.phi, self.F0, self.F1, self.F2])
-
-        if os.path.isdir(self.outdir) is False:
-            os.makedirs(self.outdir)
-        if self.tref is None:
-            self.tref = self.tstart
-        self.config_file_name = os.path.join(self.outdir, self.label + ".cff")
         self.sftfilenames = [
             lalpulsar.OfficialSFTFilename(
                 dets[0],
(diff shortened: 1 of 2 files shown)


pyfstat/make_sfts.py
```diff
@@ -134,6 +122,55 @@ def basic_setup(self):
...
+        self.IFOs = ",".join(['"{}"'.format(d) for d in IFOs])
+
+        # Split noiseSFTs to account for multiple sets at once
+        noise_sft_sets = self.noiseSFTs.split(";")
+        self.sftfilepath = ""
+        for sft_set in noise_sft_sets:
+            self.sftfilepath += ";".join(glob.glob(sft_set)) + ";"
+        self.sftfilepath = self.sftfilepath[:-1]
+
+        self.sftfilenames = [
```
Rodrigo-Tenorio, line 157: Ok, my mistake: This does NOT refer to the noise SFTs, but to the output produced by MFD, which uses the noiseSFTs.

Makes total sense!

Rodrigo-Tenorio: SFT name is constructed in the same way as [MFD_V5] (loop over ifos, longest time span per each ifo).

I'll have to change some tests to make sure it works

pyfstat/make_sfts.py
```diff
@@ -129,10 +117,78 @@ def basic_setup(self):
...
+        else:
+            self.Tsft = Tsft[0]
+
+    def basic_setup(self):
+        os.makedirs(self.outdir, exist_ok=True)
+        self.config_file_name = os.path.join(self.outdir, self.label + ".cff")
+        self.theta = np.array([self.phi, self.F0, self.F1, self.F2])
+
+        if self.noiseSFTs is not None:
+            logging.warning(
```
Rodrigo-Tenorio, line 179: @dbkeitel opinion on how stronger this warning should be?

Would you set "0-like" default values for everything?

pyfstat/make_sfts.py
```diff
@@ -85,38 +87,24 @@ def __init__(
...
+
         self.set_ephemeris_files()
         self.basic_setup()
         self.calculate_fmin_Band()
 
         self.tbounds = [self.tstart, self.tend]
         logging.info("Using segment boundaries {}".format(self.tbounds))
 
-    def basic_setup(self):
+    def _get_setup_from_input_data(self):
```
dbkeitel, line 100: Let's call this 
```suggestion
    def _get_setup_from_tstart_duration(self):
```
because until I was half-way through the function, I thought "data" meant files. ;)

pyfstat/make_sfts.py
```diff
@@ -129,10 +117,78 @@ def basic_setup(self):
             )
             for ind, dets in enumerate(IFOs)
         ]
+        self.IFOs = ",".join(['"{}"'.format(d) for d in IFOs])
```
dbkeitel, line 120: isn't this just the inverse of `IFOs = self.detectors.split(",")` above?

pyfstat/make_sfts.py
```diff
@@ -129,10 +117,78 @@ def basic_setup(self):
...
+
+        # Information to be extracted from the SFTs themselves
+        IFOs = []
+        tstart = []
+        tend = []
+        Tsft = []
+        self.sftfilenames = []  # This refers to the MFD output!
+
+        # SWIG-LAL does not import this functionality
+        gps_to_float = lambda x: x.gpsSeconds
```
dbkeitel, line 135: Shouldn't this be
```suggestion
        gps_to_int = lambda x: x.gpsSeconds
```
? In the no-noisesfts case, self.tstart and self.duration are also explicitly cast to int.

pyfstat/make_sfts.py
```diff
@@ -129,10 +117,78 @@ def basic_setup(self):
...
+                    this_Tsft,
+                    this_start_time,
+                    this_end_time - this_start_time + this_Tsft,
+                    self.label,
+                )
+            )
+
+            IFOs.append(ifo_name)
+            tstart.append(this_start_time)
+            tend.append(this_end_time)
```
dbkeitel, line 159:
```suggestion
            tend.append(this_end_time+ this_Tsft)
```
for consistency with the other case? (please double-check)

pyfstat/make_sfts.py
```diff
@@ -129,10 +117,78 @@ def basic_setup(self):
...
+            Tsft.append(this_Tsft)
+
+        # Get the "overall" values of the search
+        self.tstart = min(tstart)
+        self.tend = max(tend)
+        self.duration = self.tend - self.tstart + this_Tsft
```
dbkeitel, line 165:
```suggestion
        self.duration = self.tend - self.tstart + self.Tsft
```

pyfstat/make_sfts.py
```diff
@@ -129,10 +117,78 @@ def basic_setup(self):
...
+        self.IFOs = ",".join(['"{}"'.format(d) for d in IFOs])
+        Tsft = np.unique(Tsft)
```
dbkeitel, line 167: move these 5 lines about Tsft to the start of the block, so my comment above this one works (doesn't make a practical difference, but makes the logic clearer)

pyfstat/make_sfts.py
dbkeitel: I think this only has to be a warning if the other options are explicitly set - so yes, making them `None` by default would make sense.

However I'm wondering if it would also be good to be able to use noiseSFTs but still override tstart, duration and detectors with the existing keywords, by having PyFstat assemble the constraints itself.

pyfstat/make_sfts.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         outdir=".",
         sqrtSX=0.0,
         noiseSFTs=None,
+        SFTconstraint=None,
```
dbkeitel, line 50: maybe we can assemble that internally instead - see reply below

pyfstat/make_sfts.py
```diff
@@ -97,12 +97,12 @@ def __init__(
         self.tbounds = [self.tstart, self.tend]
         logging.info("Using segment boundaries {}".format(self.tbounds))
 
-    def _get_setup_from_input_data(self):
+    def _get_setup_from_tstart_duration(self):
         self.tstart = int(self.tstart)
         self.duration = int(self.duration)
-        self.tend = self.tstart + self.duration
+        self.tend = self.tstart + self.duration + self.Tsft
```
dbkeitel, line 103: Now I'm getting a bit confused myself, but I think here it must be just
```suggestion
        self.tend = self.tstart + self.duration
```
without the extra Tsft because the `tstart` and `duration` come straight from human input who would e.g. say `duration=86400` if they actually mean 1 day and not 23.5 hours.

pyfstat/make_sfts.py
```diff
@@ -132,14 +131,14 @@ def _get_setup_from_noiseSFTs(self):
...
+        gps_to_int = lambda x: x.gpsSeconds
 
         for ifo_catalog in noise_multi_sft_catalog.data:
             ifo_name = lalpulsar.ListIFOsInCatalog(ifo_catalog).data[0]
 
             time_stamps = lalpulsar.TimestampsFromSFTCatalog(ifo_catalog)
-            this_start_time = gps_to_float(time_stamps.data[0])
-            this_end_time = gps_to_float(time_stamps.data[-1])
+            this_start_time = gps_to_int(time_stamps.data[0])
+            this_end_time = gps_to_int(time_stamps.data[-1])
```
dbkeitel, line 141: Here `this_end_time` is a "last SFT starttime" as per internal lalpulsar convention...

pyfstat/make_sfts.py
```diff
@@ -156,19 +155,19 @@ def _get_setup_from_noiseSFTs(self):
 
             IFOs.append(ifo_name)
             tstart.append(this_start_time)
-            tend.append(this_end_time)
+            tend.append(this_end_time + this_Tsft)
```
dbkeitel, line 158: ...here Tsft is added, meaning `tend` becomes an "actual endtime"...

pyfstat/make_sfts.py
```diff
@@ -156,19 +155,19 @@ def _get_setup_from_noiseSFTs(self):
...
-        self.duration = self.tend - self.tstart + this_Tsft
-        self.IFOs = ",".join(['"{}"'.format(d) for d in IFOs])
         Tsft = np.unique(Tsft)
         if len(Tsft) != 1:
             raise ValueError("SFTs contain different basetimes: {}".format(Tsft))
         else:
             self.Tsft = Tsft[0]
+        self.tstart = min(tstart)
+        self.tend = max(tend)
+        self.duration = self.tend - self.tstart + self.Tsft
```
dbkeitel, line 169: ...then here the `Tsft` is doubled up?
```suggestion
        self.duration = self.tend - self.tstart
```

Rodrigo-Tenorio: Yes, definitely

pyfstat/make_sfts.py
Rodrigo-Tenorio: Ok, that makes sense

pyfstat/make_sfts.py
```diff
@@ -76,7 +76,8 @@ def __init__(
         minStartTime, maxStartTime: float
             DEPRECATED, use [tstart,duration] and/or
             [transientWindowType,transientStartTime,transientTau] instead!
-
+        noiseSFTs: str
+            Will be expanded using glob
```
dbkeitel, line 80:
```suggestion
            Will be expanded using glob.
            Additional constraints can be applied with the arguments
            tstart, duration and detectors.
```

pyfstat/make_sfts.py
```diff
@@ -92,31 +93,47 @@ def __init__(
...
         logging.info("Using segment boundaries {}".format(self.tbounds))
 
-    def basic_setup(self):
+    def _get_sft_constraints_from_tstart_duration(self):
+        """
+        Use detectors, tstart and duration to set up a lalpulsar.SFTConstraints
+        object. This method is only used if noiseSFTs is not None.
+        """
+        SFTConstraint = lalpulsar.SFTConstraints()
+        SFTConstraint.detector = self.detectors
```
dbkeitel, line 102: Will this fail when `self.detectors` is a list? https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_s_f_tfile_i_o_8h_source.html#l00225 seems to only allow single-IFO constraints.

pyfstat/make_sfts.py
```diff
@@ -92,31 +93,47 @@ def __init__(
...
+
+        if self.tstart is None:
```
dbkeitel, line 104: double-checked: [XLALCWGPSinRange](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tfile_i_o__h.html#gaca1afa637e9523e43ac631271cc34418) allows for either `minStartTime` or `maxStartTime` to be Null, so having only tstart but not `duration` here (the `elif` case) is indeed legal. Only setting `maxStartTime` would be possible too, but doesn't make sense with our `(tstart,duration)` input tuple.
So this looks allright!

pyfstat/make_sfts.py
```diff
@@ -129,10 +146,92 @@ def basic_setup(self):
...
+
+    def basic_setup(self):
+        os.makedirs(self.outdir, exist_ok=True)
+        self.config_file_name = os.path.join(self.outdir, self.label + ".cff")
+        self.theta = np.array([self.phi, self.F0, self.F1, self.F2])
+
+        if self.noiseSFTs is not None:
+            logging.warning(
+                "noiseSFTs is not None: Inferring tstart, duration, tend, Tsft. "
+                "Input tstart, duration, Tsft and detectors will be treated "
```
dbkeitel, line 222: Technically Tsft is now not used in the constraint, but it is sanity-checked, so I guess this is ok.

dbkeitel:
Rebased and squashed, extended the test to explicitly check with/without constraints.

By the way, in the end our whole constraint handling is a bit over the top because we eventually fall back to a MFDv5 commandline with --startTime and --duration after all, but it seems to work consistently now and is a better basis for at a later point to switch to natively calling the lalpulsar functions without the commandline detour.

Issue #83, opened 2020-07-13
add test case for generalised Nsfts counting in Writer


dbkeitel:
@Rodrigo-Tenorio as discussed, a simple unit test to check the issue fixed in #82 would be great:
Doesn't need to be your practical overlapping-SFTs case, the same failure mode should get triggered by a non-contiguous set -  i.e. three calls to Writer:

1.    generate [T0,T0+duration]
1. [T0+duration+Tgap,T0+2*duration+Tgap]
1. try to inject

The `Writer.test_noise_sfts` test case should be easy to amend for this.

PR #119, opened 2020-08-10
tests.py/Include test case using data with gaps

Rodrigo-Tenorio:
After #83, `noiseSFTs` are handled in a proper way (meaning things like #82 are  trivial to fix/modify since they are practically a method on their own.

This PR simply ensures that data with gaps (or non-contiguous data) is properly handled both by MFD and PyFstat (which needs to set the proper `sftfilepath` by counting the number of SFTs).

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -253,6 +253,60 @@ class Writer(Test):
         # we've used half the data, so expect half the FS
         self.assertTrue(np.abs(FS_1 - 2 * FS_3) / FS_1 < 0.05)
 
+    def test_data_with_gaps(self):
+        duration = 10 * self.Tsft
+        gap_time = 4 * self.Tsft
+        window = "tukey"
+        window_beta = 0.01
+        detectors = "H1"
+
+        first_chunk_of_data = self.writer_class_to_test(
+            label="first_chunk_of_data",
+            outdir=self.outdir,
+            duration=duration,
+            Tsft=self.Tsft,
+            tstart=self.tstart,
+            detectors=detectors,
+            SFTWindowType=window,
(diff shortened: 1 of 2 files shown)


Rodrigo-Tenorio:
Also change a default value which should have been included in #84

Rodrigo-Tenorio:
>If you were to actually compare the data (or run CFS on it), I think it won't quite do what I infer you want it to do from fixing the random seed - i.e. the second half of the both_chunks_of_data output will probably be different from the second_chunk_of_data output.

Yes, that's true, the random seed is just there because I looked into the previous test case. If you think that misleads the reader, I'm Ok removing it as it is.

>Do we actually need to call make_data() here? sftfilepath is already filled in basic_setup(). Then again, explicitly checking that nothing goes wrong in the actual run is not a bad idea, and more detailed checking (of the actual output data) can still be added later.

That's my mistake: I intended to test the expected output file name to the one generated by MFD. Correction incoming.

tests.py
```diff
@@ -253,6 +253,60 @@ def test_noise_sfts(self):
...
+        Tsft = both_chunks_of_data.Tsft
+        total_duration = 2 * duration + gap_time
+        Nsft = int((total_duration - gap_time) / Tsft)
+        expected_SFT_filepath = (
+            self.outdir
+            + "/H-{}_H1_{}SFT_both_chunks_of_data-{}-{}.sft".format(
+                Nsft, Tsft, self.tstart, total_duration
+            )
+        )
+        self.assertTrue(os.path.isfile(expected_SFT_filepath))
```
dbkeitel, line 308: Ok good, that's a slightly stronger test now!

PR #121, opened 2020-08-10
Writer: auto-estimate Band if None

dbkeitel:
Still need to fix some of the tests, but this makes it so that, as per the doc string:
```
        Band: float or None
            If float, output SFTs cover [F0-Band/2,F0+Band/2].
            If None, a minimal covering band for a perfectly-matched
            single-template ComputeFstat analysis is estimated.
```

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -227,8 +227,10 @@ class Writer(BaseSearchClass):
                 "internal consistency accross input SFTs."
             )
             self._get_setup_from_noiseSFTs()
-        else:
+        elif self.tstart is not None and self.duration is not None:
             self._get_setup_from_tstart_duration()
+        else:
+            raise ValueError("Need either noiseSFTs or both of (tstart,duration).")
 
         self.sftfilepath = ";".join(
             [os.path.join(self.outdir, fn) for fn in self.sftfilenames]
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -49,7 +49,7 @@ class Writer(BaseSearchClass):
         noiseSFTs=None,
         SFTWindowType=None,
         SFTWindowBeta=0.0,
-        Band=4,
+        Band=None,
         detectors="H1",
         minStartTime=None,
         maxStartTime=None,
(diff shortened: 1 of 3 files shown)


dbkeitel:
Now also includes a small error message fixup to #117 and a significant test suite improvement:

- rename all classes to Test* (matching the first example given at https://docs.python.org/3/library/unittest.html and reducing risk of confusion)
- use base classes for joint setup/methods of all Grid/MCMC variants (this is the reason it's in this PR, makes the next point much easier)
- manually fix SFT generation bands where required (fixup to the failures from the 2nd commit)

pyfstat/make_sfts.py
```diff
@@ -373,7 +379,50 @@ def make_cff(self):
...
+                lalpulsar.FstatOptionalArgsDefaults.Dterms
+                + int(lalpulsar.FstatOptionalArgsDefaults.runningMedianWindow / 2)
+                + 1
+            )
+            logging.info(
+                "Estimating required SFT frequency range from properties"
+                " of signal to inject plus {:d} extra bins either side"
+                " (corresponding to default F-statistic settings).".format(extraBins)
+            )
+            minCoverFreq, maxCoverFreq = helper_functions.get_covering_band(
```
Rodrigo-Tenorio, line 405: I propose to put this bit in a separate method such that Isolated and Binary writers are decoupled.

E.g.

```
class Writer:

    def _calculate_Band_aux(self):
         return helper_functions.get_covering_band(
                tref=self.tref,
                tstart=self.tstart,
                tend=self.tend,
                F0=self.F0,
                F1=self.F1,
                F2=self.F2,
                F0band=0.0,
                F1band=0.0,
                F2band=0.0,
                maxOrbitAsini=0,
                minOrbitPeriod=0,
                maxOrbitEcc=0,
            )

class BinaryModulatedWriter:

    def _calculate_Band_aux(self):
        return helper_functions.get_covering_band(
                tref=self.tref,
                tstart=self.tstart,
                tend=self.tend,
                F0=self.F0,
                F1=self.F1,
                F2=self.F2,
                F0band=0.0,
                F1band=0.0,
                F2band=0.0,
                maxOrbitAsini=self.asini
                minOrbitPeriod=self.period
                maxOrbitEcc=self.ecc
            )
```

dbkeitel: If `get_covering_band()` had variable arguments (kwargs) for many different types of signals, I'd definitely agree. But since both it and the underlying lalpulsar function have a fixed set of arguments (fkdot+binary), I don't see a reason any future Writer children would need to deal with it much differently, and so I'd say the current getattr-with-default solution rather saves some code duplication...?

Rodrigo-Tenorio: I'm going to say "Ok" since the underlying lal-based machinery is quite unlikely to change enough in the foreseeable future as to break this bit of code.

PR #125, opened 2020-08-14
Writer simplifications and tests overhaul

dbkeitel:
This started out as just the two open items from #90:

> Several tests manipulate Writer.tstart, Writer.duration and Writer.tend manually after the object has been set up, to then re-use it for a second data generation, and this seems like something real users may also be tempted to do. But it is potentially dangerous since the 3 variables really only have 2 degrees of freedom. A joint update_data_duration(), or so, method should be provided.

--> instead, in faea91298bee488762a4e3d0e9149ad6ec60932f I implemented `.tend()` as a method and no static value for it is stored. This should be the safest way.

> Check if some more defaults can be set to None instead.

--> The answer is YES, but trying it out caused an avalanche of test failures, many just due to bad inheritance of default values from the Frankenstein way the test suite is set up, some from small issues elswhere in the code.

@Rodrigo-Tenorio I've already squashed a few evolutionary stages of this branch, but for the remaining lot of commits I suggest to click them through one by one focusing on the library changes and ignoring anything in `tests.py`, then only looking at that file in the overall diff at the end. Because I went back and forth with some things like default setups, but don't want to squash it all completely because some of the intermediate bits may be worth resurrecting later if new issues crop up with my latest solutions. Also I of course claim no optimality for any of it.


pyfstat/helper_functions.py
```diff
@@ -327,19 +327,43 @@ def convert_array_to_gsl_matrix(array):
...
+def get_sft_array(sftfilepattern, data_duration=None, F0=None, dF0=None):
+    """ Return the raw data (absolute value) from a set of SFTs
+
+    FIXME: currently only returns data for first detector
+    """
+
+    if F0 is None and dF0 is None:
+        fMin = -1
+        fMax = -1
+    elif F0 is None or dF0 is None:
```
Rodrigo-Tenorio, line 339: @dbkeitel  Python's `or` is the inclusive one, I think you were thinking about he exclusive one?

| F0 | dF0 | raise ValueError |
|----|-----|-----------|
| 0  | 0   | 0         | None of them: This is fine
| 1  | 0   | 1         |
| 0  | 1   | 1         |
| 1  | 1   | 0         | Both of them: This is fine!

Proposal using the Bitwise `XOR` operator
```
elif (F0 is None) ^ (dF0 is None):
```

pyfstat/helper_functions.py
```diff
@@ -327,19 +327,43 @@ def convert_array_to_gsl_matrix(array):
...
+    times = [sft.epoch.gpsSeconds for sft in SFTs.data]
+    data = [np.abs(sft.data.data) for sft in SFTs.data]
     data = np.array(data).T
-    n, nsfts = data.shape
-    freqs = np.linspace(sft.f0, sft.f0 + n * sft.deltaF, n)
-    times = np.linspace(0, data_duration, nsfts)
+    nbins, nsfts = data.shape
+
+    sft0 = SFTs.data[0]
+    freqs = np.linspace(sft0.f0, sft0.f0 + nbins * sft0.deltaF, nbins)
```
Rodrigo-Tenorio, line 366: @dbkeitel Maybe you already did the math, but shouldn't this be
```
freqs = np.linspace(sft0.f0, sft0.f0 + (nbins-1) * sft0.deltaF, nbins)
```
since `np.linspace` *includes* the very last point? Alternatively 
```
freqs = np.linspace(sft0.f0, sft0.f0 + nbins * sft0.deltaF, nbins, endpoint=False)
```
Examples:
```
>>> import numpy as np
>>> np.linspace(0, 10, 10)
array([ 0.        ,  1.11111111,  2.22222222,  3.33333333,  4.44444444,
        5.55555556,  6.66666667,  7.77777778,  8.88888889, 10.        ])
>>> np.linspace(0, 9, 10)
array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
>>> np.linspace(0, 10, 10, endpoint=False)
array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
```

pyfstat/make_sfts.py
```diff
@@ -238,9 +238,12 @@ def basic_setup(self):
...
         return self.tstart + self.duration
 
-    def make_data(self):
+    def make_data(self, verbose=False):
         """ A convienience wrapper to generate a cff file then sfts """
-        self.make_cff()
+        if self.h0 == 0:
+            logging.info("Got h0=0, not writing an injection .cff file.")
+        else:
+            self.make_cff(verbose)
```
Rodrigo-Tenorio, line 275: This `verbose` option doesn't apply if `self.h0==0`.

I'd say that's not a problem at all.

pyfstat/make_sfts.py
```diff
@@ -214,6 +225,22 @@ def basic_setup(self):
         self.config_file_name = os.path.join(self.outdir, self.label + ".cff")
         self.theta = np.array([self.phi, self.F0, self.F1, self.F2])
 
+        required_signal_params = [
+            # leaving out "F1","F2","psi","phi","tref" as they have defaults
```
Rodrigo-Tenorio, line 229: I understand why `F1, F2` have zero as default (They are not required for every kind of CW signal);`tref` comes from `tstart` unless explicitly given; `h0` and `cosi` set the "overall amplitude" of the CW (if `h0` has to be set, makes sense to set `cosi` as well).

Any criterion on why `psi, phi` are also set to zero other that "if you don't set them then you (most likely) don't care about them since Tobs will be long enough to swap out any bias"?

pyfstat/core.py
```diff
@@ -1588,16 +1588,15 @@ def _init_semicoherent_window_range(self):
...
                 "Initialising semicoherent parameters from {} to {} in" " {} segments"
-            ).format(self.minStartTime, self.maxStartTime, self.nsegs)
+            ).format(self.minStartTime, self.endTime, self.nsegs)
         )
         self.transientWindowType = "rect"
         self.whatToCompute = lalpulsar.FSTATQ_2F + lalpulsar.FSTATQ_ATOMS_PER_DET
-        self.tboundaries = np.linspace(
-            self.minStartTime, self.maxStartTime, self.nsegs + 1
-        )
+        self.tboundaries = np.linspace(self.minStartTime, self.endTime, self.nsegs + 1)
```
Rodrigo-Tenorio, line 1599: Doesn't this contradict Line 1609
```
if self.tboundaries[-1] > self.SFT_timestamps[-1]:
```
since `self.endTime = self.maxStartTime + self.Tsft` and `self.SFT_timestamps[-1] === self.maxStartTime`?

pyfstat/core.py
```diff
@@ -681,7 +681,7 @@ def _get_SFTCatalog(self):
         output = helper_functions.run_commandline(cl_tconv2, log_level=logging.DEBUG)
         tconvert2 = output.rstrip("\n")
         logging.info(
-            "Data spans from {} ({}) to {} ({})".format(
+            "Data contains SFT timestamps from {} ({}) to {} ({})".format(
```
Rodrigo-Tenorio, line 684: Should we add an `included` somewhere?
```
"Data contains SFT timestamps from {} ({}) to {} ({}) (closed interval)"
```
```
"Data contains SFT timestamps from {} ({}) to {} ({}) (last time stamp included)"
```

pyfstat/helper_functions.py
dbkeitel: Since it's an `elif` to the previous case of both being `None`, I think it's valid...?

00 -> L336 catches
11 or 01 or 10 -> L339 would catch, but in 11 case it never gets that far
11 -> else L341

pyfstat/helper_functions.py
dbkeitel: Good catch! This was already present in the original code, so I did in fact not check the math. Will fix.

pyfstat/make_sfts.py
dbkeitel: Well yes, in that case the make_cff is now no longer called at all, verbose or not.

pyfstat/make_sfts.py
dbkeitel: Scientific answer: because, unlike cosi, fixing psi and phi to 0 won't bias injection campaigns.

Pragmatic answer, I'd say this is the stronger one: because MFDv5 itself will throw an error if h0=0 but no explicit cosi given, while it also silently defaults to psi=phi0=0.

pyfstat/core.py
dbkeitel: Yes, that looks inconsistent. I hadn't checked that "SFT_timestamps" case, let's see if we can add that to the test suite... (insanity ensues)

dbkeitel: No, wait, that's not a separate user option; `self.SFT_timestamps` always gets set  unless `sftfilepattern==None`, so that block should get triggered in almost all test cases. That's confusing then that it never fails so far... Will look into it.

dbkeitel: ...oh, it's just a `logging.debug`, not an error. And yes it does get triggered by default. Not sure why it's there though, I think this is actually what we expect following lal's conventions:

- SFT timestamps are always the start times of the data stretches used for each one
- segment boundaries say "take all data within"

So in combination, we indeed expect in an ideal situation (no-gaps data, PyFstat splits it into segments itself, with neat integer division working out):
- self.tboundaries[0] == self.SFT_timestamps[0]
- self.tboundaries[-1] = self.SFT_timestamps[-1] + 1800

pyfstat/helper_functions.py
Rodrigo-Tenorio: Let me think about it...

0 means "None", 1 means "other than None".

###  00
336: 1 and 1 -> 0 
If none of them are given, take the whole band. Ok.

### 01, 10
336: 0 and 1 -> 0 Nothing happens
339: 0 or 1 -> 1 raise Value error
If only one of them is given, **raise error**

### 11
336: 0 and 0 -> 0 Nothing happens
339: 0 or 0 -> 0 Nothing happens
341: else -> Do something
If both are given, use them. Ok.

Rodrigo-Tenorio: Fair enough then.

pyfstat/make_sfts.py
Rodrigo-Tenorio: Thanks

pyfstat/core.py
Rodrigo-Tenorio: `self.tboundaries` are used in  [pyfstat/core.py/SemiCoherentSearch](https://github.com/PyFstat/PyFstat/blob/4af3c01a988f8633e2024d1cc3522d701702be59/pyfstat/core.py#L1579) to define `semicoherentWindowRange`.  `-self.Tcoh` is included because we want the last starting time.


The following situations define their own `tboundaries` (this is just for the record)
1. [pyfstat/optimal_setup_functions.py/get_Nstar_estimate](https://github.com/PyFstat/PyFstat/blob/master/pyfstat/optimal_setup_functions.py#L232) where it's used consistently (i.e. the last `tboundary` gets used only as end-segment time)

2. [pyfstat/mcmc_based_searches.py/get_p_value](https://github.com/PyFstat/PyFstat/blob/master/pyfstat/mcmc_based_searches.py#L2060) and [pyfstat/mcmc_based_searches.py/plot_cumulative_max](https://github.com/PyFstat/PyFstat/blob/master/pyfstat/mcmc_based_searches.py#L2423), both of which are related to

pyfstat/helper_functions.py
dbkeitel: done in https://github.com/PyFstat/PyFstat/pull/125/commits/99a6a90e785292be3a610aed59f8db09e2f71b04

pyfstat/core.py
dbkeitel: --> #126

pyfstat/core.py
dbkeitel: done in #126

pyfstat/core.py
dbkeitel: Not resolving this one here for now, will check tomorrow if this part of the change can be dropped from this PR without breaking the tests, as #126 would just revert it anyway.

dbkeitel: Essentially without this part of the diff, the SemiCoherentSearch test (or any call to that class without manually adjusting maxStartTime) will fail (or use too little data) because for maxStartTime to be the right choice here, when it was determined from SFTs the manual shift upwards as [here](https://github.com/PyFstat/PyFstat/pull/126/files#diff-ed8f52637a7d535159fd46ef4d8a91a2R715) is needed.

I'm replacing 2a33e7fdcb4463528e1da438f01d417fabd4d81a with a minimal one-line fix to just do just that, with locally defined Tsft, to be then further cleaned up by #125.

dbkeitel: Simplified commit: 1147c101098a37ade8fa18cf826d466f65ba8dfc

dbkeitel:
@Rodrigo-Tenorio good to merge now, I think.

PR #126, opened 2020-08-17
Semicoherent cleanup

dbkeitel:
To be applied after #125. Does not appear to need any additional test suite fixes. :)

    various fixes/cleanups of SemiCoherentSearch
    
     -consistently use maxStartTime both for SFT reading constraints
      and for segment setup, should be consistent with LAL conventions
     -clearer logging and error messages on user input checking and
      data readin
     -make Tsft an option of ComputeFstat and derived classes
      and don't re-set it multiple times
     -clean up when/where windows and atoms choices are made
     -added in-code comments to justify some choices

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -485,6 +485,7 @@ class ComputeFstat(BaseSearchClass):
         sftfilepattern=None,
         minStartTime=None,
         maxStartTime=None,
+        Tsft=1800,
         binary=False,
         BSGL=False,
         transientWindowType=None,
@@ -517,9 +518,10 @@ class ComputeFstat(BaseSearchClass):
         sftfilepattern : str
             Pattern to match SFTs using wildcards (*?) and ranges [0-9];
             mutiple patterns can be given separated by colons.
-        minStartTime, maxStartTime : float GPStime
-            Only use SFTs with timestemps starting from (including, excluding)
-            this epoch
+        minStartTime, maxStartTime : int
+            Only use SFTs with timestamps starting from this range,
+            following the XLALCWGPSinRange convention:
+            half-open intervals [minStartTime,maxStartTime]
         binary : bool
             If true, search of binary parameters.
         BSGL : bool
@@ -624,19 +626,21 @@ class ComputeFstat(BaseSearchClass):
                     )
             C1 = getattr(self, "injectSources", None) is None
             C2 = getattr(self, "injectSqrtSX", None) is None
-            if C1 and C2:
+            C3 = getattr(self, "Tsft", None) is None
+            if (C1 and C2) or C3:
                 raise ValueError(
-                    "If sftfilepattern==None, you must specify either one of"
-                    " injectSources or injectSqrtSX."
+                    "If sftfilepattern==None, you must specify Tsft and"
+                    " either one of injectSources or injectSqrtSX."
                 )
             SFTCatalog = lalpulsar.SFTCatalog()
-            Tsft = 1800
             Toverlap = 0
-            Tspan = self.maxStartTime - self.minStartTime
             self.detector_names = self.detectors.split(",")
             detNames = lal.CreateStringVector(*[d for d in self.detector_names])
+            # MakeMultiTimestamps follows the same [minStartTime,maxStartTime)
+            # convention as the SFT library, so we can pass Tspan like this
+            Tspan = self.maxStartTime - self.minStartTime
             multiTimestamps = lalpulsar.MakeMultiTimestamps(
-                self.minStartTime, Tspan, Tsft, Toverlap, detNames.length
+                self.minStartTime, Tspan, self.Tsft, Toverlap, detNames.length
             )
             SFTCatalog = lalpulsar.MultiAddToFakeSFTCatalog(
                 SFTCatalog, detNames, multiTimestamps
@@ -646,6 +650,7 @@ class ComputeFstat(BaseSearchClass):
 
         logging.info("Initialising SFTCatalog")
         constraints = lalpulsar.SFTConstraints()
+        constr_str = []
         if self.detectors:
             if "," in self.detectors:
                 logging.warning(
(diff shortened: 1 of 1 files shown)


pyfstat/core.py
```diff
@@ -624,19 +626,21 @@ def _get_SFTCatalog(self):
...
+                    " either one of injectSources or injectSqrtSX."
                 )
             SFTCatalog = lalpulsar.SFTCatalog()
-            Tsft = 1800
             Toverlap = 0
-            Tspan = self.maxStartTime - self.minStartTime
             self.detector_names = self.detectors.split(",")
             detNames = lal.CreateStringVector(*[d for d in self.detector_names])
+            # MakeMultiTimestamps follows the same [minStartTime,maxStartTime)
+            # convention as the SFT library, so we can pass Tspan like this
```
dbkeitel, line 640: Need to double-check when more awake - the docs [here](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tutils__h.html#ga65aa00943b3a912b7dc40ea6213fd3a3) seem to imply the contrary, but I think I ran a check already that seems to imply it's fine...

pyfstat/core.py
```diff
@@ -1590,29 +1616,45 @@ def _init_semicoherent_window_range(self):
...
-                "Initialising semicoherent parameters from {} to {} in" " {} segments"
+                "Initialising semicoherent parameters from"
+                " minStartTime={:d} to maxStartTime={:d} in {:d} segments..."
             ).format(self.minStartTime, self.maxStartTime, self.nsegs)
         )
-        self.transientWindowType = "rect"
-        self.whatToCompute = lalpulsar.FSTATQ_2F + lalpulsar.FSTATQ_ATOMS_PER_DET
         self.tboundaries = np.linspace(
             self.minStartTime, self.maxStartTime, self.nsegs + 1
         )
```
dbkeitel, line 1625: see @Rodrigo-Tenorio 's comments [here](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tutils__h.html#ga65aa00943b3a912b7dc40ea6213fd3a3):

> `self.tboundaries` are used in [pyfstat/core.py/SemiCoherentSearch](https://github.com/PyFstat/PyFstat/blob/4af3c01a988f8633e2024d1cc3522d701702be59/pyfstat/core.py#L1579) to define `semicoherentWindowRange`. `-self.Tcoh` is included because we want the last starting time.
> 
> If this bit of code I just quoted is correct, then no problems.
> 
> The following situations define their own `tboundaries` (this is just for the record)
>     1. [pyfstat/optimal_setup_functions.py/get_Nstar_estimate](https://github.com/PyFstat/PyFstat/blob/master/pyfstat/optimal_setup_functions.py#L232) where it's used consistently (i.e. the last `tboundary` gets used only as end-segment time)
>     2. [pyfstat/mcmc_based_searches.py/get_p_value](https://github.com/PyFstat/PyFstat/blob/master/pyfstat/mcmc_based_searches.py#L2060) and [pyfstat/mcmc_based_searches.py/plot_cumulative_max](https://github.com/PyFstat/PyFstat/blob/master/pyfstat/mcmc_based_searches.py#L2423),

pyfstat/core.py
```diff
@@ -1590,29 +1616,45 @@ def _init_semicoherent_window_range(self):
     def init_semicoherent_parameters(self):
         logging.info(
             (
-                "Initialising semicoherent parameters from {} to {} in" " {} segments"
+                "Initialising semicoherent parameters from"
+                " minStartTime={:d} to maxStartTime={:d} in {:d} segments..."
             ).format(self.minStartTime, self.maxStartTime, self.nsegs)
         )
-        self.transientWindowType = "rect"
-        self.whatToCompute = lalpulsar.FSTATQ_2F + lalpulsar.FSTATQ_ATOMS_PER_DET
```
Rodrigo-Tenorio, line 1597: I think we need `lalpulsar.FSTATQ_ATOMS_PER_DET` to compute the semicoherent Fstat... I'll drop this over here so I don't forget.

dbkeitel: See [L1588](https://github.com/PyFstat/PyFstat/pull/126/files#diff-ed8f52637a7d535159fd46ef4d8a91a2R1588): when we set `self.computeAtoms` in `SemiCoherentSearch`, then the `ComputeFstat` it inherits from will automatically select the right `whatToCompute` ([L740](https://github.com/PyFstat/PyFstat/pull/126/files#diff-ed8f52637a7d535159fd46ef4d8a91a2R740))

pyfstat/core.py
dbkeitel: I've removed the introduction of `tEnd` from #125 and instead just fix the automatic setting of `maxStartTime` (1147c101098a37ade8fa18cf826d466f65ba8dfc) which seems to have been the only core issue. Now this rebased commit/PR here looks much simpler in that aspect.

There's still some remaining ambiguity for real-world cases where data has gaps and the user may want to align their segment boundaries, especially the final one, more nicely. But the correct solution to that is not to tinker further with the automated setup, but to include an option for custom segments (like HSGCT's `--segmentList`).

The other special `tboundaries` usages pointed out by Rodrigo all seem fine, as long as the underlying `self.maxStartTime` is correctly set to be "> last actual SFT timestamp we have/want".

pyfstat/core.py
dbkeitel: Ok, so the official logic of that functon is:
> the returned list of timestamps are guaranteed to cover the interval [tStart, tStart+duration), assuming a each timestamp covers a length of 'Tsft' This implies that the actual timestamps-coverage can extend up to 'Tsft' beyond 'tStart+duration'. 

(Note that the actual argument is called `Tspan`, not `duration`.)

We pass `Tspan = self.maxStartTime - self.minStartTime` so it should "cover" the interval `[minStartTime,maxStartTime)` which would mean from the `)` notation that `maxStartTime` is **not** included, but from the word `cover` and the following sentence, it **should** be included. If my English reading comprehension isn't failing me, that documentation is self-contradictory. Or at least confusing; it'd be fine if it'd split the ambiguous term `actual timestamps-coverage` into `actual timestamps` which are _only in_ the open interval, and `data coverage` or sth like that which `covers` it. 

And what it actually does:
```python
>>> import lalpulsar
>>> ts=lalpulsar.MakeTimestamps(tStart=700000000,Tspan=1,Tsft=1800,Toverlap=0); print(ts.data)
[LIGOTimeGPS(700000000, 0)]
>>> ts=lalpulsar.MakeTimestamps(tStart=700000000,Tspan=1799,Tsft=1800,Toverlap=0); print(ts.data)
[LIGOTimeGPS(700000000, 0)]
>>> ts=lalpulsar.MakeTimestamps(tStart=700000000,Tspan=1800,Tsft=1800,Toverlap=0); print(ts.data)
[LIGOTimeGPS(700000000, 0)]
>>> ts=lalpulsar.MakeTimestamps(tStart=700000000,Tspan=1801,Tsft=1800,Toverlap=0); print(ts.data)
[LIGOTimeGPS(700000000, 0) LIGOTimeGPS(700001800, 0)]
```
which is consistent with `[tStart, tStart+Tspan)` for the "actual timestamps".

So I'd say it's consistent with our own intentions here and no further PyFstat changes are needed; I'll submit a lalsuite MR rewording the documentation for clarity.

Rodrigo-Tenorio: We agree this makes sense after discussing/clarifying with @dbkeitel

PR #127, opened 2020-08-18
new class SearchForSignalWithJumps

dbkeitel:
- move special methods for glitch / timing noise cases
  from BaseSearchClass to this
 - SemiCoherentGlitchSearch and GlitchWriter now inherit from new class
 - adapt tests

@GregoryAshton to be honest I always found these special methods a bit distracting right at the top of `BaseSearchClass`, and having them hidden away further down will also make life a bit easier with reviewers. ;) Is this ok with you? (Superficial look at the idea of the change is sufficient.) We don't really have very good test coverage for the glitch-robust functionality, but at least the basic tests that exist still pass.

@Rodrigo-Tenorio You seem to understand inheritance better than me, any problems you can see with the two parents for SemiCoherentGlitchSearch and GlitchWriter?

PS: The alternative to dual inheritance would have been to turn these methods into standalone functions in helper_functions (or a separate module) that are called from both classes, but this seemed more elegant and more future-proof in case we want to have e.g. a SemiCoherentTimingNoiseSearch, some specialised MCMC classes, or whatever at some point.

### DIFF
--- a/pyfstat/__init__.py
+++ b/pyfstat/__init__.py
@@ -1,6 +1,7 @@
 from .core import (
     BaseSearchClass,
     ComputeFstat,
+    SearchForSignalWithJumps,
     SemiCoherentSearch,
     SemiCoherentGlitchSearch,
 )
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -294,109 +294,6 @@ class BaseSearchClass(object):
         )
         logging.getLogger().addHandler(fh)
 
-    def _shift_matrix(self, n, dT):
-        """ Generate the shift matrix
-
-        Parameters
-        ----------
-        n : int
-            The dimension of the shift-matrix to generate
-        dT : float
-            The time delta of the shift matrix
-
-        Returns
-        -------
-        m : ndarray, shape (n,)
-            The shift matrix.
-
-        """
-        m = np.zeros((n, n))
-        factorial = np.math.factorial
-        for i in range(n):
-            for j in range(n):
-                if i == j:
-                    m[i, j] = 1.0
-                elif i > j:
-                    m[i, j] = 0.0
-                else:
-                    if i == 0:
-                        m[i, j] = 2 * np.pi * float(dT) ** (j - i) / factorial(j - i)
-                    else:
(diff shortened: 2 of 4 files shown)


Rodrigo-Tenorio (commented):
I'll drop the MRO over here so we are all aware of what is going on:
```
>>> pyfstat.SearchForSignalWithJumps.mro()
[<class 'pyfstat.core.SearchForSignalWithJumps'>, <class 'pyfstat.core.BaseSearchClass'>, <class 'object'>]
>>> pyfstat.SemiCoherentGlitchSearch.mro()
[<class 'pyfstat.core.SemiCoherentGlitchSearch'>, <class 'pyfstat.core.SearchForSignalWithJumps'>, <class 'pyfstat.core.ComputeFstat'>, <class 'pyfstat.core.BaseSearchClass'>, <class 'object'>]
>>> pyfstat.GlitchWriter.mro()
[<class 'pyfstat.make_sfts.GlitchWriter'>, <class 'pyfstat.core.SearchForSignalWithJumps'>, <class 'pyfstat.make_sfts.Writer'>, <class 'pyfstat.core.BaseSearchClass'>, <class 'object'>]
```
mermaid-diagram-20200819124832

@dbkeitel you are proposing to use a class (careful, [meta-classes](https://stackoverflow.com/questions/100003/what-are-metaclasses-in-python) are a whole kerfuffle on their own) to strictly *extend* other existing classes in order to load them with all those glitch-specific methods.

I like the idea of grouping all those methods on a class on their own, rather than having them spread as functions in a completely separated module.

Regarding multiple inheritance, I'd normally say "Don't"; however, this case is different, since 1) there are no overriden methods and 2) you need those new methods in two "sort of unrelated" classes (meaning this class *strictly extends* and hence you'd have to create a couple of intermediate classes in order to fulfill the same purpose without multiple inheritance).

dbkeitel:
Thanks for spending time on that diagram, that's exactly how it's intended!

I've fixed up the docstring to avoid saying "meta class".

GregoryAshton:
Hey folks, yes this looks sensible. Refactoring, especially to less glitch-specific setups makes a lot of sense.

Issue #123, opened 2020-08-13
check examples after Writer changes


dbkeitel:
Once the current set of Writer changes is over, examples are pretty much guaranteed to be broken...

Rodrigo-Tenorio:
Examples to look at:

- [ ] binary_examples
- [ ] followup_examples
- [ ] glitch_examples
- [ ] grid_examples
- [ ] mcmc_examples
- [ ] mcmc_vs_grid_simple_examples
- [ ] other_examples
- [ ] transient_examples

Rodrigo-Tenorio:
# Binary Examples
```
%% LAL: 7.0.0 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% LALFrame: 1.5.1 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% LALMetaIO: 1.6.1 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% LALSimulation: 2.2.0 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% LALBurst: 1.5.5 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% LALInspiral: 1.10.1 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% LALPulsar: 2.0.0 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% LALInference: 2.0.3 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% LALApps: 6.26.0 (UNCLEAN 00bf27317eabda0614e2e879038b940bb5ba36ae)
%% FstatMethod used: 'DemodSSE'
%% Started search: 2020-08-19 17:24:27.7703
%% Loaded SFTs: [ H1:4800 ]
%% GPS starttime         = 1000000000 (Wed Sep 14 01:46:25 2011 GMT)
%% Total time spanned    = 8640000 s (2400.00 hours)
%% Pulsar-params refTime = 1004320000
%% Spin-range at refTime: fkdot = [ 30:30, 0:0, 0:0, 0:0, 0:0, 0:0, 0:0 ]
2020-08-19 17:24:27.7731 (42281) [normal]: Search finished.
17:24 INFO    : Plotting temperature 0 chains
17:24 WARNING : Too few points to create valid contours
[...]

PR #129, opened 2020-08-19
Make examples compatible again

Rodrigo-Tenorio:
After #125 #122 and #121 examples are likely to be broken due to some backwards incompatible changes in the default arguments.

Refer to #123 for a brief summary (i.e. sanity check) of examples' outputs.


dbkeitel (approved):
Seems we have different opinions on what is a "default" cosi, which means it was a good decision to not have it hardcoded in the library itself any more. ;)

Apart from that, I don't like the commit that says
```
[WIP] other_examples/inject in noise

- Possible bug encountered
```
and only changes a blank line; to avoid a needless cycle of "request changes", I'll just squash everything on here into a single commit.

Issue #134, opened 2020-08-23
Add more diagnosis tools


Rodrigo-Tenorio:
There's a serious lack of convergence diagnosis tools (either they are not properly exposed/exemplified or they simply are not there).

List of what to add:

- [ ] Convergence check (either Gelman-Rubin or variance ratio).
A number is sometimes more enlightening than looking at each plot to see if "it looks right".
- [ ] Fstat trace plot
The histogram is fine, but we could really benefit if we showed the actual 2F values of the ensemble.

- Add option to plot injection parameters.
I'd say this is implemented somehow for the corner plots, but it would be useful to plot the injected value (if any)
directly on top of the walkers. This would  help to check whether there are any biases in the estimation.
    - [x] Add injection parameters to corner plot
    - [x] Add injection parameters to prior-posterior plot
    - [x] Add injection parameters to walker plot

- [x] Add example using these new tools.

As always, this could be done in two ways: Either we implement it ourselves (just like quite a fraction of the plotting functionalities) or we rely on an external packages ([ChainConsumer](https://github.com/Samreay/ChainConsumer), which is already used in some parts of the code, looks good at a first glance).

Rodrigo-Tenorio:
Plot-related improvements are done in #135

dbkeitel:
Note there are some commented-out methods to MCMCSearch already:

- `setup_burnin_convergence_testing`
- `_test_autocorr_convergence`
- `_test_GR_convergence`
- `_test_convergence`

PR #135, opened 2020-08-23
Add options to plot injection parameters for diagnosis

Rodrigo-Tenorio:
Work in #134 

- A *single* point can be added through the "truths" parameters (this is
due to the way corner is coded).
- If it's a list of values, plot straight.
- If it's a dictionary, get a list from it.


dbkeitel:
Sounds good! But do we really need a new example for it, which then needs maintenance too, or can you just add the kwarg to the basic PyFstat_example_fully_coherent_MCMC_search.py ?

Rodrigo-Tenorio:
@dbkeitel  Rebased onto Black 19

dbkeitel (approved):
Changes look fine, and I'll take the plots posted in #137 as proof that they actually work.

PR #136, opened 2020-08-24
cleanup: remove Bunch class, refactor some functions, remove deprecated options

dbkeitel:
- some minor help/logging improvements

- remove the various options/methods deprecated in the 1.5.x series.

- more significant cleanup of core module:
     - remove Bunch class which wasn't used except for an isolated test case
     - turn translate_keys_to_lal() into a BaseSearchClass method
     - move read_par() to helper_functions but also add a BaseSearchClass method
     - simplify read_par() by always returning a regular dict

- move predict_fstat() from core to helper_functions

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -440,8 +440,9 @@ class ComputeFstat(BaseSearchClass):
             grid resolutions in transient start-time and duration,
             both default to Tsft
         detectors : str
-            Two character reference to the data to use, specify None for no
-            contraint. If multiple-separate by comma.
+            Two-character references to the detectors for which to use data.
+            Specify None for no constraint.
+            For multiple detectors, separate by comma.
         minCoverFreq, maxCoverFreq : float
             The min and max cover frequency passed to CreateFstatInput.
             For negative values, these will be used as offsets from the min/max
@@ -551,8 +552,8 @@ class ComputeFstat(BaseSearchClass):
         if self.detectors:
             if "," in self.detectors:
                 logging.warning(
-                    "Multiple detector selection not available,"
-                    " using all available data"
+                    "Multiple-detector constraints not available,"
+                    " using all available data."
                 )
             else:
                 constraints.detector = self.detectors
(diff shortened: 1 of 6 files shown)


dbkeitel:
@GregoryAshton @Rodrigo-Tenorio what do you think?

pyfstat/core.py
```diff
@@ -371,6 +133,62 @@ def get_output_file_header(self):
...
+        params_dict = helper_functions.read_par(
+            filename=filename,
+            label=label or getattr(self, "label", None),
+            outdir=outdir or getattr(self, "outdir", None),
+            suffix=suffix,
+            raise_error=raise_error,
+        )
+        return params_dict
+
+    def translate_keys_to_lal(self, dictionary):
```
Rodrigo-Tenorio, line 148: Should we really bury these "general" functions into a class?
I can take  "Everything relevant in PyFstat is done through `BaseSearchClass`" as a satisfactory answer.

pyfstat/helper_functions.py
```diff
@@ -628,3 +628,164 @@ def get_parameters_dict_from_file_header(outfile, comments="#", eval_values=Fals
             val = eval(val)  # DANGER
         params_dict[key] = val
     return params_dict
+
+
+def read_par(
```
Rodrigo-Tenorio, line 633: But I see you decided to keep this one and not the other one...

Rodrigo-Tenorio: @dbkeitel Are you really keeping this function here just to do a test?
I thought Test classes were derived from BaseSearch so this function can be accessed as an inherited method...?

pyfstat/core.py
dbkeitel: For this one specifically I didn't really see any use case where someone would need to call it manually, as it's just there to resolve some internal variable name mismatch.

For `read_par()` on the other hand I made both an individually-accesible helper function and the class method which just wraps it, because I can easily imagine using it from an external script to read in pulsar ephemerides and then do something nonstandard with the ranges before calling a Writer or XYZSearch class.

pyfstat/helper_functions.py
dbkeitel: See above for why it's a stand-alone function.

The test could indeed be integrated better, however I think we've spent enough time on test cleanups for now. ;)

Rodrigo-Tenorio: No problem: It's a stand-alone function, it has a stand-alone test.

Rodrigo-Tenorio (approved):

dbkeitel:
I'm merging this now for the next release and to unblock the inevitable black version update. @GregoryAshton if you actually still have a good use-case for the `Bunch` class in mind, we can always resurrect it, just let us know.

dbkeitel:
Fixed a regression related to parsing of CFSv2 "loudest" files which actually still require `raise_error=False` because they contain complex numbers which `get_dictionary_from_lines` cannot yet parse. Also added a `#FIXME` for that, but the "just ignore errors" workaround should still work for now.

dbkeitel:
Tests and an [all-examples run](https://github.com/PyFstat/PyFstat/runs/1052452062?check_suite_focus=true) passed.

PR #138, opened 2020-08-26
Pin Black version to 19.10b0

Rodrigo-Tenorio:
Black (literally) just released a new version. 

We (@dbkeitel ) decided to pin this repo's version to the old one until current PRs are done.

PR #139, opened 2020-08-26
integration.yml: pin black to 19.10b0

dbkeitel:
- new version 20.8b0 which came out *just today* seems to be much stricter
 - should update only when no PRs are pending

--> https://pypi.org/project/black/#history

Screenshot_20200826_175646


dbkeitel:
duplicate of #138

PR #141, opened 2020-08-26
Fix plot_prior_posterior for a single inference

Rodrigo-Tenorio:
Hot fix some options for a single inference I did not properly test .


dbkeitel (approved):

dbkeitel (approved):
...github's auto-not-expanded diff view was really useless for the second change here. ;) But all looks good after I understood what it was about.

PR #143, opened 2020-08-28
Fix periastron time format

Rodrigo-Tenorio:
`tp` suffers from the same problem as `tref` in #120 . This makes sense, as both of them [are parsed using the same function in MFD](https://lscsoft.docs.ligo.org/lalsuite/lal/_config_file_test_8c_source.html).

I'll check the rest of parsing functions to make sure nothing else is broken.

### DIFF
--- a/examples/binary_examples/PyFstat_example_semi_coherent_binary_search_using_MCMC.py
+++ b/examples/binary_examples/PyFstat_example_semi_coherent_binary_search_using_MCMC.py
@@ -22,7 +22,7 @@ signal_parameters = {
     "F1": 0,
     "F2": 0,
     "Alpha": 0.15,
-    "Delta": 0.15,
+    "Delta": 0.45,
     "tp": mid_time,
     "argp": 0.3,
     "asini": 10.0,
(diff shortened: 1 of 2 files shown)


Rodrigo-Tenorio:
[The following parsing functions used by MFD](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_c_w_make_fake_data_8c_source.html#l00808) are relevant to us:
- XLALReadConfigEPOCHVariable (`tref`, `tp`)
- XLALReadConfigRAJVariable (`RA`)
- XLALReadConfigDECJVariable (`DEC`)
- XLALReadConfigREAL8Variable (the rest)

`XLALReadConfigEPOCHVariable` is already proven to work, so I'll check the ones related to sky positions in order to make sure they work as intended with their formats (they should, as they are natively implemented in PyFstat, but it well may be that someone overlook them as most examples are directed searches).

I include RA and DEC in the following inference (it's basically the binary example + sky priors). The fact that PyFstat  recovers the injected signal with no significant biases proves that `RA` and `DEC` are properly parsed.

As a side note, [the documentation for this kind of functions](https://lscsoft.docs.ligo.org/lalsuite/lal/_config_file_8c.html#a640d14ddaa2378256ae44347226107bf) seems to imply that `RA` and `DEC` are actually parsed in the same way as a `REAL8`, so there shouldn't be any problems related to them.
PyFstat_example_semi_coherent_binary_search_using_MCMC_prior_posterior
PyFstat_example_semi_coherent_binary_search_using_MCMC_walkers

dbkeitel (approved):
Yes, that makes a lot of sense! The function could be cleaned up further to make this sort of thing easier to catch and maintain -> #144 but merging for now.

Issue #144, opened 2020-08-29
build a dict directly in parse_args_consistent_with_mfd()


dbkeitel:
Following #120 and #143 it might be a good idea to change `parse_args_consistent_with_mfd()` to build a dict directly, instead of lists that are then zipped into a dict. Makes the code more readable and easier to maintain.

Rodrigo-Tenorio:
Closed by #145

Issue #137, opened 2020-08-26
Better understand time of periastron passage


Rodrigo-Tenorio:
After doing some more informative plots using #135 , it looks like there's something odd related to the inference of time of periastron passage.

![PyFstat_example_semi_coherent_binary_search_using_MCMC_walkers](https://user-images.githubusercontent.com/49152330/91310455-2dfd6e80-e7b2-11ea-8e56-79cd8785d31d.png)

Rodrigo-Tenorio:
Things don't look to improve if we add some leeway into the periastron argument...
![PyFstat_example_semi_coherent_binary_search_using_MCMC_walkers](https://user-images.githubusercontent.com/49152330/91312010-04454700-e7b4-11ea-9714-c91348348a78.png)

Rodrigo-Tenorio:
High eccentricity doesn't look to change the game:
![PyFstat_example_semi_coherent_binary_search_using_MCMC_walkers](https://user-images.githubusercontent.com/49152330/91312985-24c1d100-e7b5-11ea-93f7-c679d133e678.png)

dbkeitel:
I know almost nothing about how tp is implemented, so guessing completely in the dark: is there any chance this may have something to do with reference time offset between injection and recovery?

Rodrigo-Tenorio:
> I know almost nothing about how tp is implemented, so guessing completely in the dark: is there any chance this may have something to do with reference time offset between injection and recovery?

My guesses right now are either that or a different convetion: tp is essentially an initial phase (it's actually `tp * Omega`), so you can either choose [0, 2pi) or [-pi, pi). BinarySkyHough chooses the latter, and I'd say [LALSuite ](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_s_btimes__h.html#gac88f72b4d1e7281a9bb1f0bbf131bf3a) somehow chooses the former...

tl;dr It may well just be a matter of conventions

Rodrigo-Tenorio:
Just checked a different `tref`: No changes.

In fact, if I'm not mistaken, `tref` is consistently used as the reference point to Taylor-expand the phase evolution of a continuous wave in terms of the `source emission time` i.e. it's defined before taking into account any proper motion (the only affected parameters are the initial phase, the frequency and spin-down parameters).

dbkeitel:
That should be it, yes! I just thought about it because I know we set a `tref=tstart` default somewhere, while most lal codes actually default to `tref=tmid`.

[...]

PR #145, opened 2020-08-30
Make simpler parameter parsing method

Rodrigo-Tenorio:
Implement #144 and probably further simplify writer classes

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -291,7 +291,7 @@ phi0 = {:1.18e}
 Freq = {:1.18e}
 f1dot = {:1.18e}
 f2dot = {:1.18e}
-refTime = {:10.6f}"""
+refTime = {:1.9f}"""
 
     def get_single_config_line_cw(
         self, i, Alpha, Delta, h0, cosi, psi, phi, F0, F1, F2, tref
@@ -778,8 +778,8 @@ class BinaryModulatedWriter(Writer):
         gps_time_and_string_formats = {
             # GPS times should NOT be parsed using scientific notation
             # LAL routines silently parse them wrongly
-            "orbitTp": ":10.6f",
-            "refTime": ":10.6f",
+            "orbitTp": ":1.9f",
+            "refTime": ":1.9f",
             "transientWindowType": ":s",
         }
 
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -775,29 +775,25 @@ class BinaryModulatedWriter(Writer):
             "period",
             "transientWindowType",
         ]
-
-        signal_parameters = {
-            key: self.__dict__.get(key, None) for key in signal_parameter_labels
+        gps_time_and_string_formats = {
+            # GPS times should NOT be parsed using scientific notation
+            # LAL routines silently parse them wrongly
+            "orbitTp": ":10.6f",
+            "refTime": ":10.6f",
+            "transientWindowType": ":s",
         }
-        self.signal_parameters = {
-            key: value for key, value in signal_parameters.items() if value is not None
-        }
 
-        signal_parameter_formats = (len(signal_parameter_labels) - 1) * [":1.18e"] + [
-            ":s"
-        ]
-        # Pledge MFD's way of parsing time-related variables in a .cff
-        for time_label in "tref", "tp":
-            signal_parameter_formats[
-                signal_parameter_labels.index(time_label)
-            ] = ":10.6f"
-        signal_formats = dict(zip(signal_parameter_labels, signal_parameter_formats))
-
-        self.signal_parameters = translate_keys_to_lal(self.signal_parameters)
-        signal_formats = translate_keys_to_lal(signal_formats)
+        self.signal_parameters = translate_keys_to_lal(
+            {
+                key: self.__dict__[key]
+                for key in signal_parameter_labels
+                if self.__dict__.get(key, None) is not None
+            }
+        )
 
         self.signal_formats = {
-            key: signal_formats[key] for key in self.signal_parameters.keys()
+            key: gps_time_and_string_formats.get(key, ":1.18e")
+            for key in self.signal_parameters
         }
 
         if self.signal_parameters["transientWindowType"] != "none":
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -291,7 +291,7 @@ phi0 = {:1.18e}
 Freq = {:1.18e}
 f1dot = {:1.18e}
 f2dot = {:1.18e}
-refTime = {:1.9f}"""
+refTime = {:10.9f}"""
 
     def get_single_config_line_cw(
         self, i, Alpha, Delta, h0, cosi, psi, phi, F0, F1, F2, tref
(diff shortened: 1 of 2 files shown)


pyfstat/make_sfts.py
```diff
@@ -775,29 +775,25 @@ def parse_args_consistent_with_mfd(self):
...
         }
 
-        signal_parameter_formats = (len(signal_parameter_labels) - 1) * [":1.18e"] + [
-            ":s"
-        ]
-        # Pledge MFD's way of parsing time-related variables in a .cff
-        for time_label in "tref", "tp":
-            signal_parameter_formats[
-                signal_parameter_labels.index(time_label)
-            ] = ":10.6f"
```
dbkeitel, line 793: Do you remember why `.6f` is used here, while lal internally stores nanoseconds (which would be `.9f`)?

dbkeitel (commented):
Should we take the opportunity to deliver on the `This will allow us to get rid of the get_single_config_* family` suggestion? Anything missing to use your simpler function in the standard Writer class instead of those?

pyfstat/make_sfts.py
Rodrigo-Tenorio: > Should we take the opportunity to deliver on the `This will allow us to get rid of the get_single_config_* family` suggestion? Anything missing to use your simpler function in the standard Writer class instead of those?

I'll work on that in #146 

> Do you remember why `.6f` is used here, while lal internally stores nanoseconds (which would be `.9f`)?
IIRC it was written like that since the beginning... anyways, I just changed it in 93a7517 and it doesn't seem to produce any harm,

dbkeitel: Ok great!

dbkeitel:
Sorry I know you're busy with other things too, but how about adding a test to explicitly check the tp parsing works as intended?

1. it could be done with an F-stat search, explicitly checking that the max is actually recovered at the intended parameter value
1. or more simply, extend `TestBinaryModulatedWriter` to just grab the MFD commandline from the SFTs file (using `lalpulsar.SFTdataFind` and `helper_functions.get_lalapps_commandline_from_SFTDescriptor` as in [here](https://github.com/PyFstat/PyFstat/blob/f68f106bb80ebaa5ab0ee34026f587e7a9fd6452/pyfstat/make_sfts.py#L530)) only check that the parsing hasn't been mangled

Rodrigo-Tenorio:
>     2. or more simply, extend `TestBinaryModulatedWriter` to just grab the MFD commandline from the SFTs file (using `lalpulsar.SFTdataFind` and `helper_functions.get_lalapps_commandline_from_SFTDescriptor` as in [here](https://github.com/PyFstat/PyFstat/blob/f68f106bb80ebaa5ab0ee34026f587e7a9fd6452/pyfstat/make_sfts.py#L530)) only check that the parsing hasn't been mangled

I'd say that will not work, since MFD reads the parameters from a `.cff` (i.e. there are no explicit parameters in the command line).

tests.py
```diff
@@ -358,6 +366,52 @@ class TestBinaryModulatedWriter(TestWriter):
...
+            maxStartTime=this_writer.tend(),
+            **mcmc_kwargs,
+        )
+        mcmc.run(plot_walkers=False)
+        max_twoF_sample, _ = mcmc.get_max_twoF()
+
+        relative_difference = np.abs(
+            1.0 - max_twoF_sample["tp"] / default_binary_params["tp"]
+        )
+        self.assertTrue(relative_difference < 1e-5)
```
Rodrigo-Tenorio, line 413: As per the results shown in #137 and #146 , this tolerance should be enough.

pyfstat/make_sfts.py
```diff
@@ -291,7 +291,7 @@ def get_base_template(self, i, Alpha, Delta, h0, cosi, psi, phi, F0, F1, F2, tre
 Freq = {:1.18e}
 f1dot = {:1.18e}
 f2dot = {:1.18e}
-refTime = {:10.6f}"""
+refTime = {:1.9f}"""
```
dbkeitel, line 294: 10.9 ???

pyfstat/make_sfts.py
```diff
@@ -775,29 +775,25 @@ def parse_args_consistent_with_mfd(self):
...
-        signal_parameters = {
-            key: self.__dict__.get(key, None) for key in signal_parameter_labels
-        }
-        self.signal_parameters = {
-            key: value for key, value in signal_parameters.items() if value is not None
+        gps_time_and_string_formats = {
+            # GPS times should NOT be parsed using scientific notation
+            # LAL routines silently parse them wrongly
+            "orbitTp": ":1.9f",
+            "refTime": ":1.9f",
```
dbkeitel, line 782: 10.9 ???

tests.py
```diff
@@ -358,6 +366,52 @@ class TestBinaryModulatedWriter(TestWriter):
...
+        }
+        theta_prior.update({key: value for key, value in default_binary_params.items()})
+        theta_prior["tp"] = {
+            "type": "unif",
+            "lower": default_binary_params["tp"]
+            - 0.5 * default_binary_params["period"],
+            "upper": default_binary_params["tp"]
+            + 0.5 * default_binary_params["period"],
+        }
+        theta_prior.pop
```
dbkeitel, line 388: Why is the `pop` needed? (not a request for change, just want to make sure I'm not missing something similar in other use cases)

tests.py
```diff
@@ -358,6 +366,52 @@ class TestBinaryModulatedWriter(TestWriter):
...
+
+        mcmc_kwargs = {
+            "nsteps": [500],
```
dbkeitel, line 391: Have you checked if a shorter run may be sufficient? MCMC tests tend to be quite slow.

dbkeitel (dismissed):

pyfstat/make_sfts.py
Rodrigo-Tenorio:
```
>>> a = 123456789.123456789
>>> print("{:10.9f}".format(a))
123456789.123456791
>>> print("{:1.9f}".format(a))
123456789.123456791
>>> a = 1.123456789
>>> print("{:10.9f}".format(a))
1.123456789
>>> print("{:1.9f}".format(a))
1.123456789
```
This is odd... I though  that number represented the number of ensured places...?

tests.py
Rodrigo-Tenorio: Because `default_signal_params` includes `h0` and `cosi`, and I'm using a comprehension to fill up the `theta_prior`. I can manually fill up `theta_prior` if you prefer it.

tests.py
Rodrigo-Tenorio: This one runs fast on my computer, but I'll check again.

pyfstat/make_sfts.py
dbkeitel: > I though that number represented the number of ensured places...?

For `f`, it's actually the number of formatted characters summing up pre- and post-decimals, +1 for the `.` itself:
```
print("{:20.9f}".format(a))
 123456789.123456791
```
(notice leading blank space).

Since I don't think the `.cff` format cares for blanks, `{:1.9f}` should do just as well, or even just `{:.9f}`.

tests.py
dbkeitel: Ok if it works!

dbkeitel (approved):
looks good now

PR #146, opened 2020-08-31
Refactor Writer class

Rodrigo-Tenorio:
Implement the argument-independent first introduced in #39 as part of BinaryModulatedWriter into Writer.

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -44,7 +44,7 @@ class Writer(BaseSearchClass):
     gps_time_and_string_formats_as_LAL = {
         # GPS times should NOT be parsed using scientific notation
         # LAL routines silently parse them wrongly
-        "refTime": ":1.9f",
+        "refTime": ":10.9f",
         "transientWindowType": ":s",
     }
 
@@ -657,16 +657,15 @@ class BinaryModulatedWriter(Writer):
             [transientWindowType,transientStartTime,transientTau] instead!
         see `lalapps_Makefakedata_v5 --help` for help with the other paramaters
         """
-        signal_parameter_labels = super().signal_parameter_labels + [
+        self.signal_parameter_labels = super().signal_parameter_labels + [
             "tp",
             "argp",
             "asini",
             "ecc",
             "period",
         ]
-        gps_time_and_string_formats_as_LAL = super().gps_time_and_string_formats_as_LAL.update(
-            {"orbitTp": ":1.9f"}
-        )
+        self.gps_time_and_string_formats_as_LAL["orbitTp"] = ":10.9f"
+
         super().__init__(
             label=label,
             tstart=tstart,
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -28,6 +28,26 @@ class KeyboardInterruptError(Exception):
 class Writer(BaseSearchClass):
     """ Instance object for generating SFTs """
 
+    signal_parameter_labels = [
+        "tref",
+        "F0",
+        "F1",
+        "F2",
+        "Alpha",
+        "Delta",
+        "h0",
+        "cosi",
+        "psi",
+        "phi",
+        "transientWindowType",
+    ]
+    gps_time_and_string_formats_as_LAL = {
+        # GPS times should NOT be parsed using scientific notation
+        # LAL routines silently parse them wrongly
+        "refTime": ":1.9f",
+        "transientWindowType": ":s",
+    }
+
     @helper_functions.initializer
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


pyfstat/make_sfts.py
```diff
@@ -272,157 +293,29 @@ def basic_setup(self):
...
-                window,
-                transientStartTime,
-                transientTau,
-            )
-
-    def make_cff(self, verbose=False):
-        """
-        Generates a .cff file
-
+    def parse_args_consistent_with_mfd(self):
```
Rodrigo-Tenorio, line 296: Here's where all the magic happens: This method looks into the class itself using the `__dict__` attribute. If any of the attributes specified in `signal_parameter_labels` are not `None`, they are included into any output of
`get_single_config_line`, which is the main way for MFD to get signal parameters.

This way, `BinaryModulatedWriter` only needs to include its own parameters and nothing more.

Rodrigo-Tenorio:
This change works fine for `Writer` and `BinaryModulatedWriter`.

The problem now is `GlitchWriter` and the way in which `get_single_config_line` is used in there.

dbkeitel:
> The problem now is `GlitchWriter` and the way in which `get_single_config_line` is used in there.

If you don't want to spend time dealing with that one right now, it's fine if you just move the get_single function(s) to be methods of that child class. That way we have the main `Writer` cleaned up and can revisit the glitchy one later.

Rodrigo-Tenorio:
> > The problem now is `GlitchWriter` and the way in which `get_single_config_line` is used in there.
> 
> If you don't want to spend time dealing with that one right now, it's fine if you just move the get_single function(s) to be methods of that child class. That way we have the main `Writer` cleaned up and can revisit the glitchy one later.

I've just checked what would have to be done in order to make this work: 
So far I've been assuming each `Writer` injects one signal at a time; as obvious as it sounds, `GlitchWriter` injects more than one signal (i.e. a "glitch" is two signals with a discontinuity), so it's not going to be as trivial as for the binary class.

I'll take your advice.

Rodrigo-Tenorio:
Looks like I messed up something in here. I'll look into this after implementing a test in #145 
PyFstat_example_semi_coherent_binary_search_using_MCMC_walkers

Rodrigo-Tenorio:
> Looks like I messed up something in here. I'll look into this after implementing a test in #145

Solved in cf57a67, it was just a matter of class attributes not being properly set.
PyFstat_example_semi_coherent_binary_search_using_MCMC_walkers

pyfstat/make_sfts.py
```diff
@@ -722,6 +657,15 @@ def __init__(
             [transientWindowType,transientStartTime,transientTau] instead!
         see `lalapps_Makefakedata_v5 --help` for help with the other paramaters
         """
+        self.signal_parameter_labels = super().signal_parameter_labels + [
```
dbkeitel, line 660: very elegant!

pyfstat/make_sfts.py
```diff
@@ -484,6 +377,40 @@ def calculate_fmin_Band(self):
...
+            logging.info(
+                "Writing the following injection parameters"
+                " to config file {:s}:".format(self.config_file_name)
+            )
+            logging.info(content)
+
+        if self.check_if_cff_file_needs_rewritting(content):
+            config_file = open(self.config_file_name, "w+")
+            config_file.write(content)
+            config_file.close()
```
dbkeitel, line 412: seems we lost some verbosity option here, but it didn't seem very informative, so I guess that's ok

pyfstat/make_sfts.py
```diff
@@ -938,6 +797,112 @@ def __init__(
             np.array([delta_phi, delta_F0, delta_F1, delta_F2]).T
         )
 
+    def get_base_template(self, i, Alpha, Delta, h0, cosi, psi, phi, F0, F1, F2, tref):
```
dbkeitel, line 800:
```suggestion
    def get_base_template(self, i, Alpha, Delta, h0, cosi, psi, phi, F0, F1, F2, tref):
        """ FIXME: ported over from Writer, should be replaced by a more elegant re-use of parse_args_consistent_with_mfd """
```
(black may require linebreaks or something)

pyfstat/make_sfts.py
```diff
@@ -272,157 +293,29 @@ def basic_setup(self):
...
-                transientTau,
-            )
-
-    def make_cff(self, verbose=False):
-        """
-        Generates a .cff file
-
+    def parse_args_consistent_with_mfd(self):
+        """ This will allow us to get rid of the get_single_config_* family
+            Future rework may use a dictionary as the default input method,
```
dbkeitel, line 298: I think these two comments are now resolved. :+1:

pyfstat/make_sfts.py
dbkeitel: Makes a lot of sense!

pyfstat/make_sfts.py
Rodrigo-Tenorio: Ups, this was not on purpose, I'll add it: It's useful to have the actual parameters after parsing on display (*ejm*)

Rodrigo-Tenorio: Actually... we did not loose any verbosity, did we?

dbkeitel (approved):

PR #148, opened 2020-08-31
create action to run all examples

dbkeitel:
- manual workflow to run all examples, e.g. before making a release


dbkeitel:
Apparently this has to be on master to even show up on the actions tab, so I'll just merge it in and live-test it.

PR #152, opened 2020-09-02
predict_fstat updates and require lalsuite >= 6.76

### LINKED ISSUE #132: lalapps_PredictFStat update
A lalapps release containing a PFS fix from https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1422 will be out imminently. ~~At that point, our combination of --DataFiles and --maxStartTime will stop working,~~ [see correction blow] so we'll need a small change and a lalsuite minimum version requirement bump [only for new, without-DataFiles operation mode].

~~In fact, I should have written a **maximum** version into the v1.6.0 requirements. :(~~

dbkeitel:
- since lalapps 6.26.1, PFS option combinations have changed a bit
  including new --duration when *not* using --DataFiles
 - now wrap the various combinations, making most arguments optional,
  in helper_functions.predict_fstat()
 - using maxStartTime=minStartTime+duration in with-DataFiles case
  is safe here following the usual [,) convention
 - but keep the original mode in Writer.predict_fstat():
  always use sftfilepattern, since (tstart,duration) do not
  fully prescribe the actual data set if we started from
  noiseSFTs with gaps
 - also improved documentation of helper_functions.predict_fstat()
 - closes  #132

@Rodrigo-Tenorio for review, but not very urgent.

PS: Forgot the most important bit: will bump lalsuite dependency to >= 6.76.

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -715,15 +715,16 @@ def get_dictionary_from_lines(lines, comments, raise_error):
 
 
 def predict_fstat(
-    h0,
-    cosi,
-    psi,
-    Alpha,
-    Delta,
-    Freq,
-    sftfilepattern,
-    minStartTime,
-    maxStartTime,
+    h0=None,
+    cosi=None,
+    psi=None,
+    Alpha=None,
+    Delta=None,
+    Freq=None,
+    sftfilepattern=None,
+    timestampsFiles=None,
+    minStartTime=None,
+    duration=None,
     IFOs=None,
     assumeSqrtSX=None,
     tempory_filename="fs.tmp",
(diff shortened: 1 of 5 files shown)


dbkeitel:
Note: I should take my own advice and add a dedicated test for the additional modes which aren't covered through `Writer`.

Rodrigo-Tenorio (dismissed):
Looks right to me at a first glance.

dbkeitel:
Converted to draft while wrangling more issues of PFS itself: https://git.ligo.org/lscsoft/lalsuite/-/issues/376

dbkeitel:
> Converted to draft while wrangling more issues of PFS itself: https://git.ligo.org/lscsoft/lalsuite/-/issues/376

I think I've worked around these. Extra test for non-`DataFiles` cases added and [examples all still pass](https://github.com/PyFstat/PyFstat/runs/1061329743?check_suite_focus=true).

pyfstat/helper_functions.py
```diff
@@ -738,40 +739,106 @@ def predict_fstat(
...
-    cl_pfs.append("--Freq={}".format(Freq))
-
-    cl_pfs.append("--DataFiles='{}'".format(sftfilepattern))
-    if assumeSqrtSX:
+
+    if h0 is not None:
+        cl_pfs.append("--h0={}".format(h0))
+    if cosi is not None:
+        cl_pfs.append("--cosi={}".format(cosi))
+    # NOTE: as of lalsuite 6.76, [psi,Alpha,Delta] are required even for
```
Rodrigo-Tenorio, line 791: Fair enough

Rodrigo-Tenorio (approved):

PR #153, opened 2020-09-02
adopt flake8 linting

dbkeitel:
We've had `flake8` config settings in `setup.cfg` for a while since `black` also uses them, but we have not actually run this linter. Locally I get at the moment:
```
1     E266 too many leading '#' for block comment
1     E722 do not use bare 'except'
1     E731 do not assign a lambda expression, use a def
3     E741 ambiguous variable name 'l'
5     F401 'matplotlib.pyplot as plt' imported but unused
3     F811 redefinition of unused 'outdir' from line 5
28    F821 undefined name 'sampler'
5     F841 local variable 'e' is assigned to but never used
66    W605 invalid escape sequence '\m'
```

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -29,6 +29,10 @@ jobs:
         wget $EPHEM_SRC/sun00-40-DE405.dat.gz -P $LALPULSAR_DATADIR
         echo "Successfully downloaded ephemerides:"
         ls -l $LALPULSAR_DATADIR
+    - name: Linting with flake8
+      run: |
+        pip install flake8
+        flake8 --count --statistics .
     - name: Style check with black
       run: |
         pip install black
--- a/README.md
+++ b/README.md
@@ -151,13 +151,17 @@ For a general introduction to installing modules, see
 [here](https://docs.python.org/3/installing/index.html).
 
 *Optional dependencies*:
-* [pycuda](https://pypi.org/project/pycuda/), required for the `tCWFstatMapVersion=pycuda`
+* [pycuda](https://pypi.org/project/pycuda/),
+  required for the `tCWFstatMapVersion=pycuda`
   option of the `TransientGridSearch` class.
   (Note: `pip install pycuda` requires a working `nvcc` compiler in your path.)
-* [pytest](https://docs.pytest.org) for running the test suite locally (`python -m pytest tests.py`)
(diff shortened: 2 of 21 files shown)


dbkeitel:
[CI output](https://github.com/PyFstat/PyFstat/pull/153/checks?check_run_id=1061853954) for some reason finds 1 issue less (112 instead of 113) but I'll try to reduce that number by 2 orders of magnitude first, before worrying about +-1. ;)

dbkeitel:
I've fixed all the easy ones. Remaining:

1. escape-character warnings from `versioneer.py` - I've just added that one to the ignore list, since fixing such small fry in it will only cause conflicts if there is ever an update to versioneer
1. actual problem to fix in `MCMCFollowUpSearch.run()`:
```
./pyfstat/mcmc_based_searches.py:3109:39: F821 undefined name 'sampler'
./pyfstat/mcmc_based_searches.py:3113:22: F821 undefined name 'sampler'
```
It wasn't immediately clear to me from a first reading of the code what's going on here, it looks simply invalid, but probably it's in reality always defined in the early iterations of the loop before ever hitting that branch. May just need a `None` ininitaliser, but let's have a careful look.
1. **lots** of `undefined name` in `tcw_fstat_map_funcs.py` due to my `_optional_import()` hack which flake8 really doesn't like. I've temporarily added this to ignore as well and will make a reminder issue for a deeper fix.

dbkeitel:
@Rodrigo-Tenorio I've marked this "ready for review" and will put the fix for the remaining `MCMCFollowUpSearch` into a separate PR because I think a slight global cleanup will be a neater solution to that.

dbkeitel:
Rebased on top of #155 and updated README.

dbkeitel:
For the `W605 invalid escape sequence` errors on LaTeX labels, the recommend solution I found was to use `r"strings"`, but as it turns out from running the glicht examples, this actually breaks as soon as `\n` newlines are involved. Need to switch to the more traditional `\\` double-escape solution.

Rodrigo-Tenorio (approved):

dbkeitel:
> For the `W605 invalid escape sequence` errors on LaTeX labels, the recommend solution I found was to use `r"strings"`, but as it turns out from running the glicht examples, this actually breaks as soon as `\n` newlines are involved. Need to switch to the more traditional `\\` double-escape solution.

Changed to doing this in 84ee9307727f73e38facf56fb2c388b9fba15f2b. All examples now pass.

PR #155, opened 2020-09-03
MCMC: store sampler as self.sampler

dbkeitel:
- allows to remove redundant arguments from various methods
 - also introduce a new MCMCFollowUpSearch._get_p0_per_stage() helper method making use of it,   which along the way will avoid F821 when we enable flake8 (#153)

@Rodrigo-Tenorio we'll need to merge this one first, does it look safe?


Rodrigo-Tenorio (approved):
Just as discussed, promote `sampler` to attribute and put the logic inside a method.

PR #158, opened 2020-09-08
Add artifact upload

Rodrigo-Tenorio:
Extend #147 by saving the output as artifacts (#149 )

PR #159, opened 2020-09-08
close figures after saving (if not returning)

dbkeitel:
The current version can lead to Frankenstein figures when doing custom plots in a wrapper script after calling some of the pre-implemented plotting methods.

 - do this in various grid and MCMC search methods
 - also add save_fig option to MCMCSearch.plot_prior_posterior()

There is still some horrible inconsistency in argument names and returns of these plotting methods, I'll open a separate reminder issue about those.

### DIFF
--- a/pyfstat/grid_based_searches.py
+++ b/pyfstat/grid_based_searches.py
@@ -978,6 +978,7 @@ class SliceGridSearch(GridSearch):
 
         if save:
             fig.savefig(os.path.join(self.outdir, self.label + "_slice_projection.png"))
+            plt.close(fig)
         else:
             return fig, axes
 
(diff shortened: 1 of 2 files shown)


dbkeitel:
Will wait for #158 on this one and use the opportunity to check the examples output artifacts.

dbkeitel:
Rebased after #158 and launched an example action to double-check plots: https://github.com/PyFstat/PyFstat/actions/runs/246925595

dbkeitel:
Plots all look fine except for the unrelated issue #164. Any comments @Rodrigo-Tenorio ?

Rodrigo-Tenorio (approved):

PR #162, opened 2020-09-08
Writer: fix check_cached_data_okay_to_use() for multiple IFOs

### LINKED ISSUE #161: Writer.check_cached_data_okay_to_use() broken for multi-IFOs
It seems that `Writer.check_cached_data_okay_to_use()` is currently broken for multi-IFO `sftfilepath`, always saying there is no match.

Test: repeatedly running e.g. `examples/mcmc_vs_grid_simple_example/PyFstat_example_make_data_for_simple_mcmc_vs_grid.py` .
->
```
No SFT file matching PyFstat_example_data/PyFstat_example_simple_mcmc_vs_grid/H-96_H1_1800SFT_simulated_signal-1000000000-172800.sft;PyFstat_example_data/PyFstat_example_simple_mcmc_vs_grid/L-96_L1_1800SFT_simulated_signal-1000000000-172800.sft found. Will create new SFT file.
```

Obvious source of the problem: `os.path.isfile(self.sftfilepath)` doesn't understand `;`-separated lists of files.

dbkeitel:
- need to loop over `self.sftfilepath.split(";")`
 - also once again store MFD option `--IFOs` in quoted format matching its internal string-list handling (regression from 523454ee)
 - improve check logic and logging output
 - also fix method name `check_if_cff_file_needs_rewritting()` -> `check_if_cff_file_needs_rewriting()`
 - closes #161

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -394,45 +394,49 @@ class Writer(BaseSearchClass):
         else:
             logging.info("Writing config file {:s}...".format(self.config_file_name))
 
-        if self.check_if_cff_file_needs_rewritting(content):
+        if self.check_if_cff_file_needs_rewriting(content):
             config_file = open(self.config_file_name, "w+")
             config_file.write(content)
             config_file.close()
 
     def check_cached_data_okay_to_use(self, cl_mfd):
-        """ Check if cached data exists and, if it does, if it can be used """
+        """ Check if matching data already exists in SFT files and can be used. """
 
-        need_new = "Will create new SFT file."
+        need_new = "Will create new SFT file(s)."
 
-        logging.info("Checking if cached data good to reuse...")
-        if os.path.isfile(self.sftfilepath) is False:
-            logging.info(
-                "No SFT file matching {} found. {}".format(self.sftfilepath, need_new)
-            )
-            return False
-        else:
-            logging.info("OK: Matching SFT file found.")
+        logging.info("Checking if we can re-use existing SFT data file(s)...")
+        sftfiles = self.sftfilepath.split(";")
+        for sftfile in sftfiles:
+            if os.path.isfile(sftfile) is False:
+                logging.info(
+                    "...no SFT file matching '{}' found. {}".format(sftfile, need_new)
+                )
+                return False
+        logging.info("...OK: file(s) found matching '{}'.".format(sftfile))
 
         if "injectionSources" in cl_mfd:
             if os.path.isfile(self.config_file_name):
-                if os.path.getmtime(self.sftfilepath) < os.path.getmtime(
-                    self.config_file_name
+                if np.any(
+                    [
+                        os.path.getmtime(sftfile)
+                        < os.path.getmtime(self.config_file_name)
+                        for sftfile in sftfiles
+                    ]
                 ):
                     logging.info(
                         (
-                            "The config file {} has been modified since the SFT file {} "
-                            + "was created. {}"
+                            "...the config file '{}' has been modified since"
+                            " creation of the SFT file(s) '{}'. {}"
                         ).format(self.config_file_name, self.sftfilepath, need_new)
                     )
                     return False
                 else:
                     logging.info(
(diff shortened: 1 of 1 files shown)


pyfstat/make_sfts.py
```diff
@@ -441,50 +445,61 @@ def check_cached_data_okay_to_use(self, cl_mfd):
...
-            return False
-        if not helper_functions.match_commandlines(cl_old, cl_mfd):
-            logging.info("Commandlines unmatched. " + need_new)
-            return False
-        else:
-            logging.info("OK: Commandline matched with old SFT header.")
-        logging.info("Looks like cached data matches current options, will re-use it!")
+            if len(cl_old) == 0:
+                logging.info(
+                    "......ould not obtain comparison commandline from first SFT"
```
Rodrigo-Tenorio, line 459: I guess this is just a typo:
```suggestion
                    "......could not obtain comparison commandline from first SFT"
```

pyfstat/make_sfts.py
```diff
@@ -394,45 +394,49 @@ def make_cff(self, verbose=False):
...
 
-        if self.check_if_cff_file_needs_rewritting(content):
+        if self.check_if_cff_file_needs_rewriting(content):
             config_file = open(self.config_file_name, "w+")
             config_file.write(content)
             config_file.close()
 
     def check_cached_data_okay_to_use(self, cl_mfd):
-        """ Check if cached data exists and, if it does, if it can be used """
+        """ Check if matching data already exists in SFT files and can be used. """
```
Rodrigo-Tenorio, line 403: This line suggest a stronger condition than what it actually does: We don NOT check that the SFTs contain the signal present in the `.cff`, rather we

1) Check that SFT files are actually present.
2) Asses that SFT files were created after the `.cff` file.
3) Check that a file with the same name as said `.cff` file was used in the SFT creation.

This method does **not** check for the `.cff` contents, nor that should be suggested, as one has to manually trigger another method to do so. 

I could accept that `...if maching data...` actually refers to such condition, but looks unclear to me.

```suggestion
        """ Check if a signal was injected into the SFT files and these can be re-used. """
```

pyfstat/make_sfts.py
```diff
@@ -441,50 +445,61 @@ def check_cached_data_okay_to_use(self, cl_mfd):
...
-                    "File contents match, no update of {} required.".format(
+                    "...OK: file contents match, no update of {} required.".format(
                         self.config_file_name
                     )
                 )
                 return False
             else:
                 logging.info(
-                    "File contents unmatched, updating {}.".format(
+                    "...file contents unmatched, updating {}.".format(
```
Rodrigo-Tenorio, line 502: Just a quick question:

This way of comparing two files assumes two differently formatted files to be different even when they contain the very same signal parameters.

Hence, by doing things in this way we are stating "If you want to use PyFstat in all its glory, please follow the right and obvious way to do an injection and stop messing with `.cff` files yourself.".

I'm Ok with this. Any comments/objections?

pyfstat/make_sfts.py
dbkeitel: Good point, though the commandline consistency check is actually quite powerful. Also your suggestion about "a signal was injected" is also too strong by the same logic. Let's go with an extended docstring:
```suggestion
        """ Check if SFT files already exist that can be re-used.
        
             This does not check the actual data contents of the SFTs,
             but only the following criteria:
              * filename
              * if an injection is requested, that the .cff file is older than the SFTs
                (but its contents should have been checked separately)
              * that the commandline stored in the (first) SFT header matches
        """
```

pyfstat/make_sfts.py
dbkeitel: We could use set operations as in `helper_functions.match_commandlines()` over the set of lines to allow for re-ordering, and whitespace-ignoring comparisons on each line. Let's make a reminder issue for that as a future improvement.

dbkeitel: #165

Rodrigo-Tenorio (approved):

Issue #163, opened 2020-09-08
Fix logging behaviour


Rodrigo-Tenorio:
There are at least two known issues related to logging and PyFstat:

- [ ] Logging messages appear twice when run under GitHub's CI (and containers as well!)
- [ ] Logging output goes to STDERR rather than STDOUT

Not a high priority issue: Any error worth of going into STDERR will terminate the job, although it would be worth knowing why this happens.

dbkeitel:
- [ ] each Writer / search class should self-identify in the log at init time. Currently the first message from each class is often something very random like band estimation. (I'm planning to do this one quickly after the current bugfix PRs are in.)

dbkeitel:
Regarding duplicate logging: This also happens for me with pytest locally.

Regarding stderr vs stdout:

- from https://docs.python.org/3/howto/logging.html :
> By default, no destination is set for any logging messages. You can specify a destination (such as console or file) by using basicConfig() as in the tutorial examples. If you call the functions debug(), info(), warning(), error() and critical(), they will check to see if no destination is set; and if one is not set, they will set a destination of the console (sys.stderr) and a default format for the displayed message before delegating to the root logger to do the actual message output.

- https://stackoverflow.com/questions/16061641/python-logging-split-between-stdout-and-stderr
- https://stackoverflow.com/questions/2302315/how-can-info-and-debug-logging-message-be-sent-to-stdout-and-higher-level-messag

Rodrigo-Tenorio:
>     * [ ]  each Writer / search class should self-identify in the log at init time. Currently the first message from each class is often something very random like band estimation. (I'm planning to do this one quickly after the current bugfix PRs are in.)

Regarding this, I'm sure we can just add something fancy using `self.__name__` in `BaseSearchClass` which gets automatically inherited by everyone else (since I'd say everything inherits from`BaseSearchClass` in the end).

Rodrigo-Tenorio:
[According to logging's documentation](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) it's the application, rather than the library, the ultimate responsible to take care of logging.

PR #169, opened 2020-09-10
Include logging print after every instantiation

Rodrigo-Tenorio:
This solves part of #163 : Every single class will announce itself before doing any action.

Also cleans up inheritances from `object`, which are not needed anymore (welcome to Python3, PyFstat!)

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -40,6 +40,11 @@ detector_colors = {"h1": "C0", "l1": "C1"}
 class BaseSearchClass(object):
     """ The base search class providing parent methods to other searches """
 
+    def __new__(cls, *args, **kwargs):
+        logging.info(f"Creating {cls.__name__} object...")
+        instance = super().__new__(cls)
+        return instance
+
     def _add_log_file(self, header=None):
         """ Log output to a file, requires class to have outdir and label """
         header = [] if header is None else header
@@ -1043,7 +1048,7 @@ class ComputeFstat(BaseSearchClass):
                 maxStartTime=t,
                 sftfilepattern=self.sftfilepattern,
                 IFO=IFO,
-                **pfs_input
+                **pfs_input,
             )
             for t in times
         ]
@@ -1062,7 +1067,7 @@ class ComputeFstat(BaseSearchClass):
         savefig=True,
         title=None,
         plt_label=None,
-        **kwargs
+        **kwargs,
     ):
         """Plot the twoF value cumulatively
 
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -37,7 +37,7 @@ args, tqdm = helper_functions.set_up_command_line_arguments()
 detector_colors = {"h1": "C0", "l1": "C1"}
 
 
-class BaseSearchClass(object):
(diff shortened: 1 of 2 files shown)


pyfstat/core.py
```diff
@@ -40,6 +40,12 @@
 class BaseSearchClass(object):
     """ The base search class providing parent methods to other searches """
 
+    def __new__(cls, *args, **kwargs):
+        logging.info(f"Creating {cls.__name__} object...")
+        instance = super().__new__(cls)
```
dbkeitel, line 45: I'm new to `__new__`, but shouldn't this pass on `, *args, **kwargs` here...?

pyfstat/core.py
```diff
@@ -40,6 +40,12 @@
...
+        logging.info("...done!")
```
dbkeitel, line 46: I think one line is enough, would drop this

pyfstat/core.py
```diff
@@ -1043,7 +1049,7 @@ def _calculate_predict_fstat_cumulative(
                 maxStartTime=t,
                 sftfilepattern=self.sftfilepattern,
                 IFO=IFO,
-                **pfs_input
+                **pfs_input,
```
dbkeitel, line 1052: I guess this and below is another case of "new black" getting randomly picky on unchanged lines...?

dbkeitel (changes requested):

pyfstat/core.py
Rodrigo-Tenorio: My understanding is "No" since `__new__` falls back to the most fundamental class (which I believe is `object`) which takes nothing. 

The reason for those `*args, **kwargs` is because child classes will try to pass further parameters to `__new__` (the same ones as `__init__` I believe) but we don't care about them since we are only interested in adding a default print every time a new object is instantiated.

pyfstat/core.py
Rodrigo-Tenorio: Yes, it is.

pyfstat/core.py
Rodrigo-Tenorio: [Let me quote the actual docs](https://docs.python.org/3/reference/datamodel.html#object.__new__)

- `__new__` gets the same arguments as `__init__`.
- After returning a new instance, `__init__` is run on that instance using the arguments passed to `__new__`.

In other words, whatever is going into `__init__` goes first through `__new__`. Since we don't do anything fancy in `__new__`, we can ignore all of those parameters and fallback to the default `__new__`  method; but we still need to allow for such parameters to get included as they will then go to `__init__`, where we do need them.

dbkeitel: "using super().__new__(cls[, ...]) with appropriate arguments" seems to imply the (kw)args should be passed through, doesn't it?

dbkeitel: But the examples I found recommending args/kwargs passing actually all seem to be from python2 days, and don't run in python 3.7.3 themselves. So: objection retracted; if your version runs, let's go with it.

dbkeitel (approved):

dbkeitel (approved):

PR #171, opened 2020-09-15
make start, tend optional in get_fullycoherent_twoF()

### LINKED ISSUE #167: TransientGridSearch problem with minStartTime=None, maxStartTime=None
`TransientGridSearch` has a problem with `minStartTime=None`, `maxStartTime=None`, trying to initialise from t0 from `int(None)`. Even with those options, a proper start time should be available, read from SFTs, through `ComputeFstat.__init__()`, but that is called too late and the result not propagated back to the `self` itself.

My first idea was to move the `inititate_search_object()` call from `run()` to `__init__()` but that's a bad idea because it doesn't play well with reusing results from disk. Will instead need some slightly smarter if-first-run-or-not if/else logic.

dbkeitel:
This is another case where fixing something small ballooned into reworking a core feature, because working around it was just too messy.

The main change is to make `tstart` and `tend` optional arguments in the core function `get_fullycoherent_twoF()`, falling back to `self.minStartTime` and `self.maxStartTime` if they're `None`, which is sufficient for the basic CW classes `GridSearch` and `MCMCSearch` and only needs to be overridden point-by-point for some of the transient and glitch classes.

- requires lots of tracking changes in grid and mcmc searches
- change grid searches to use named-column ndarrays internally to avoid fragile numeric indexing
- partially clean up keys storage in MCMC searches
- change helper_functions.get_doppler_params_output_format() to return dict
- amend tests and examples
~- WIP: SlidingWindow and FrequencySlidingWindow broken for now~
- closes #167

### DIFF
--- a/examples/grid_examples/PyFstat_example_grid_search_F0.py
+++ b/examples/grid_examples/PyFstat_example_grid_search_F0.py
@@ -69,11 +69,10 @@ search = pyfstat.GridSearch(
 )
 search.run()
 
+print("Plotting 2F(F0)...")
 fig, ax = plt.subplots()
-xidx = search.keys.index("F0")
-frequencies = np.unique(search.data[:, xidx])
-twoF = search.data[:, -1]
-
+frequencies = search.data["F0"]
+twoF = search.data["twoF"]
 # mismatch = np.sign(x-F0)*(duration * np.pi * (x - F0))**2 / 12.0
 ax.plot(frequencies, twoF, "k", lw=1)
 DeltaF = frequencies - F0
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -861,8 +861,6 @@ class ComputeFstat(BaseSearchClass):
 
     def get_fullycoherent_twoF(
         self,
-        tstart,
-        tend,
         F0,
         F1,
         F2,
@@ -873,6 +871,8 @@ class ComputeFstat(BaseSearchClass):
         ecc=None,
         tp=None,
         argp=None,
+        tstart=None,
+        tend=None,
     ):
         """ Returns twoF or ln(BSGL) fully-coherently at a single point """
         self.PulsarDopplerParams.fkdot = np.array([F0, F1, F2, 0, 0, 0, 0])
@@ -903,6 +903,12 @@ class ComputeFstat(BaseSearchClass):
             log10_BSGL = lalpulsar.ComputeBSGL(twoF, self.twoFX, self.BSGLSetup)
             return log10_BSGL / np.log10(np.exp(1))
 
+        tstart = tstart or getattr(self, "minStartTime", None)
+        tend = tend or getattr(self, "maxStartTime", None)
+        if tstart is None or tend is None:
+            raise ValueError(
+                "Need tstart or self.minStartTime, and tend or self.maxStartTime!"
+            )
         self.windowRange.t0 = int(tstart)  # TYPE UINT4
         if self.windowRange.tauBand == 0:
             # true single-template search also in transient params:
@@ -1691,13 +1697,13 @@ class SemiCoherentGlitchSearch(SearchForSignalWithJumps, ComputeFstat):
             ts, te = tboundaries[i], tboundaries[i + 1]
             if te - ts > 1800:
                 twoFVal = self.get_fullycoherent_twoF(
-                    ts,
-                    te,
-                    theta_i_at_tref[1],
-                    theta_i_at_tref[2],
-                    theta_i_at_tref[3],
-                    Alpha,
-                    Delta,
+                    F0=theta_i_at_tref[1],
+                    F1=theta_i_at_tref[2],
+                    F2=theta_i_at_tref[3],
+                    Alpha=Alpha,
+                    Delta=Delta,
+                    tstart=ts,
+                    tend=te,
                 )
                 twoFSum += twoFVal
 
@@ -1725,20 +1731,26 @@ class SemiCoherentGlitchSearch(SearchForSignalWithJumps, ComputeFstat):
         )
 
         twoFsegA = self.get_fullycoherent_twoF(
-            self.minStartTime, tglitch, theta[0], theta[1], theta[2], Alpha, Delta
+            F0=theta[0],
+            F1=theta[1],
+            F2=theta[2],
+            Alpha=Alpha,
+            Delta=Delta,
+            tstart=self.minStartTime,
+            tend=tglitch,
         )
 
         if tglitch == self.maxStartTime:
             return twoFsegA
 
         twoFsegB = self.get_fullycoherent_twoF(
-            tglitch,
-            self.maxStartTime,
-            theta_post_glitch[0],
-            theta_post_glitch[1],
-            theta_post_glitch[2],
-            Alpha,
-            Delta,
+            F0=theta_post_glitch[0],
+            F1=theta_post_glitch[1],
+            F2=theta_post_glitch[2],
+            Alpha=Alpha,
+            Delta=Delta,
+            tstart=tglitch,
+            tend=self.maxStartTime,
         )
 
         return twoFsegA + twoFsegB
--- a/pyfstat/grid_based_searches.py
+++ b/pyfstat/grid_based_searches.py
@@ -336,14 +336,23 @@ class GridSearch(BaseSearchClass):
             self.data = data
             self.save_array_to_disk()
 
-    def get_savetxt_fmt(self):
-        fmt = helper_functions.get_doppler_params_output_format(self.output_keys)
-        fmt[self.detstat] = "%.9g"
-        return fmt
+    def get_savetxt_fmt_dict(self):
+        fmt_dict = helper_functions.get_doppler_params_output_format(self.output_keys)
+        fmt_dict[self.detstat] = "%.9g"
+        return fmt_dict
+
+    def get_savetxt_fmt_list(self):
+        """Returns a list of output format specifiers, ordered like the data.
+        This is required because the output of get_savetxt_fmt_dict()
+        will depend on the order in which those entries have been coded up.
+        """
+        fmt_dict = self.get_savetxt_fmt_dict()
+        fmt_list = [fmt_dict[key] for key in self.output_keys]
+        return fmt_list
 
     def _get_tolerance_from_savetxt_fmt(self):
         """ decide appropriate input grid comparison tolerance from fprintf formats """
-        fmt = self.get_savetxt_fmt()
+        fmt = self.get_savetxt_fmt_dict()
         rtol = {}
         atol = {}
         for key, f in fmt.items():
@@ -370,7 +379,7 @@ class GridSearch(BaseSearchClass):
         logging.info("Saving data to {}".format(self.out_file))
         header = "\n".join(self.output_file_header)
         header += "\n" + " ".join(self.output_keys)
-        outfmt = list(self.get_savetxt_fmt().values())
+        outfmt = self.get_savetxt_fmt_list()
         Ncols = len(self.data.dtype)
         if len(outfmt) != Ncols:
             raise RuntimeError(
@@ -379,7 +388,7 @@ class GridSearch(BaseSearchClass):
                 " do not match."
                 " If your search class uses different"
                 " keys than the base GridSearch class,"
-                " override the get_savetxt_fmt"
+                " override the get_savetxt_fmt_dict"
                 " method.".format(Ncols, len(outfmt))
             )
         np.savetxt(
@@ -841,12 +850,12 @@ class TransientGridSearch(GridSearch):
         f = self.tCWfilebase + fmt.format(*vals) + ".dat"
         return f
 
-    def get_savetxt_fmt(self):
-        fmt = helper_functions.get_doppler_params_output_format(self.output_keys)
-        fmt[self.detstat] = "%.9g"
-        fmt["t0"] = "%d"
-        fmt["tau"] = "%d"
-        return fmt
+    def get_savetxt_fmt_dict(self):
+        fmt_dict = helper_functions.get_doppler_params_output_format(self.output_keys)
+        fmt_dict[self.detstat] = "%.9g"
+        fmt_dict["t0"] = "%d"
+        fmt_dict["tau"] = "%d"
+        return fmt_dict
 
     def write_F_mn(self, tCWfile, F_mn, windowRange):
         with open(tCWfile, "w") as tfp:
@@ -964,13 +973,13 @@ class GridGlitchSearch(GridSearch):
         self.set_out_file()
         self.output_file_header = self.get_output_file_header()
 
-    def get_savetxt_fmt(self):
-        fmt = helper_functions.get_doppler_params_output_format(self.output_keys)
-        fmt["delta_F0"] = "%.16g"
-        fmt["delta_F1"] = "%.16g"
-        fmt["tglitch"] = "%d"
-        fmt[self.detstat] = "%.9g"
-        return fmt
+    def get_savetxt_fmt_dict(self):
+        fmt_dict = helper_functions.get_doppler_params_output_format(self.output_keys)
+        fmt_dict["delta_F0"] = "%.16g"
+        fmt_dict["delta_F1"] = "%.16g"
+        fmt_dict["tglitch"] = "%d"
+        fmt_dict[self.detstat] = "%.9g"
+        return fmt_dict
 
 
 class SlidingWindow(DefunctClass):
--- a/pyfstat/grid_based_searches.py
+++ b/pyfstat/grid_based_searches.py
@@ -101,14 +101,13 @@ class GridSearch(BaseSearchClass):
             os.mkdir(outdir)
         self.set_out_file()
         self.search_keys = ["F0", "F1", "F2", "Alpha", "Delta"]
-        self.input_keys = ["minStartTime", "maxStartTime"] + self.search_keys
-        self.keys = self.input_keys.copy()
+        self.output_keys = self.search_keys.copy()
         if self.BSGL:
             self.detstat = "BSGL"
-            self.keys += ["logBSGL"]
+            self.output_keys += ["logBSGL"]
         else:
             self.detstat = "twoF"
-            self.keys += ["twoF"]
+            self.output_keys += [self.detstat]
         for k in self.search_keys:
             setattr(self, k, np.atleast_1d(getattr(self, k + "s")))
         self.output_file_header = self.get_output_file_header()
@@ -157,11 +156,11 @@ class GridSearch(BaseSearchClass):
                 detectors=self.detectors,
                 injectSources=self.injectSources,
             )
-
-            def cut_out_tstart_tend(*vals):
-                return self.search.get_semicoherent_det_stat(*vals[2:])
-
-            self.search.get_det_stat = cut_out_tstart_tend
+            self.search.get_det_stat = self.search.get_semicoherent_det_stat
+        # make sure to overwrite the min/max starttime in case the user
+        # passed None and they were read from SFTs
+        self.minStartTime = self.search.minStartTime
+        self.maxStartTime = self.search.maxStartTime
 
     def get_array_from_tuple(self, x):
         if len(x) == 1:
@@ -184,7 +183,7 @@ class GridSearch(BaseSearchClass):
     def get_input_data_array(self):
         logging.info("Generating input data array")
         coord_arrays = []
-        for sl in self.input_keys:
+        for sl in self.search_keys:
             coord_arrays.append(
                 self.get_array_from_tuple(np.atleast_1d(getattr(self, sl)))
             )
@@ -195,7 +194,13 @@ class GridSearch(BaseSearchClass):
             input_data = []
             for vals in itertools.product(*coord_arrays):
                 input_data.append(vals)
-            self.input_data = np.array(input_data)
+            input_dtype = np.dtype(
+                {
+                    "names": self.search_keys,
+                    "formats": np.repeat(float, len(self.search_keys)),
+                }
+            )
+            self.input_data = np.array(input_data, dtype=input_dtype)
 
     def check_old_data_is_okay_to_use(self):
         if args.clean:
@@ -238,30 +243,27 @@ class GridSearch(BaseSearchClass):
             )
 
         logging.info("Loading old data from '{:s}'.".format(self.out_file))
-        old_data = np.atleast_2d(np.genfromtxt(self.out_file, delimiter=" "))
-        # need to convert any "None" entries in input_data array safely to 0s
-        # to make np.allclose() work reliably
-        new_data = np.nan_to_num(self.input_data.astype(np.float64))
-        if np.shape(old_data)[0] != np.shape(new_data)[0]:
+        old_data = helper_functions.read_txt_file_with_header(self.out_file, names=True)
+        if len(old_data) != len(self.input_data):
             logging.info(
                 "Old data found in '{:s}', but differs"
                 " in length ({:d} points in file, {:d} points requested);"
                 " continuing with grid search.".format(
-                    self.out_file, np.shape(old_data)[0], np.shape(new_data)[0]
+                    self.out_file, np.shape(old_data)[0], np.shape(self.input_data)[0]
                 )
             )
             return False
-        if np.shape(old_data)[1] < np.shape(new_data)[1]:
+        if len(old_data.dtype) < len(self.input_data.dtype):
             logging.info(
                 "Old data found in '{:s}', but has less columns ({:d})"
                 " than new input parameters grid ({:d});"
                 " continuing with grid search.".format(
-                    self.out_file, np.shape(old_data)[1], np.shape(new_data)[1]
+                    self.out_file, np.shape(old_data)[1], np.shape(self.input_data)[1]
                 )
             )
             return False
         # not yet explicitly testing the case of
-        # np.shape(old_data)[1] >= np.shape(new_data)[1]
+        # len(old_data.dtype) >= len(self.input_data.dtype)
         # because output file can have detstat and post-proc quantities
         # added and hence have different number of dimensions
         # (this could in principle be cleverly predicted at this point)
@@ -269,22 +271,24 @@ class GridSearch(BaseSearchClass):
         rtol, atol = self._get_tolerance_from_savetxt_fmt()
         column_matches = [
             np.allclose(
-                old_data[:, n],
-                new_data[:, n],
-                rtol=rtol[n],
-                atol=atol[n],
+                old_data[key],
+                self.input_data[key],
+                rtol=rtol[key],
+                atol=atol[key],
             )
-            for n in range(len(self.coord_arrays))
+            for key in self.search_keys
         ]
         if np.all(column_matches):
             logging.info(
                 "Old data found in '{:s}' with matching input parameters grid,"
-                " no search performed.".format(self.out_file)
+                " no search performed. Data grid size: {:d}x{:d}".format(
+                    self.out_file, len(old_data), len(old_data.dtype)
+                )
             )
             return old_data
         else:
             logging.info(
-                "Old data found in '{:s}', input parameters grid differs,h"
+                "Old data found in '{:s}', input parameters grid differs,"
                 "  continuing with grid search.".format(self.out_file)
             )
             return False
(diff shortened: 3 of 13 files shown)


examples/grid_examples/PyFstat_example_grid_search_F0.py
```diff
@@ -69,11 +69,10 @@
...
 search.run()
 
+print("Plotting 2F(F0)...")
 fig, ax = plt.subplots()
-xidx = search.keys.index("F0")
-frequencies = np.unique(search.data[:, xidx])
-twoF = search.data[:, -1]
-
+frequencies = search.data["F0"]
+twoF = search.data["twoF"]
```
dbkeitel, line 75: this is the basic example for why the change to the internal data structures makes usage easier

pyfstat/core.py
```diff
@@ -851,8 +851,6 @@ def estimate_min_max_CoverFreq(self):
 
     def get_fullycoherent_twoF(
         self,
-        tstart,
-        tend,
         F0,
```
dbkeitel, line 854: It might be a good further step to change this function to **all keyword args only**, so that there can be no confusion about argument order.

pyfstat/core.py
```diff
@@ -863,8 +861,11 @@ def get_fullycoherent_twoF(
         ecc=None,
         tp=None,
         argp=None,
+        tstart=None,
+        tend=None,
     ):
         """ Returns twoF or ln(BSGL) fully-coherently at a single point """
+        # print("[F0,F1,F2,Alpha,Delta,tstart,tend] = ", [F0,F1,F2,Alpha,Delta,tstart,tend])
```
dbkeitel, line 868: oops, surviving debug statement, will kill on next push

pyfstat/grid_based_searches.py
```diff
@@ -101,14 +101,13 @@ def __init__(
             os.mkdir(outdir)
         self.set_out_file()
         self.search_keys = ["F0", "F1", "F2", "Alpha", "Delta"]
-        self.input_keys = ["minStartTime", "maxStartTime"] + self.search_keys
```
dbkeitel, line 104: this one seemed superfluous

pyfstat/grid_based_searches.py
```diff
@@ -101,14 +101,13 @@ def __init__(
...
-        self.keys = self.input_keys.copy()
+        self.output_keys = self.search_keys.copy()
```
dbkeitel, line 104: And `keys` seemed too ambiguous, what it actually already was was the list for the output `data` array, so I renamed it appropriately.

pyfstat/grid_based_searches.py
```diff
@@ -119,7 +118,7 @@ def _get_search_ranges(self):
         else:
             return None
 
-    def inititate_search_object(self):
+    def _initiate_search_object(self):
```
dbkeitel, line 121: I don't think this should ever be called directly by an outside user.

pyfstat/grid_based_searches.py
```diff
@@ -157,11 +156,11 @@ def inititate_search_object(self):
...
-
-            def cut_out_tstart_tend(*vals):
-                return self.search.get_semicoherent_det_stat(*vals[2:])
-
-            self.search.get_det_stat = cut_out_tstart_tend
+            self.search.get_det_stat = self.search.get_semicoherent_det_stat
+        # make sure to overwrite the min/max starttime in case the user
+        # passed None and they were read from SFTs
+        self.minStartTime = self.search.minStartTime
+        self.maxStartTime = self.search.maxStartTime
```
dbkeitel, line 163: This is what should solve #167 and was impossible to do without the bigger changes or a similarly huge reordering.

pyfstat/grid_based_searches.py
```diff
@@ -157,11 +156,11 @@ def inititate_search_object(self):
                 detectors=self.detectors,
                 injectSources=self.injectSources,
             )
-
-            def cut_out_tstart_tend(*vals):
-                return self.search.get_semicoherent_det_stat(*vals[2:])
```
dbkeitel, line 162: ugly and no longer needed :)

pyfstat/grid_based_searches.py
```diff
@@ -229,29 +234,29 @@ def check_old_data_is_okay_to_use(self):
             )
 
         logging.info("Loading old data from '{:s}'.".format(self.out_file))
-        old_data = np.atleast_2d(np.genfromtxt(self.out_file, delimiter=" "))
-        # need to convert any "None" entries in input_data array safely to 0s
-        # to make np.allclose() work reliably
-        new_data = np.nan_to_num(self.input_data.astype(np.float64))
+        old_data = helper_functions.read_txt_file_with_header(self.out_file, names=True)
+        # old_data = np.atleast_2d(old_data) # FIXME do we still need this?
```
dbkeitel, line 238: TODO: check a single-point test

pyfstat/grid_based_searches.py
```diff
@@ -271,48 +276,51 @@ def run(self, return_data=False):
...
+        )
+        data = np.zeros(len(self.input_data), dtype=output_dtype)
+        for n, vals in enumerate(
+            tqdm(iterable, total=getattr(self, "total_iterations", None))
+        ):
             detstat = self.search.get_det_stat(*vals)
             thisCand = list(vals) + [detstat]
-            data.append(thisCand)
+            for k, key in enumerate(self.output_keys):
+                data[key][n] = thisCand[k]
```
dbkeitel, line 294: I didn't find a nice one-liner for this due to the general clunkiness of numpy structured arrays. Maybe we want to check timing on this.

pyfstat/grid_based_searches.py
```diff
@@ -774,22 +778,32 @@ def run(self, return_data=False):
...
 
-        data = np.array(data, dtype=np.float)
         if return_data:
             return data
         else:
-            self.save_array_to_disk(data)
             self.data = data
+            self.save_array_to_disk()
+
+    def get_transient_fstat_map_filename(self, param_point):
```
dbkeitel, line 787: This was supposed to be its own commit, but probably too bothersome to extract now that I've somehow squashed it in... should amend commit message to acknowledge the addition, though.

pyfstat/grid_based_searches.py
```diff
@@ -1132,17 +1145,15 @@ def __init__(
...
-        fmt += ["%.9g"]  # for detection statistic
+        fmt = helper_functions.get_doppler_params_output_format(self.output_keys)
+        fmt["delta_F0"] = "%.16g"
+        fmt["delta_F1"] = "%.16g"
+        fmt["tglitch"] = "%d"
+        fmt[self.detstat] = "%.9g"
         return fmt
 
 
+# FIXME below here not fixed for [minStartTime,maxStartTime] removal yet
```
dbkeitel, line 1156: TODO I'm not yet sure enough what those classes are supposed to be doing to fix them...

pyfstat/mcmc_based_searches.py
```diff
@@ -283,28 +283,17 @@ def logp(self, theta_vals, theta_prior, theta_keys, search):
         return np.sum(H)
 
     def logl(self, theta, search):
+        in_theta = copy.copy(self.fixed_theta)
         for j, theta_i in enumerate(self.theta_idxs):
-            self.fixed_theta[theta_i] = theta[j]
```
dbkeitel, line 287: It seemed really ugly to overwrite `self.fixed_theta` in each iteration, so I cleaned this up along the way, here and for several child classes.

pyfstat/mcmc_based_searches.py
```diff
@@ -283,28 +283,17 @@ def logp(self, theta_vals, theta_prior, theta_keys, search):
...
+        twoF = search.get_fullycoherent_twoF(*in_theta)
         return twoF / 2.0 + self.likelihoodcoef
 
     def _unpack_input_theta(self):
-        full_theta_keys = ["F0", "F1", "F2", "Alpha", "Delta"]
-        if self.binary:
-            full_theta_keys += ["asini", "period", "ecc", "tp", "argp"]
-        full_theta_keys_copy = copy.copy(full_theta_keys)
-
-        full_theta_symbols = [
```
dbkeitel, line 299: also removed redundant definitions of these

pyfstat/mcmc_based_searches.py
```diff
@@ -283,28 +283,17 @@ def logp(self, theta_vals, theta_prior, theta_keys, search):
...
-            r"$f$",
-            r"$\dot{f}$",
-            r"$\ddot{f}$",
-            r"$\alpha$",
-            r"$\delta$",
-        ]
+        self.full_theta_keys = ["F0", "F1", "F2", "Alpha", "Delta"]
```
dbkeitel, line 293: now storing these in `self` so they can be indexed over

pyfstat/mcmc_based_searches.py
```diff
@@ -3288,40 +3284,25 @@ def _initiate_search_object(self):
...
+        for j, theta_i in enumerate(self.theta_idxs):
+            in_theta[theta_i] = theta[j]
+        t0idx = self.full_theta_keys.index("transient_tstart")
+        t1idx = self.full_theta_keys.index("transient_duration")
+        in_theta[t1idx] = in_theta[t0idx] + in_theta[t1idx]  # endtime = t0+tau
+        if in_theta[t1idx] > self.maxStartTime:
             return -np.inf
-        twoF = search.get_fullycoherent_twoF(*in_theta)
+        twoF = search.get_fullycoherent_twoF(
+            *in_theta[:-2], tstart=in_theta[t0idx], tend=in_theta[t1idx]
```
dbkeitel, line 3296: TODO: one surviving instance of hard-coded numeric indices, I should fix that

dbkeitel:
Solved rebase conflicts against current master. So far it still fails two tests:
```
FAILED tests.py::TestComputeFstat::test_cumulative_twoF - RuntimeError: Internal function call failed: Invalid argument
FAILED tests.py::TestGridSearch::test_sliding_window - AttributeError: 'FrequencySlidingWindow' object has no attribute 'output_keys'
```
and probably some examples.

Code needs a thorough self-review too.

pyfstat/mcmc_based_searches.py
```diff
@@ -3322,40 +3318,34 @@ def _initiate_search_object(self):
...
+        )
+        # FIXME: this can be simplified when changing all theta lists to dicts
+        t0idx = self.theta_keys.index("transient_tstart")
+        t1idx = self.theta_keys.index("transient_duration")
+        in_theta["tstart"] = theta[t0idx]
+        in_theta["tend"] = theta[t0idx] + theta[t1idx]  # endtime = t0+tau
+        if in_theta["tend"] > self.maxStartTime:
             return -np.inf
-        twoF = search.get_fullycoherent_twoF(*in_theta)
+        twoF = search.get_fullycoherent_twoF(**in_theta)
```
dbkeitel, line 3340: Ideally this is how I think the function should always be called, and a follow-up PR could change the whole MCMC module to always use dicts and no lists for all its `theta` needs and then do it like here.

pyfstat/grid_based_searches.py
dbkeitel: proper solution: fix `helper_functions.read_txt_file_with_header`

dbkeitel:
After rebasing on #215 which removes `FrequencySlidingWindow` all tests pass.
However, [running examples](https://github.com/PyFstat/PyFstat/runs/1333945558?check_suite_focus=true) gives two failures:
```
Failed to run 2/23 example scripts: ['/home/runner/work/PyFstat/PyFstat/examples/transient_examples/PyFstat_example_long_transient_MCMC_search.py', '/home/runner/work/PyFstat/PyFstat/examples/mcmc_examples/PyFstat_example_fully_coherent_MCMC_search.py.bk']
```

dbkeitel:
* `PyFstat_example_fully_coherent_MCMC_search.py.bk` was obviously a spurious file that shouldn't have been in the commit
* `PyFstat_example_long_transient_MCMC_search.py` failed because it searches over tau only (fixed t0), a situation which was not covered in the tests. I've now fixed this and extended the tests.

dbkeitel:
@Rodrigo-Tenorio Now here's the big monster to review. I've already left a few comments on crucial parts of the diff, but there's a lot more to it...

Rodrigo-Tenorio (commented):
Summary of this changes: Make `tstart` and `tend` optional in `get_fullycoherent_fstat`. This make sense, since these options are not what one would think they are, since a window needs to be specified by the time this is called.

pyfstat/helper_functions.py
```diff
@@ -478,23 +477,21 @@ def get_doppler_params_output_format(keys):
         "tp",
         "argp",
     ]
-
-    for k in keys:
-        if k in doppler_keys:
-            fmt += [CFSv2_fmt]
+    fmt = {k: CFSv2_fmt for k in keys if k in doppler_keys}
```
Rodrigo-Tenorio, line 480: Reminder to check where is this function used

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: Is fixed_theta a dictionary? Why not just `self.fixed_theta.copy()`?

pyfstat/mcmc_based_searches.py
```diff
@@ -284,28 +284,17 @@ def logp(self, theta_vals, theta_prior, theta_keys, search):
...
 
     def logl(self, theta, search):
+        in_theta = copy.copy(self.fixed_theta)
         for j, theta_i in enumerate(self.theta_idxs):
-            self.fixed_theta[theta_i] = theta[j]
-        twoF = search.get_fullycoherent_twoF(
-            self.minStartTime, self.maxStartTime, *self.fixed_theta
-        )
+            in_theta[theta_i] = theta[j]
+        twoF = search.get_fullycoherent_twoF(*in_theta)
```
Rodrigo-Tenorio, line 290: I feel quite uncomfortable having this call without keywords; but you said that would be a future rework, so fine by now.

pyfstat/mcmc_based_searches.py
```diff
@@ -1731,15 +1720,15 @@ def _check_old_data_is_okay_to_use(self):
...
+        fmt["twoF"] = "%.9g"
         return fmt
 
     def export_samples_to_disk(self):
         self.samples_file = os.path.join(self.outdir, self.label + "_samples.dat")
         logging.info("Exporting samples to {}".format(self.samples_file))
         header = "\n".join(self.output_file_header)
         header += "\n" + " ".join(self.theta_keys) + " twoF"
-        outfmt = self.get_savetxt_fmt()
+        outfmt = list(self.get_savetxt_fmt().values())
```
Rodrigo-Tenorio, line 1731: Ok!

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: Agree

pyfstat/mcmc_based_searches.py
```diff
@@ -3322,40 +3318,41 @@ def _initiate_search_object(self):
...
-        full_theta_keys = [
-            "transient_tstart",
-            "transient_duration",
-            "F0",
-            "F1",
-            "F2",
-            "Alpha",
-            "Delta",
-        ]
+        self.full_theta_keys = ["F0", "F1", "F2", "Alpha", "Delta"]
```
Rodrigo-Tenorio, line 3351: I've seen this same list hard coded too many times...

Rodrigo-Tenorio (commented):
Excluding `grid` and `tests`, changes look fine to me

pyfstat/mcmc_based_searches.py
dbkeitel: It should be a dict, but as elsewhere: future work.

pyfstat/helper_functions.py
dbkeitel:
```
grep -n get_doppler_params_output_format pyfstat/*.py
pyfstat/grid_based_searches.py:340:        fmt = helper_functions.get_doppler_params_output_format(self.output_keys)
pyfstat/grid_based_searches.py:845:        fmt = helper_functions.get_doppler_params_output_format(self.output_keys)
pyfstat/grid_based_searches.py:968:        fmt = helper_functions.get_doppler_params_output_format(self.output_keys)
pyfstat/helper_functions.py:464:def get_doppler_params_output_format(keys):
pyfstat/mcmc_based_searches.py:1722:        fmt = helper_functions.get_doppler_params_output_format(self.theta_keys)
pyfstat/mcmc_based_searches.py:2485:        fmt = helper_functions.get_doppler_params_output_format(
pyfstat/mcmc_based_searches.py:2488:        fmt = helper_functions.get_doppler_params_output_format(self.theta_keys)
pyfstat/mcmc_based_searches.py:3388:        fmt = helper_functions.get_doppler_params_output_format(self.theta_keys)
```

All use it in `dict` fashion, but `pyfstat/mcmc_based_searches.py:2485` + `pyfstat/mcmc_based_searches.py:2488` makes no sense, let me see what and why I did there.

pyfstat/mcmc_based_searches.py
```diff
@@ -2494,10 +2485,14 @@ def get_savetxt_fmt(self):
         fmt = helper_functions.get_doppler_params_output_format(
             [k.split("_")[1] for k in self.theta_keys if k.startswith("delta_F")]
         )
+        fmt = helper_functions.get_doppler_params_output_format(self.theta_keys)
```
dbkeitel, line 2488: Why am I calling this twice...? :thinking:

pyfstat/mcmc_based_searches.py
dbkeitel: Yeah, at least it's no worse than before.

pyfstat/mcmc_based_searches.py
dbkeitel: Maybe it should be an attribute of the `ComputeFStatistic` core class, `default_doppler_parameter_keys` or something...? But that should be another separate change, I'd say.

pyfstat/mcmc_based_searches.py
dbkeitel: Actually, not ok! This will result in `outfmt` in the order the dict was filled, which does not need to be the same as the order of samples columns. This currently causes a bug at least for `MCMCGlitchSearch`.

pyfstat/mcmc_based_searches.py
dbkeitel: ...so that I would have reason to fix a hidden output bug. ;)

pyfstat/helper_functions.py
dbkeitel: Duplication resolved, but it was helpful in noticing and fixing a sorting issue with the final output formats list.

pyfstat/mcmc_based_searches.py
dbkeitel: I think the `MCMCGlitchSearch` glitch output was already buggy on master, see #219, though this PR initially made it worse. ;). Should now be fixed in 662d47217d97ef528663f121f0e02f432b0e3fad but I still need to double-check if similar issues appear in the grid classes.

pyfstat/core.py
Rodrigo-Tenorio: Agree. Future work.

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: This looks like on of those "Here's yet another way of writing data down!" which #191 should take care of. I'll mark this.

pyfstat/mcmc_based_searches.py
```diff
@@ -3378,21 +3388,23 @@ def _unpack_input_theta(self):
...
         self.theta_idxs = [self.theta_idxs[i] for i in idxs]
         self.theta_symbols = [self.theta_symbols[i] for i in idxs]
         self.theta_keys = [self.theta_keys[i] for i in idxs]
 
-    def get_savetxt_fmt(self):
-        fmt = []
+        self.output_keys = self.theta_keys.copy()
+        self.output_keys += ["logBSGL" if self.BSGL else "twoF"]
+
+    def get_savetxt_fmt_dict(self):
```
Rodrigo-Tenorio, line 3403: Sweet method duplication :wink:

Rodrigo-Tenorio (commented):
I agree with your last changes related to the output. #191

pyfstat/grid_based_searches.py
```diff
@@ -101,14 +101,13 @@ def __init__(
...
-        self.keys = self.input_keys.copy()
+        self.output_keys = self.search_keys.copy()
         if self.BSGL:
             self.detstat = "BSGL"
-            self.keys += ["logBSGL"]
+            self.output_keys += ["logBSGL"]
         else:
             self.detstat = "twoF"
-            self.keys += ["twoF"]
+            self.output_keys += [self.detstat]
```
Rodrigo-Tenorio, line 110: I love the "inconsistency" of "BSGL -> logBSGL" but "twoF" is just "twoF" :wink:

pyfstat/grid_based_searches.py
Rodrigo-Tenorio: Ok. Question: Does it make sense to choose automatically from `ComputeFstat` or `SemiCoherentSearch` depending on the number of segments? I guess the answer is "Obviously yes, why would you specify the number of segments otherwise?" and "Yes, there's a logging line declaring which class is being instantiated"

pyfstat/grid_based_searches.py
```diff
@@ -195,7 +194,13 @@ def get_input_data_array(self):
...
             for vals in itertools.product(*coord_arrays):
                 input_data.append(vals)
-            self.input_data = np.array(input_data)
+            input_dtype = np.dtype(
+                {
+                    "names": self.search_keys,
+                    "formats": np.repeat(float, len(self.search_keys)),
+                }
+            )
+            self.input_data = np.array(input_data, dtype=input_dtype)
```
Rodrigo-Tenorio, line 203: Spooky, what's going on here?

pyfstat/grid_based_searches.py
Rodrigo-Tenorio: Not sure of what's going on here either. Why do you use `np.dtype`?

pyfstat/grid_based_searches.py
```diff
@@ -359,12 +366,12 @@ def _get_tolerance_from_savetxt_fmt(self):
...
 
-    def save_array_to_disk(self, data):
+    def save_array_to_disk(self):
         logging.info("Saving data to {}".format(self.out_file))
         header = "\n".join(self.output_file_header)
-        header += "\n" + " ".join(self.keys)
-        outfmt = self.get_savetxt_fmt()
-        Ncols = np.shape(data)[1]
+        header += "\n" + " ".join(self.output_keys)
+        outfmt = list(self.get_savetxt_fmt().values())
```
Rodrigo-Tenorio, line 373: There was an "actually not Ok" regarding a similar line. I'm noticing quite a volume of duplicated lines (that's what we have, I guess...)

pyfstat/grid_based_searches.py
```diff
@@ -750,22 +763,22 @@ def run(self, return_data=False):
...
-                "Will save per-Doppler Fstatmap"
-                " results to {}*.dat".format(self.tCWfilebase)
-            )
+        output_dtype = np.dtype(
+            {
+                "names": self.output_keys,
+                "formats": np.repeat(float, len(self.output_keys)),
+            }
+        )
+        data = np.zeros(len(self.input_data), dtype=output_dtype)
```
Rodrigo-Tenorio, line 774: Same as before, what?

Rodrigo-Tenorio (commented):
Summary: Why the `np.dtype`?

Rodrigo-Tenorio (commented):
Regarding `tests`, essentially accommodating the changes and adding some testing functionality for the transient module.

pyfstat/grid_based_searches.py
dbkeitel: Actually this makes sense: in the same notation as the BSGL family, F is essentially logBSG (log Bayes factor for Signal vs pure Gaussian noise). (Never mind the "two".)

pyfstat/grid_based_searches.py
dbkeitel: I think mostly to work nicely with `np.savetxt` and `np.genfromtxt`. Without setting the `dtype` manually, an `np.array` with named columns would always have them all as floats.

pyfstat/grid_based_searches.py
dbkeitel: Same logic as before, but on rereading this, I'm wondering why in this case they're still all `float` and no `int` for the `["t0", "tau"]` parameters... Hmm...

pyfstat/grid_based_searches.py
dbkeitel: for reference, that was this thread: https://github.com/PyFstat/PyFstat/pull/171#discussion_r516936239

pyfstat/grid_based_searches.py
Rodrigo-Tenorio: I'm not familiar with these parameters.

We use `int` for time whenever time is set using some variation of `LALEpoch` or `LALGPS` or that sort of thing (otherwise parsing becomes a nightmare, as seen multiple times in this package). 

Are these parameters treated as epochs in LALSuite?

dbkeitel: ...[no](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/structtransient_window_range__t.html)...

Rodrigo-Tenorio: I see, they are literally what in python would be `int` (if my understanding about swig magic is not mistaken).

dbkeitel: ...while [CFSv2](https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalapps/src/pulsar/Fstatistic/ComputeFstatistic_v2.c#L1086) takes either `transient_t0Epoch`, which is an `EPOCH` type, or `transient_t0Offset`, which is not, and that makes sense. `tau` (duration) by definition is always a length, so doesn't need to be an epoch.

Rodrigo-Tenorio:
```suggestion
        output_dtype = np.dtype(
            {
                "names": self.output_keys,
                "formats": np.array([float if key not in ["t0", "tau"] else int for key in self.output_keys]),
            }
        )
        data = np.zeros(len(self.input_data), dtype=output_dtype)
```

pyfstat/grid_based_searches.py
dbkeitel: I think it didn't cause any errors for the grid-based searches, but to be on the safe side I've replicated the same fix as for MCMC here too: d6a50ffb771107809400bb71ac6826f402c96851

pyfstat/grid_based_searches.py
dbkeitel: This does work, but has one problem: if an existing data file is found, `np.genfromtxt` (called via `helper_functions.read_txt_file_with_header`) guesses the dtypes as float (even if in the text file none of the t0/tau colum entries have any decimals, as our _output_ format for those is already `%d`). So in that case, `self.data = old_data` will have all-float dtype, while if an actual search is run, it will have `[float,float,...,int,int]`. If one just runs a search and writes the output to disk, there's no problem, as the mentioned output formatting will make the difference disappear. But if someone works with the search class interactively, this might cause some confusion.

pyfstat/grid_based_searches.py
dbkeitel: Testing and researching a bit more: and the specific reason for why I explicitly set the `formats` (i.e. actual data type) of the `dtype` as well is that I can't give it `names` alone, nor define a plain `np.array` and later change its `data.dtype.names` (-> "ValueError: there are no fields defined").

Alternatives would be to use `pandas`, or simple dicts of arrays (which would be more analogous to the MCMC class, so maybe actually a good idea!).

pyfstat/grid_based_searches.py
dbkeitel: We decided for now to leave it as is: no real problems from having them all stored as float internally. The whole thing can always get improved further in the future, but then we may want to switch to a unified dict-based storage for both grid- and MCMC-searches anyway, making extra work on this superfluous for now.

pyfstat/grid_based_searches.py
dbkeitel: see the other threads

Rodrigo-Tenorio (approved):

dbkeitel:
examples look fine: https://github.com/PyFstat/PyFstat/actions/runs/452416869

PR #174, opened 2020-09-16
add mcmc vs grid binary comparison example

Rodrigo-Tenorio:
There is no actual implementation of a Binary Grid Search, so I had to do it by hand.

It's obviously based in [this other example](https://github.com/PyFstat/PyFstat/blob/master/examples/mcmc_vs_grid_simple_example/PyFstat_example_simple_mcmc_vs_grid_comparison.py).

### DIFF
--- a/examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
+++ b/examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
@@ -55,6 +55,9 @@ print("")
 print("Performing Grid Search...")
 
 # Create ad-hoc grid and compute Fstatistic around injection point
+# There's no class supporting a binary search in the same way as
+# grid_based_searches.GridSearch, so we do it by hand constructing
+# a grid and using core.ComputeFstat.
 half_points_per_dimension = 2
 search_keys = ["period", "asini", "tp"] + ["argp", "ecc"]
 search_keys_label = [
@@ -194,7 +197,7 @@ print(
         max_grid_2F,
         ", ".join(
             [
-                "\n\t\t{1:s}: {0:.4e} ({2:.4g})%".format(
+                "\n\t\t{1:s}: {0:.4e} ({2:.4f}%)".format(
                     max_grid_parameters[search_keys.index(key)]
                     - signal_parameters[key],
                     key,
@@ -216,7 +219,7 @@ print(
         max_2F_mcmc,
         ", ".join(
             [
-                "\n\t\t{1:s}: {0:.4e} ({2:.4g})%".format(
+                "\n\t\t{1:s}: {0:.4e} ({2:.4f}%)".format(
                     max_dict_mcmc[key] - signal_parameters[key],
                     key,
                     100
@@ -235,7 +238,7 @@ print(
     "\n\tExpressed as fractions of 2sigma intervals: {:s}.".format(
         ", ".join(
             [
-                "\n\t\t{1:s}: {0:.4e} ({2:.4g})%".format(
+                "\n\t\t{1:s}: {0:.4e} ({2:.4f}%)".format(
                     stats_dict_mcmc[key]["mean"] - signal_parameters[key],
                     key,
                     100
@@ -247,7 +250,7 @@ print(
         ),
         ", ".join(
             [
-                "\n\t\t{1:s}: {0:.4e}".format(
+                "\n\t\t{1:s}: {0:.4f}%".format(
                     100
                     * np.abs(stats_dict_mcmc[key]["mean"] - signal_parameters[key])
                     / (2 * stats_dict_mcmc[key]["std"]),
@@ -263,7 +266,7 @@ print(
     "\n\tExpressed as fractions of 90% confidence intervals: {:s}.".format(
         ", ".join(
             [
-                "\n\t\t{1:s}: {0:.4e} ({2:.4g})%".format(
+                "\n\t\t{1:s}: {0:.4e} ({2:.4f}%)".format(
                     stats_dict_mcmc[key]["median"] - signal_parameters[key],
                     key,
                     100
@@ -275,7 +278,7 @@ print(
         ),
         ", ".join(
             [
-                "\n\t\t{1:s}: {0:.4e}".format(
+                "\n\t\t{1:s}: {0:.4f}%".format(
                     100
                     * np.abs(stats_dict_mcmc[key]["median"] - signal_parameters[key])
                     / (
--- a/examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
+++ b/examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
@@ -8,7 +8,11 @@ import matplotlib.pyplot as plt
 # Set to false to include eccentricity
 circular_orbit = False
 
-label = os.path.splitext(os.path.basename(__file__))[0]
+label = (
+    os.path.splitext(os.path.basename(__file__))[0] + "_circular_orbit"
+    if circular_orbit
+    else ""
+)
 outdir = os.path.join("PyFstat_example_data", label)
 
 # Parameters to generate a data set
@@ -59,7 +63,7 @@ print("Performing Grid Search...")
 # grid_based_searches.GridSearch, so we do it by hand constructing
 # a grid and using core.ComputeFstat.
 half_points_per_dimension = 2
-search_keys = ["period", "asini", "tp"] + ["argp", "ecc"]
+search_keys = ["period", "asini", "tp", "argp", "ecc"]
 search_keys_label = [
     r"$P$ [s]",
     r"$a_p$ [s]",
@@ -67,6 +71,7 @@ search_keys_label = [
     r"$\omega$ [rad]",
     r"$e$",
 ]
+
 grid_arrays = np.meshgrid(
     *[
         signal_parameters[key]
--- a/examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
+++ b/examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
@@ -34,7 +34,7 @@ signal_parameters = {
     "F2": 0,
     "Alpha": 0.5,
     "Delta": 0.5,
-    "period": 5 * 24 * 3600.0,
+    "period": 85 * 24 * 3600.0,
     "asini": 4.0,
     "tp": mid_time * 1.05,
     "argp": 0.0 if circular_orbit else 0.54,
--- a/examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
+++ b/examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
@@ -127,8 +127,7 @@ fig.savefig(os.path.join(outdir, "grid_twoF_per_dimension.pdf"))
 
 
 print("Performing MCMCSearch...")
-
-# Set up priors for the binary parameters
+# Fixed points in frequency and sky parameters
 theta_prior = {
     "F0": signal_parameters["F0"],
     "F1": signal_parameters["F1"],
@@ -137,6 +136,7 @@ theta_prior = {
     "Delta": signal_parameters["Delta"],
 }
 
+# Set up priors for the binary parameters
 for key in search_keys:
     theta_prior.update(
         {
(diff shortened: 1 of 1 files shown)


examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+circular_orbit = False
+
+label = os.path.splitext(os.path.basename(__file__))[0]
+outdir = os.path.join("PyFstat_example_data", label)
+
+# Parameters to generate a data set
+data_parameters = {
+    "sqrtSX": 1e-22,
+    "tstart": 1000000000,
+    "duration": 60 * 86400,
```
dbkeitel, line 18: may going longer break some of the degeneracies seen in plots?

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+search_keys = ["period", "asini", "tp"] + ["argp", "ecc"]
+search_keys_label = [
+    r"$P$ [s]",
+    r"$a_p$ [s]",
+    r"$t_{p}$ [s]",
+    r"$\omega$ [rad]",
+    r"$e$",
+]
+grid_arrays = np.meshgrid(
+    *[
```
dbkeitel, line 68: sometimes black style is really horrible :(
 (just venting, no action required)

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+    outdir=outdir,
+    **data_parameters,
+    **signal_parameters,
+)
+writer.make_data()
+print("")
+
+print("Performing Grid Search...")
+
+# Create ad-hoc grid and compute Fstatistic around injection point
```
dbkeitel, line 57: add a comment here that GridSearch doesn't support binaries yet, hence the manual gridding and ComputeFstat'ing?

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+        F2=signal_parameters["F2"],
+        Alpha=signal_parameters["Alpha"],
+        Delta=signal_parameters["Delta"],
+        period=point[0],
+        asini=point[1],
+        tp=point[2],
+        argp=point[3],
+        ecc=point[4],
+    )
+print(f"2Fstat computed on {grid_points.shape[0]} points")
```
dbkeitel, line 106: :cookie: for first use of f-strings in PyFStat!

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+    **signal_parameters,
+)
+writer.make_data()
+print("")
+
+print("Performing Grid Search...")
+
+# Create ad-hoc grid and compute Fstatistic around injection point
+half_points_per_dimension = 2
+search_keys = ["period", "asini", "tp"] + ["argp", "ecc"]
```
dbkeitel, line 59: should the second part be only on `not circular_orbit`?

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+    )
+if circular_orbit:
+    for key in ["ecc", "argp"]:
+        theta_prior[key] = 0
+        search_keys.remove(key)
+# theta_prior["tp"]["lower"] = signal_parameters["tp"] - 0.5 * signal_parameters["period"]
+# theta_prior["tp"]["upper"] = (
+#    signal_parameters["tp"]
+#    + 0.5 * (1 + signal_parameters["argp"] / np.pi) * signal_parameters["period"]
+# )
```
dbkeitel, line 150: status of this block?

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+mcmcsearch.plot_corner(truths=signal_parameters)
+print("Making MCMCSearch prior-posterior comparison plot...")
+mcmcsearch.plot_prior_posterior(injection_parameters=signal_parameters)
+print("")
+
+print("*" * 20)
+print("Quantitative comparisons:")
+print("*" * 20)
+
+# some informative command-line output comparing search results and injection
```
dbkeitel, line 189: Stopped reading here, but will run it once myself and if output is ok, will assume code behind it is ok too.

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+                    100
+                    * (stats_dict_mcmc[key]["mean"] - signal_parameters[key])
+                    / signal_parameters[key],
+                )
+                for key in search_keys
+            ]
+        ),
+        ", ".join(
+            [
+                "\n\t\t{1:s}: {0:.4e}".format(
```
dbkeitel, line 255: as these are given as percentages (note factor 100 in next line):
```suggestion
                "\n\t\t{1:s}: {0:.4f}%".format(
```

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,296 @@
...
+                    100
+                    * (stats_dict_mcmc[key]["median"] - signal_parameters[key])
+                    / signal_parameters[key],
+                )
+                for key in search_keys
+            ]
+        ),
+        ", ".join(
+            [
+                "\n\t\t{1:s}: {0:.4e}".format(
```
dbkeitel, line 283: again percentages:
```suggestion
                "\n\t\t{1:s}: {0:.4f}%".format(
```

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
Rodrigo-Tenorio: That depends on how much time we want to wait: Using the current set up, it already takes ~5 minutes.

Maybe we can increase length and reduce the number of MCMC steps?

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
Rodrigo-Tenorio:
```suggestion
# Create ad-hoc grid and compute Fstatistic around injection point
# This is manually implemented as GridSearch does not support binary
# searches as of now.
```

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
Rodrigo-Tenorio: I'll polish that part a bit. I literally implemented the circular orbit as a "put ifs everywhere" (i.e. I didn't think about it in the beginning), so this can be done more elegantly.

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
Rodrigo-Tenorio: Excuse me, I though Black already decided for you that this is the best style. Period. ( :wink: )

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
Rodrigo-Tenorio: Pending removal.

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
Rodrigo-Tenorio: I agree, I used `e` because I didn't want to run something and get four zeroes. But we just checked it makes sense this way, so I'll do it as such.

Rodrigo-Tenorio (commented):
I'll polish some bits of code, re-factor the last part into something more edible, and think about how to do the circular/eccentric switch.

Question @dbkeitel , what time would you like this example to spend running? Is 5 minutes too much (yes) ? Should I try to keep it at 2min at most?

dbkeitel:
> re-factor the last part into something more edible

Don't worry about that.

> Question @dbkeitel , what time would you like this example to spend running? Is 5 minutes too much (yes) ? Should I try to keep it at 2min at most?

The current online version takes 18-20 min for all examples, so adding 5 would be a bit disproportionate, but not really too much in absolute terms. 2 would be nice, 5 is ok if it can't be meaningfully cut down too much.

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
Rodrigo-Tenorio: I achieved similar results with a shorter setup.

Rodrigo-Tenorio:
@dbkeitel  Faster, clearer, tidier. 

Any other suggestion?

dbkeitel:
Do I understand right that the `circular_orbit` flag now only changes the injection parameters, but the search always covers eccentricity?

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,299 @@
...
+    a.set(xlabel=search_keys_label[ind], ylabel=r"$2 \mathcal{F}$", yscale="log")
+    a.plot(grid_points[:, ind], twoF_values, "o")
+    a.axvline(signal_parameters[search_keys[ind]], label="Injection", color="orange")
+plt.tight_layout()
+fig.savefig(os.path.join(outdir, "grid_twoF_per_dimension.pdf"))
+
+
+print("Performing MCMCSearch...")
+
+# Set up priors for the binary parameters
```
dbkeitel, line 131:
```suggestion
# Fixed points in frequency and sky parameters
```
and move current comment to line 139

Rodrigo-Tenorio:
> Do I understand right that the `circular_orbit` flag now only changes the injection parameters, but the search always covers eccentricity?

Grid Search: Set eccentricity and periastron argument to zero.
MCMCSearch: No prior, set to zero.

Yes.

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,299 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+
+# Set to false to include eccentricity
+circular_orbit = False
+
+label = (
+    os.path.splitext(os.path.basename(__file__))[0] + "_circular_orbit"
+    if circular_orbit
+    else ""
```
dbkeitel, line 14: This dumps all files into "PyFstat_example_data" if `not circular_orbit`! Missing a pair of brackets.

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
```diff
@@ -0,0 +1,299 @@
...
+dim = len(search_keys)
+fig, ax = plt.subplots(dim, 1, figsize=(10, 10))
+for ind in range(dim):
+    a = ax.ravel()[ind]
+    a.grid()
+    a.set(xlabel=search_keys_label[ind], ylabel=r"$2 \mathcal{F}$", yscale="log")
+    a.plot(grid_points[:, ind], twoF_values, "o")
+    a.axvline(signal_parameters[search_keys[ind]], label="Injection", color="orange")
+plt.tight_layout()
+fig.savefig(os.path.join(outdir, "grid_twoF_per_dimension.pdf"))
```
dbkeitel, line 126: Please make this a png for consistency with all other plots.

dbkeitel (changes requested):

examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py
Rodrigo-Tenorio: That's right, behold the dangers of the ternary operator.

dbkeitel (dismissed):

dbkeitel:
> but the [grid] search always covers eccentricity?

Just pulled and ran for a test: The grid part is very fast anyway, so I guess that's fine.

Approved so far, but out of interest I'm letting it run locally with [300,200] steps (or maybe even more after that) to see what's going on with the funny structures...

dbkeitel (approved):

Issue #182, opened 2020-09-18
Rebirth chainconsumer


Rodrigo-Tenorio:
[Chainconsumer](https://samreay.github.io/ChainConsumer/) is yet another MCMC plotting engine wrapped by PyFstat in some codes. However, we haven't actively maintained it since... a long time ago.

We must

- [x] Update Readme with this optional dependency.
- [x] Update github actions to install this optional dependency.
- [x] Check the actual well functioning of any method wrapping this package.
- [x] Use the method in some example.

Rodrigo-Tenorio:
Basic interactions with `chainconsumer` are set up.
Let's keep discussions in #183 .

Rodrigo-Tenorio:
Closed by #183

Issue #160, opened 2020-09-08
unify plotting methods in/outputs


dbkeitel:
In #159 I'm fixing various plotting methods to close their figures after saving.

Follow-up tasks:

- [ ] unify `savefig`, `save_fig`, `save` input arguments
- [ ] unify input arguments and return values for fig and/or axes

PR #183, opened 2020-09-18
Fix and update chainconsumer functionalities 

Rodrigo-Tenorio:
Implement #182

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,10 +162,14 @@ For a general introduction to installing modules, see
   and [black](https://black.readthedocs.io) style checker
   locally,
   as these checks are required to pass by the online integration pipeline.
-* Some of the [examples](./examples) require [gridcorner](https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner);
+* Some optional plotting methods depend on two additional packages,
+  and some of the [examples](./examples) require these to run:
+ [gridcorner](https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner) 
+  and [chainconsumer](https://github.com/Samreay/ChainConsumer);
 for `pip` users this is most conveniently installed by
 ```
 pip install git+https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner
+pip install chainconsumer
 ```
 * If you prefer to make your own LALSuite installation
 [from source](https://git.ligo.org/lscsoft/lalsuite/),
--- a/pyfstat/mcmc_based_searches.py
+++ b/pyfstat/mcmc_based_searches.py
@@ -990,21 +990,40 @@ class MCMCSearch(BaseSearchClass):
             Passed to chainconsumer.plotter.plot
 
         """
-
-        if "truths" in kwargs and len(kwargs["truths"]) != self.ndim:
-            logging.warning("len(Truths) != ndim, Truths will be ignored")
-            kwargs["truths"] = None
+        try:
+            import chainconsumer
+        except ImportError:
+            logging.warning(
+                "Could not import 'chainconsumer' package, please install it to use this method."
+            )
+            return
 
         samples_plt = copy.copy(self.samples)
         labels = self._get_labels(newline_units=True)
 
         samples_plt = self._scale_samples(samples_plt, self.theta_keys)
-
-        import chainconsumer
+        if "truth" in kwargs:
+            if not isinstance(kwargs["truth"], dict):
+                raise ValueError("'truth' must be a dictionary.")
+            missing_keys = np.setdiff1d(self.theta_keys, list(kwargs["truth"].keys()))
+            if len(missing_keys) > 0:
+                logging.warning(
+                    "plot_chainconsumer(): Missing keys {} in 'truth' dictionary,"
+                    " argument will be ignored.".format(missing_keys)
+                )
+                kwargs["truth"] = None
+            else:
+                parameters_in_order = np.array(
+                    [kwargs["truth"][key] for key in self.theta_keys]
+                ).reshape((1, -1))
+                kwargs["truth"] = self._scale_samples(
+                    parameters_in_order, self.theta_keys
+                ).ravel()
 
         c = chainconsumer.ChainConsumer()
         c.add_chain(samples_plt, parameters=labels)
-        c.configure(smooth=0, summary=False, sigma2d=True)
+        # We set usetex=False to avoid dependency on 'kpsewhich' TeX tool
+        c.configure(smooth=0, summary=False, sigma2d=True, usetex=False)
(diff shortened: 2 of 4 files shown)


Rodrigo-Tenorio:
@dbkeitel here's an output example.

PyFstat_example_semi_coherent_MCMC_search_chainconsumer_corner

README.md
```diff
@@ -162,10 +162,12 @@ For a general introduction to installing modules, see
   and [black](https://black.readthedocs.io) style checker
   locally,
   as these checks are required to pass by the online integration pipeline.
-* Some of the [examples](./examples) require [gridcorner](https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner);
+* Some of the [examples](./examples) require [gridcorner](https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner)
```
dbkeitel, line 165: Actually this comment was already incorrect for `gridcorner`, as it's not only called from the examples directly, but from methods actually implemented inside the PyFstat library modules.
```suggestion
* Some optional plotting methods depend on two additional packages,
  and some of the [examples](./examples) require these to run:
 [gridcorner](https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner) 
```

pyfstat/mcmc_based_searches.py
```diff
@@ -1000,11 +1000,17 @@ def plot_chainconsumer(self, save_fig=True, label_offset=0.25, dpi=300, **kwargs
 
         samples_plt = self._scale_samples(samples_plt, self.theta_keys)
 
-        import chainconsumer
+        try:
```
dbkeitel, line 1003: Please move to start of method, before it does potentially slow stuff with samples.

pyfstat/mcmc_based_searches.py
```diff
@@ -1000,11 +1000,17 @@ def plot_chainconsumer(self, save_fig=True, label_offset=0.25, dpi=300, **kwargs
...
+        except ImportError:
+            logging.warning(
+                "Could not import 'chainconsumer' package, please install it to use this method."
+            )
+            return
 
         c = chainconsumer.ChainConsumer()
         c.add_chain(samples_plt, parameters=labels)
-        c.configure(smooth=0, summary=False, sigma2d=True)
+        c.configure(smooth=0, summary=False, sigma2d=True, usetex=False)
```
dbkeitel, line 1013:
```suggestion
        # to avoid dependency on 'kpsewhich' TeX tool, we set usetex=False
        c.configure(smooth=0, summary=False, sigma2d=True, usetex=False)
```

dbkeitel (changes requested):
Looks good except for the minor comments, and that the example plot has the same axis label issues as the standard corner and gridcorner modules (let's not worry about those here -> #184).

dbkeitel:
Basic mode works, but while our `MCMCSearch.plot_chainconsumer` method claims to take a `truths` argument like the regular `corner` plots, this does not actually seem to be supported by `chainconsumer`. E.g. adding this in the `PyFstat_example_semi_coherent_MCMC_search.py` script as
```
mcmc.plot_chainconsumer(truths=signal_parameters)
```
actually leads to:
```
18:33 WARNING : len(Truths) != ndim, Truths will be ignored
  File "examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py", line 92, in <module>
    mcmc.plot_chainconsumer(truths=signal_parameters)
  File "/home/dkeitel/git/sources/PyFstat/pyfstat/mcmc_based_searches.py", line 1014, in plot_chainconsumer
    fig = c.plotter.plot(**kwargs)
TypeError: plot() got an unexpected keyword argument 'truths'
```

Should we just remove that argument, or figure out if `chainconsumer` has something equivalent?

Rodrigo-Tenorio:
I'm pretty sure that functionality is supported, so... the later, of course!

pyfstat/mcmc_based_searches.py
```diff
@@ -998,14 +998,21 @@ def plot_chainconsumer(self, save_fig=True, label_offset=0.25, dpi=300, **kwargs
...
-            logging.warning("len(Truths) != ndim, Truths will be ignored")
-            kwargs["truths"] = None
 
         samples_plt = copy.copy(self.samples)
         labels = self._get_labels(newline_units=True)
 
         samples_plt = self._scale_samples(samples_plt, self.theta_keys)
+        if "truth" in kwargs:
+            if len(kwargs["truth"]) != self.ndim:
+                logging.warning("len(Truths) != ndim, Truths will be ignored")
```
dbkeitel, line 1008: That current check is overzealous and insufficient at the same time: it causes the `truth=signal_parameters` example to be ignored (because it contains _too many_ injection parameters), but would not catch cases where the _right number_ of _completely wrong_ keys are given. Let's to this properly:

```suggestion
            if not isinstance(kwargs["truth"], dict):
                raise ValueError("'truth' must be a dictionary.")
            missing_keys = np.setdiff1d(self.theta_keys, list(kwargs["truth"].keys()))
            if len(missing_keys) > 0:
                logging.warning(
                    "plot_chainconsumer(): Missing keys {} in 'truth' dictionary,"
                    " argument will be ignored.".format(missing_keys)
                )
```

(In keeping with numpy's general apparent disregard of pythons "one and only one way to do something" mantra, there are probably many alternatives to `setdiff1d` in this context, but this one seems to work.)

pyfstat/mcmc_based_searches.py
```diff
@@ -998,14 +998,21 @@ def plot_chainconsumer(self, save_fig=True, label_offset=0.25, dpi=300, **kwargs
             )
             return
 
-        if "truths" in kwargs and len(kwargs["truths"]) != self.ndim:
```
dbkeitel, line 994: From a pure PyFstat user perspective, it would be nicer to keep calling the method argument called `truths` and internally change it to `truth` before calling `chainconsumer`. On the other hand, users already familiar with `chainconsumer` from the outside world would find that baffling. So let's go with this choice for now, and there's already #160 for general unification later on.

pyfstat/mcmc_based_searches.py
dbkeitel: Note: an analogus fix should probably be applied to plot_corner().

Rodrigo-Tenorio: This did the trick!

dbkeitel:
@Rodrigo-Tenorio flake8 is unhappy about whitespace here. While fixing that, could you also squash everything please? Don't worry about "co-authorship" from the one suggestion commit.

dbkeitel (approved):

Issue #170, opened 2020-09-15
check on PyFstat_example_twoF_cumulative.py


dbkeitel:
I'm not actually sure what this example is supposed to be achieving, but this plot doesn't look like an intended result any more:

![PyFstat_example_twoF_cumulative_twoFcumulative](https://user-images.githubusercontent.com/28396587/93216507-1f2e2a00-f768-11ea-9cbe-bf4d4752dea5.png)

Rodrigo-Tenorio:
The examples uses `plot_cumulative_max`, which falls back to some sort of `calculate_twoF_cumulative` in the base search class.

dbkeitel:
Going back to v.1.5.0 tag:

![PyFstat_example_twoF_cumulative_twoFcumulative](https://user-images.githubusercontent.com/28396587/93779425-4339b180-fc27-11ea-89a0-0f45001e93b9.png)

Rodrigo-Tenorio:
I get using 7112bca1aa2252dde25afd616b290927d0799fe2
![2Ftrial](https://user-images.githubusercontent.com/49152330/93782836-4fc00900-fc2b-11ea-903b-533bdb27c351.png)
Maybe the signal is too weak? Anyways, this is a starting point.

Rodrigo-Tenorio:
Solved in  c63c4ac .

dbkeitel:
This was broken since almost 3 years, as in 73fee7871018e72ed6ba965d8cae343ebc89f8a8 I blindly replaced some `transientWindowType`s without understanding what this function actually does. :crying_cat_face:

Rodrigo-Tenorio:
MultiIFO example of the latest version (this is still `ComputeFstat`)
![PyFstat_example_twoF_cumulative_twoFcumulative](https://user-images.githubusercontent.com/49152330/93793049-9bc47b00-fc36-11ea-8239-14d363007375.png)

Rodrigo-Tenorio:
Let's include some backwards-incompatible ugly-(but maybe useful)-looking PFS plots, shall we?
![PyFstat_example_twoF_cumulative_twoFcumulative](https://user-images.githubusercontent.com/49152330/93800516-fcf14c00-fc40-11ea-975f-beacca2b90db.png)

Rodrigo-Tenorio:
I just realizeed that band should as well be growing with time.

dbkeitel:
[...]

PR #186, opened 2020-09-21
Fix cumulative twoF

Rodrigo-Tenorio:
Make #170 work again

- [x] Fix cumulative 2F computation.
- [x] Fix basic plotting options of `ComputeFStatistic`
- [x] Fix options dealing with PFS.
- [x] Fix options dealing with injection parameters.
- [x] Fix MCMC wrapper (it applies this function to the maximum candidate).
- [x] Check behaviour of PFS when using single IFO: Check #189 , as this only affects to PFS and not the actual cumulative 2F.

And I guess @dbkeitel would like
- [x] Add corresponding tests. Proposal: Compute `2F` using the coherent method and the cumulative method; the last value of the cumulative method should match that of the coherent method.

More issues to deal with:
- [x] ~Fix~ Remove single IFO support (~choose your preferred fix method amongst `dd` or a longer key combination :wink:~ which means you have to manually give the input SFT you want to compute your cumulative twoF on.)

### DIFF
--- a/examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
+++ b/examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
@@ -91,5 +91,6 @@ mcmc.plot_corner(add_prior=True, truths=signal_parameters)
 mcmc.plot_prior_posterior(injection_parameters=signal_parameters)
 mcmc.plot_chainconsumer(truth=signal_parameters)
 mcmc.plot_cumulative_max(
-        savefig=True, custom_ax_kwargs={"title": "Cumulative 2F for the best MCMC candidate"},
+    savefig=True,
+    custom_ax_kwargs={"title": "Cumulative 2F for the best MCMC candidate"},
 )
--- a/examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
+++ b/examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
@@ -90,3 +90,6 @@ mcmc.print_summary()
 mcmc.plot_corner(add_prior=True, truths=signal_parameters)
 mcmc.plot_prior_posterior(injection_parameters=signal_parameters)
 mcmc.plot_chainconsumer(truth=signal_parameters)
+mcmc.plot_cumulative_max(
+    plt_label="MCMC Example", title="Cumulative 2F for the best MCMC candidate"
+)
--- a/examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
+++ b/examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
@@ -91,5 +91,7 @@ mcmc.plot_corner(add_prior=True, truths=signal_parameters)
 mcmc.plot_prior_posterior(injection_parameters=signal_parameters)
 mcmc.plot_chainconsumer(truth=signal_parameters)
 mcmc.plot_cumulative_max(
-    plt_label="MCMC Example", title="Cumulative 2F for the best MCMC candidate"
+    plot_label="MCMC Example",
+    # title="Cumulative 2F for the best MCMC candidate",
+    custom_axis_kwargs={"title": "Cumulative 2F for the best MCMC candidate"},
 )
--- a/examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
+++ b/examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
@@ -91,7 +91,5 @@ mcmc.plot_corner(add_prior=True, truths=signal_parameters)
 mcmc.plot_prior_posterior(injection_parameters=signal_parameters)
 mcmc.plot_chainconsumer(truth=signal_parameters)
 mcmc.plot_cumulative_max(
-    plot_label="MCMC Example",
-    # title="Cumulative 2F for the best MCMC candidate",
-    custom_axis_kwargs={"title": "Cumulative 2F for the best MCMC candidate"},
+        savefig=True, custom_ax_kwargs={"title": "Cumulative 2F for the best MCMC candidate"},
 )
--- a/examples/other_examples/PyFstat_example_twoF_cumulative.py
+++ b/examples/other_examples/PyFstat_example_twoF_cumulative.py
@@ -26,9 +26,14 @@ signal_parameters = {
     "tref": gw_data["tstart"],
     "cosi": 1,
     "h0": gw_data["sqrtSX"] / depth,
+    "asini": 10,
+    "period": 10 * 3600 * 24,
+    "tp": gw_data["tstart"] + gw_data["duration"] / 2.0,
+    "ecc": 0,
+    "argp": 0,
 }
 
-data = pyfstat.Writer(
+data = pyfstat.BinaryModulatedWriter(
     label=label,
     outdir=outdir,
     **gw_data,
--- a/examples/other_examples/PyFstat_example_twoF_cumulative.py
+++ b/examples/other_examples/PyFstat_example_twoF_cumulative.py
@@ -43,12 +43,14 @@ print("Predicted twoF value: {}\n".format(twoF))
 ifo_constraints = ["L1", "H1", None]
 compute_fstat_per_ifo = [
     pyfstat.ComputeFstat(
-        sftfilepattern=data.sftfilepath,
+        sftfilepattern=os.path.join(
+            data.outdir,
+            (f"{ifo_constraint[0]}*.sft" if ifo_constraint is not None else "*.sft"),
+        ),
         tref=signal_parameters["tref"],
         binary=signal_parameters.get("asini", 0),
         minCoverFreq=-0.5,
         maxCoverFreq=-0.5,
-        detectors=ifo_constraint,
     )
     for ifo_constraint in ifo_constraints
 ]
--- a/examples/other_examples/PyFstat_example_twoF_cumulative.py
+++ b/examples/other_examples/PyFstat_example_twoF_cumulative.py
@@ -2,6 +2,8 @@ import os
 import numpy as np
 import pyfstat
 
+from pyfstat.helper_functions import get_predict_fstat_parameters_from_dict
+
 label = os.path.splitext(os.path.basename(__file__))[0]
 outdir = os.path.join("PyFstat_example_data", label)
 
@@ -17,27 +19,39 @@ gw_data = {
 
 # Properties of the signal
 depth = 100
-signal_parameters = {
+phase_parameters = {
     "F0": 30.0,
     "F1": -1e-10,
     "F2": 0,
     "Alpha": np.radians(83.6292),
     "Delta": np.radians(22.0144),
     "tref": gw_data["tstart"],
-    "cosi": 1,
-    "h0": gw_data["sqrtSX"] / depth,
     "asini": 10,
     "period": 10 * 3600 * 24,
     "tp": gw_data["tstart"] + gw_data["duration"] / 2.0,
     "ecc": 0,
     "argp": 0,
 }
+amplitude_parameters = {
+    "h0": gw_data["sqrtSX"] / depth,
+    "cosi": 1,
+    "phi": np.pi,
+    "psi": np.pi / 8,
+}
 
+PFS_input = get_predict_fstat_parameters_from_dict(
+    {**phase_parameters, **amplitude_parameters}
+)
+
+# Let me grab tref here, since it won't really be needed in phase_parameters
+tref = phase_parameters.pop("tref")
 data = pyfstat.BinaryModulatedWriter(
     label=label,
     outdir=outdir,
+    tref=tref,
     **gw_data,
-    **signal_parameters,
+    **phase_parameters,
+    **amplitude_parameters,
 )
 data.make_data()
 
@@ -45,6 +59,8 @@ data.make_data()
 twoF = data.predict_fstat()
 print("Predicted twoF value: {}\n".format(twoF))
 
+# Create a search object for each of the possible SFT combinations
+# (H1 only, L1 only, H1 + L1).
 ifo_constraints = ["L1", "H1", None]
 compute_fstat_per_ifo = [
     pyfstat.ComputeFstat(
@@ -52,39 +68,24 @@ compute_fstat_per_ifo = [
             data.outdir,
             (f"{ifo_constraint[0]}*.sft" if ifo_constraint is not None else "*.sft"),
         ),
-        tref=signal_parameters["tref"],
-        binary=signal_parameters.get("asini", 0),
+        tref=data.tref,
+        binary=phase_parameters.get("asini", 0),
         minCoverFreq=-0.5,
         maxCoverFreq=-0.5,
     )
     for ifo_constraint in ifo_constraints
 ]
 
-cumulative_f_stat_params = {
-    key: signal_parameters[key]
-    for key in signal_parameters
-    if key
-    in ["F0", "F1", "F2", "Alpha", "Delta", "asini", "period", "tp", "ecc", "argp"]
-}
-cumulative_f_stat_params.update(
-    {"tstart": gw_data["tstart"], "tend": gw_data["tstart"] + gw_data["duration"]}
-)
-
-predict_f_stat_params = {
-    key: signal_parameters.get(key, 0)
-    for key in ["F0", "Alpha", "Delta", "psi", "cosi", "h0"]
-}
-
 for ind, compute_f_stat in enumerate(compute_fstat_per_ifo):
-    # taus, twoF = compute_f_stat.calculate_twoF_cumulative(**cumulative_f_stat_params)
     compute_f_stat.plot_twoF_cumulative(
         label=label + (f"_{ifo_constraints[ind]}" if ind < 2 else "_H1L1"),
         outdir=outdir,
-        signal_parameters=predict_f_stat_params,
+        savefig=True,
+        CFS_input=phase_parameters,
+        PFS_input=PFS_input,
         custom_ax_kwargs={
             "title": "How does 2F accumulate over time?",
             "label": "Cumulative 2F"
             + (f" {ifo_constraints[ind]}" if ind < 2 else " H1 + L1"),
         },
-        **cumulative_f_stat_params,
     )
--- a/examples/other_examples/PyFstat_example_twoF_cumulative.py
+++ b/examples/other_examples/PyFstat_example_twoF_cumulative.py
@@ -81,8 +81,10 @@ for ind, compute_f_stat in enumerate(compute_fstat_per_ifo):
         label=label + (f"_{ifo_constraints[ind]}" if ind < 2 else "_H1L1"),
         outdir=outdir,
         signal_parameters=predict_f_stat_params,
-        custom_axis_kwargs={"title": "How does 2F accumulate over time?"},
-        plot_label="Cumulative 2F"
-        + (f" {ifo_constraints[ind]}" if ind < 2 else " H1 + L1"),
+        custom_ax_kwargs={
+            "title": "How does 2F accumulate over time?",
+            "label": "Cumulative 2F"
+            + (f" {ifo_constraints[ind]}" if ind < 2 else " H1 + L1"),
+        },
         **cumulative_f_stat_params,
     )
--- a/examples/other_examples/PyFstat_example_twoF_cumulative.py
+++ b/examples/other_examples/PyFstat_example_twoF_cumulative.py
@@ -1,43 +1,38 @@
-import pyfstat
-import numpy as np
 import os
+import numpy as np
+import pyfstat
 
 label = os.path.splitext(os.path.basename(__file__))[0]
 outdir = os.path.join("PyFstat_example_data", label)
 
 # Properties of the GW data
-sqrtSX = 1e-23
-tstart = 1000000000
-duration = 100 * 86400
-tend = tstart + duration
+gw_data = {
+    "sqrtSX": 1e-23,
+    "tstart": 1000000000,
+    "duration": 100 * 86400,
+    "detectors": "H1,L1",
+    "Band": 4,
+    "Tsft": 1800,
+}
 
 # Properties of the signal
-F0 = 30.0
-F1 = -1e-10
-F2 = 0
-Alpha = np.radians(83.6292)
-Delta = np.radians(22.0144)
-tref = 0.5 * (tstart + tend)
-cosi = 0
-
 depth = 100
-h0 = sqrtSX / depth
+signal_parameters = {
+    "F0": 30.0,
+    "F1": -1e-10,
+    "F2": 0,
+    "Alpha": np.radians(83.6292),
+    "Delta": np.radians(22.0144),
+    "tref": gw_data["tstart"],
+    "cosi": 1,
+    "h0": gw_data["sqrtSX"] / depth,
+}
 
 data = pyfstat.Writer(
     label=label,
     outdir=outdir,
-    tref=tref,
-    tstart=tstart,
-    F0=F0,
-    F1=F1,
-    F2=F2,
-    duration=duration,
-    Alpha=Alpha,
-    Delta=Delta,
-    h0=h0,
-    cosi=cosi,
-    sqrtSX=sqrtSX,
-    detectors="H1,L1",
+    **gw_data,
+    **signal_parameters,
 )
 data.make_data()
 
@@ -45,41 +40,42 @@ data.make_data()
 twoF = data.predict_fstat()
 print("Predicted twoF value: {}\n".format(twoF))
 
-DeltaF0 = 1e-7
-DeltaF1 = 1e-13
-VF0 = (np.pi * duration * DeltaF0) ** 2 / 3.0
-VF1 = (np.pi * duration ** 2 * DeltaF1) ** 2 * 4 / 45.0
-print("\nV={:1.2e}, VF0={:1.2e}, VF1={:1.2e}\n".format(VF0 * VF1, VF0, VF1))
+ifo_constraints = ["L1", "H1", None]
+compute_fstat_per_ifo = [
+    pyfstat.ComputeFstat(
+        sftfilepattern=data.sftfilepath,
+        tref=signal_parameters["tref"],
+        binary=signal_parameters.get("asini", 0),
+        minCoverFreq=-0.5,
+        maxCoverFreq=-0.5,
+        detectors=ifo_constraint,
+    )
+    for ifo_constraint in ifo_constraints
+]
 
-theta_prior = {
-    "F0": {"type": "unif", "lower": F0 - DeltaF0 / 2.0, "upper": F0 + DeltaF0 / 2.0},
-    "F1": {"type": "unif", "lower": F1 - DeltaF1 / 2.0, "upper": F1 + DeltaF1 / 2.0},
-    "F2": F2,
-    "Alpha": Alpha,
-    "Delta": Delta,
+cumulative_f_stat_params = {
+    key: signal_parameters[key]
+    for key in signal_parameters
+    if key
+    in ["F0", "F1", "F2", "Alpha", "Delta", "asini", "period", "tp", "ecc", "argp"]
+}
+cumulative_f_stat_params.update(
+    {"tstart": gw_data["tstart"], "tend": gw_data["tstart"] + gw_data["duration"]}
+)
+
+predict_f_stat_params = {
+    key: signal_parameters.get(key, 0)
+    for key in ["F0", "Alpha", "Delta", "psi", "cosi", "h0"]
 }
 
-ntemps = 1
-log10beta_min = -1
-nwalkers = 100
-nsteps = [50, 50]
-
-mcmc = pyfstat.MCMCSearch(
-    label=label,
-    outdir=outdir,
-    sftfilepattern=os.path.join(outdir, "*" + label + "*sft"),
-    theta_prior=theta_prior,
-    tref=tref,
-    minStartTime=tstart,
-    maxStartTime=tend,
-    nsteps=nsteps,
-    nwalkers=nwalkers,
-    ntemps=ntemps,
-    log10beta_min=log10beta_min,
-)
-mcmc.run()
-mcmc.plot_corner(add_prior=True)
-mcmc.print_summary()
-
-mcmc.generate_loudest()
-mcmc.plot_cumulative_max(add_pfs=True)
+for ind, compute_f_stat in enumerate(compute_fstat_per_ifo):
+    taus, twoF = compute_f_stat.calculate_twoF_cumulative(**cumulative_f_stat_params)
+    compute_f_stat.plot_twoF_cumulative(
+        **cumulative_f_stat_params,
+        label=label + (f"_{ifo_constraints[ind]}" if ind < 2 else "_H1L1"),
+        outdir=outdir,
+        injection_parameters=predict_f_stat_params,
+        plot_label="Cumulative 2F"
+        + (f" {ifo_constraints[ind]}" if ind < 2 else " H1 + L1"),
+        custom_axis_kwargs={"title": "This is a custom title"},
+    )
--- a/examples/other_examples/PyFstat_example_twoF_cumulative.py
+++ b/examples/other_examples/PyFstat_example_twoF_cumulative.py
@@ -76,13 +76,13 @@ predict_f_stat_params = {
 }
 
 for ind, compute_f_stat in enumerate(compute_fstat_per_ifo):
-    taus, twoF = compute_f_stat.calculate_twoF_cumulative(**cumulative_f_stat_params)
+    # taus, twoF = compute_f_stat.calculate_twoF_cumulative(**cumulative_f_stat_params)
     compute_f_stat.plot_twoF_cumulative(
-        **cumulative_f_stat_params,
         label=label + (f"_{ifo_constraints[ind]}" if ind < 2 else "_H1L1"),
         outdir=outdir,
-        injection_parameters=predict_f_stat_params,
+        signal_parameters=predict_f_stat_params,
+        custom_axis_kwargs={"title": "How does 2F accumulate over time?"},
         plot_label="Cumulative 2F"
         + (f" {ifo_constraints[ind]}" if ind < 2 else " H1 + L1"),
-        custom_axis_kwargs={"title": "This is a custom title"},
+        **cumulative_f_stat_params,
     )
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1079,8 +1079,7 @@ class ComputeFstat(BaseSearchClass):
         outdir,
         signal_parameters=None,
         predict_fstat_segments=15,
-        custom_axis_kwargs=None,
-        plot_label=None,
+        custom_ax_kwargs=None,
         savefig=True,
         **calculate_twoF_cumulative_kwargs,
     ):
@@ -1099,7 +1098,7 @@ class ComputeFstat(BaseSearchClass):
             see _calculate_predict_fstat_cumulative() for details.
         predict_fstat_segments : int
             Number of points to use for PredictFStat.
-        custom_axis_kwargs : dict
+        custom_ax_kwargs : dict
             Optional axis formatting options.
         plot_label : str
             Legend label for the cumulative 2F values computed from data.
@@ -1124,6 +1123,15 @@ class ComputeFstat(BaseSearchClass):
         taus_days = taus / 86400.0
 
         tstart = calculate_twoF_cumulative_kwargs.get("tstart", self.minStartTime)
+
+        plot_label = custom_ax_kwargs.pop(
+            "label",
+            (
+                f"Cumulative 2F {taus.shape[0]:d} segments"
+                f" ({(taus_days[1] - taus_days[0]):.2g} days per segment)"
+            ),
+        )
+
         axis_kwargs = {
             "xlabel": r"Days from $t_{{\rm start}}={:.0f}$".format(tstart),
             "ylabel": r"$\log_{10}(\mathrm{BSGL})_{\rm cumulative}$"
@@ -1131,19 +1139,14 @@ class ComputeFstat(BaseSearchClass):
             else r"$\widetilde{2\mathcal{F}}_{\rm cumulative}$",
             "xlim": (0, taus_days[-1]),
         }
-        if custom_axis_kwargs is not None:
+        if custom_ax_kwargs is not None:
             for kwarg in "xlabel", "ylabel":
-                if kwarg in custom_axis_kwargs:
+                if kwarg in custom_ax_kwargs:
                     logging.warning(
                         f"Be careful, overwriting {kwarg} {axis_kwargs[kwarg]}"
-                        " with {custom_axis_kwargs[kwarg]}: Check out the units!"
+                        " with {custom_ax_kwargs[kwarg]}: Check out the units!"
                     )
-            axis_kwargs.update(custom_axis_kwargs or {})
-
-        plot_label = plot_label or (
-            f"Cumulative 2F {taus.shape[0]:d} segments"
-            f" ({(taus_days[1] - taus_days[0]):.2g} days per segment)"
-        )
+            axis_kwargs.update(custom_ax_kwargs or {})
 
         fig, ax = plt.subplots()
         ax.grid()
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1058,7 +1058,7 @@ class ComputeFstat(BaseSearchClass):
         """Calculate expected 2F, with uncertainty, over subsets of the observation span.
 
         This yields the expected behaviour that calculate_twoF_cumulative() can
-        be compared again: 2F for CW signals increases with duration
+        be compared against: 2F for CW signals increases with duration
         as we take longer and longer subsets of the total observation span.
 
         Parameters
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -970,10 +970,10 @@ class ComputeFstat(BaseSearchClass):
             Parameters at which to compute the cumulative twoF
         asini, period, ecc, tp, argp: float, optional
             Optional: Binary parameters at which to compute the cumulative 2F
-        tstart, tend: int
-            GPS times to restrict the range of data used - automatically
-            truncated to the span of data available
-            FIXME: tend currently unused
+        tstart, tend: int or None
+            GPS times to restrict the range of data used;
+            if None: falls back to self.minStartTime and self.maxStartTime;
+            if outside those: auto-truncated
         cumulative_fstat_segments: int
             Number of segments to split [tstart,tend] into
 
@@ -984,18 +984,11 @@ class ComputeFstat(BaseSearchClass):
         twoFs : ndarray of shape (cumulative_fstat_segments,)
             Values of twoF computed over [[tstart,tstart+tau] for tau in taus].
 
-        Notes
-        -----
-        The minimum cumulative twoF is hard-coded to be computed over
-        the first 6 hours from either the first timestamp in the data (if
-        tstart is smaller than it) or tstart.
-
         """
-        SFTminStartTime = self.SFT_timestamps[0]
-        SFTmaxStartTime = self.SFT_timestamps[-1]
-        tstart = np.max([SFTminStartTime, tstart])
-        min_tau = np.max([SFTminStartTime - tstart, 0]) + 3600 * 6
-        max_tau = SFTmaxStartTime - tstart
+        tstart = max(tstart, self.minStartTime) if tstart else self.minStartTime
+        tend = min(tend, self.maxStartTime) if tend else self.maxStartTime
+        min_tau = 2 * self.Tsft
+        max_tau = tend - tstart
         taus = np.linspace(min_tau, max_tau, cumulative_fstat_segments)
         twoFs = []
 
@@ -1130,10 +1123,9 @@ class ComputeFstat(BaseSearchClass):
         taus, twoFs = self.calculate_twoF_cumulative(**calculate_twoF_cumulative_kwargs)
         taus_days = taus / 86400.0
 
+        tstart = calculate_twoF_cumulative_kwargs.get("tstart", self.minStartTime)
         axis_kwargs = {
-            "xlabel": r"Days from $t_{{\rm start}}={:.0f}$".format(
-                calculate_twoF_cumulative_kwargs["tstart"]
-            ),
+            "xlabel": r"Days from $t_{{\rm start}}={:.0f}$".format(tstart),
             "ylabel": r"$\log_{10}(\mathrm{BSGL})_{\rm cumulative}$"
             if self.BSGL
             else r"$\widetilde{2\mathcal{F}}_{\rm cumulative}$",
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -955,7 +955,7 @@ class ComputeFstat(BaseSearchClass):
         argp=None,
         tstart=None,
         tend=None,
-        cumulative_fstat_segments=1000,
+        num_segments=1000,
     ):
         """Calculate the cumulative twoF over subsets of the observation span.
 
@@ -974,14 +974,14 @@ class ComputeFstat(BaseSearchClass):
             GPS times to restrict the range of data used;
             if None: falls back to self.minStartTime and self.maxStartTime;
             if outside those: auto-truncated
-        cumulative_fstat_segments: int
+        num_segments: int
             Number of segments to split [tstart,tend] into
 
         Returns
         -------
-        taus : ndarray of shape (cumulative_fstat_segments,)
+        taus : ndarray of shape (num_segments,)
             Offsets of each segment's tend from the overall tstart.
-        twoFs : ndarray of shape (cumulative_fstat_segments,)
+        twoFs : ndarray of shape (num_segments,)
             Values of twoF computed over [[tstart,tstart+tau] for tau in taus].
 
         """
@@ -989,7 +989,7 @@ class ComputeFstat(BaseSearchClass):
         tend = min(tend, self.maxStartTime) if tend else self.maxStartTime
         min_tau = 2 * self.Tsft
         max_tau = tend - tstart
-        taus = np.linspace(min_tau, max_tau, cumulative_fstat_segments)
+        taus = np.linspace(min_tau, max_tau, num_segments)
         twoFs = []
 
         if not self.transientWindowType:
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1045,7 +1045,7 @@ class ComputeFstat(BaseSearchClass):
         out = [
             helper_functions.predict_fstat(
                 minStartTime=self.minStartTime,
-                maxStartTime=t,
+                duration=t - self.minStartTime,
                 sftfilepattern=self.sftfilepattern,
                 IFO=IFO,
                 **pfs_input,
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1040,7 +1040,7 @@ class ComputeFstat(BaseSearchClass):
             for duration in cumulative_durations
         ]
 
-        return cumulative_durations, np.array(twoFs)
+        return tstart, cumulative_durations, np.array(twoFs)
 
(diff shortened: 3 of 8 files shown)


Rodrigo-Tenorio:
@dbkeitel I'll need your opinion on 174a402

Rodrigo-Tenorio:
For the record, the corresponding `cumultive_twoF` for the maximum candidate for a glitch search produces something, but I'm unable to tell whether that's correct or not, and the options are kind of different....?
standard_directed_MCMC_search_on_1_glitch_twoFcumulative

dbkeitel:
> The fact that a cumulative quantity gets lower with time bothers me, is this some quirk of dealing with glitches?

The tilde symbol might imply this is an _average_ 2F over segments? If so, and there's a glitch at ~27 days, and it's the non-glitch-aware search, then it would indeed start adding only random values (~4) per segment after that, instead of the high expected values, while still increasing the denominator, hence dropping.

Rodrigo-Tenorio:
I got rid of `pfs` switch and `injectSources` as they are kind of redundant after the latest change: If I give you `injection_sources`, you have to be smart enough to know that I want them in the plot.

pyfstat/core.py
```diff
@@ -985,10 +985,12 @@ def calculate_twoF_cumulative(
         max_tau = SFTmaxStartTime - tstart
         taus = np.linspace(min_tau, max_tau, npoints)
         twoFs = []
+
         if not self.transientWindowType:
             # still call the transient-Fstat-map function, but using the full range
```
dbkeitel, line 990: comment no longer true, now that you're fixing it

pyfstat/core.py
```diff
@@ -1017,12 +1019,12 @@ def _calculate_predict_fstat_cumulative(
         ----------
         N : int
             Number of timesteps to use between minStartTime and maxStartTime.
+        pfs_input : dict, optional
+            Input kwargs to predict_fstat (alternative to giving label and
+            outdir. Takes preference over those).
```
dbkeitel, line 1024: You mean that if this dict contains labels `"label"` or `"outdir"`, they will override those other options? Currently it sounds like the whole dict is just an alternative to those.

pyfstat/core.py
```diff
@@ -1076,10 +1077,11 @@ def plot_twoF_cumulative(
...
         add_pfs : bool
             If true, plot the predicted 2F and standard deviation
+        injection_parameters: dict
+            Dictionary with injection parameters to predict Fstat
+            (h0, cosi, psi, Alpha, Delta, F0)
         N : int
-            Number of points to use
-        injectSources : dict
-            See `ComputeFstat`
+            Number of points to use for PredictFStat.
```
dbkeitel, line 1084: If we're being backwards-incompatible with arguments anyway, we should call this `N_PFS` or something and also introduce here an explicit `N_CFS` that gets passed to `calculate_twoF_cumulative()`.

examples/other_examples/PyFstat_example_twoF_cumulative.py
```diff
@@ -1,85 +1,74 @@
...
+    maxCoverFreq=-0.5,
+)
 
-theta_prior = {
-    "F0": {"type": "unif", "lower": F0 - DeltaF0 / 2.0, "upper": F0 + DeltaF0 / 2.0},
-    "F1": {"type": "unif", "lower": F1 - DeltaF1 / 2.0, "upper": F1 + DeltaF1 / 2.0},
-    "F2": F2,
-    "Alpha": Alpha,
-    "Delta": Delta,
+cumulative_f_stat = {
```
dbkeitel, line 51:
```suggestion
cumulative_f_stat_params = {
```

examples/other_examples/PyFstat_example_twoF_cumulative.py
```diff
@@ -1,85 +1,74 @@
...
 }
+cumulative_f_stat.update(
+    {"tstart": gw_data["tstart"], "tend": gw_data["tstart"] + gw_data["duration"]}
+)
 
-ntemps = 1
-log10beta_min = -1
-nwalkers = 100
-nsteps = [50, 50]
+predict_f_stat = {
```
dbkeitel, line 61:
```suggestion
predict_f_stat_params = {
```

examples/other_examples/PyFstat_example_twoF_cumulative.py
```diff
@@ -1,85 +1,74 @@
...
+    key: signal_parameters.get(key, 0)
+    for key in ["F0", "Alpha", "Delta", "psi", "cosi", "h0"]
+}
 
-mcmc = pyfstat.MCMCSearch(
+time, twoF = compute_f_stat.calculate_twoF_cumulative(**cumulative_f_stat)
```
dbkeitel, line 66:
```suggestion
taus, twoF = compute_f_stat.calculate_twoF_cumulative(**cumulative_f_stat)
```

dbkeitel (changes requested):
Sounds promising so far, thanks, but I guess from your last example on #170 the PFS isn't quite right yet? Probably has to do with my recent changes to how that code is called, let me know if you're stuck and I should take a closer look.

Also I approve of making the basic `PyFstat_example_twoF_cumulative.py` use plain `ComputeFstat` instead of `MCMCSearch`.

dbkeitel:
> but I guess from your last example on #170 the PFS isn't quite right yet? Probably has to do with my recent changes to how that code is called, let me know if you're stuck and I should take a closer look.

Again, side-effect of another change I did recently on this code - I knew I would start breaking untested things... fixed in bdc25cdcb06f5cf14775552a1087ba9b5c2c58ae

PyFstat_example_twoF_cumulative_twoFcumulative


The problem is that `helper_functions.predict_fstat()` takes `kwargs`, but doesn't use or check them, so it was silently accepting the outdated `maxStartTime` and then not doing anything with it.

dbkeitel:
I just realized the remaining reason the plot was puzzling me: H1, L1 and H1L1 results look identical. That's because again PFS is not actually making use of the argument we're passing it, and after #188 this would also simply fail. Because `IFO` is also not a valid PFS option. (It was only a SFT format v1 backwards compatibility thing in the past.) It has `--IFOs` but as our own docs say:
>     IFOs : str or None
>        Comma-separated list of detectors.
>        Ignored if sftfilepattern is given,
>        required if it is not.

so what we actually need to do here, where we want to do separate multi- and single-IFO calls _with sftfilepattern_, is to downselect that pattern. Which in general isn't possible since the pattern wouldn't always be a `;`-separated list of per-IFO subpatterns, it could also perfectly legally be just `*.sft`...

Possible solutions:
1. implement proper multi-IFO constraints handling on the PFS side (but I don't know how deep one would need to dig for that, so let's not go that way)
1. implement logic here to only pass the flag if wanting to do single-IFO (should be easy)
1. drop single-IFO support (if 2. turns out to be more difficult than it sounds)

pyfstat/core.py
```diff
@@ -1099,27 +1101,19 @@ def plot_twoF_cumulative(
...
+        taus, twoFs = self.calculate_twoF_cumulative(**calculate_twoF_cumulative_kwargs)
         ax.plot(taus / 86400.0, twoFs, label=plt_label, color=c)
         if len(self.detector_names) > 1:
             detector_names = self.detector_names
             detectors = self.detectors
             for d in self.detector_names:
                 self.detectors = d
                 self.init_computefstatistic()
-                taus, twoFs = self.calculate_twoF_cumulative(**kwargs)
+                taus, twoFs = self.calculate_twoF_cumulative(
```
dbkeitel, line 1114:
```suggestion
                taus, twoFXs = self.calculate_twoF_cumulative(
```

Otherwise, at the end it returns these per-IFO values instead of the arguably expected multi-IFO ones.

pyfstat/core.py
```diff
@@ -1099,27 +1101,19 @@ def plot_twoF_cumulative(
...
 
-        taus, twoFs = self.calculate_twoF_cumulative(**kwargs)
+        taus, twoFs = self.calculate_twoF_cumulative(**calculate_twoF_cumulative_kwargs)
         ax.plot(taus / 86400.0, twoFs, label=plt_label, color=c)
         if len(self.detector_names) > 1:
             detector_names = self.detector_names
             detectors = self.detectors
             for d in self.detector_names:
                 self.detectors = d
                 self.init_computefstatistic()
```
dbkeitel, line 1113: This doesn't do anything, because `init_computefstatistic()` calls `_get_SFTCatalog` which itself returns immediately without change `if hasattr(self, "SFTCatalog")`. Not sure if

1. that is sane behaviour normally and we should only give those two options an optional override `enforceNewCatalog=False`
1. or we should just remove that early return and assume "whenever somebody calls a re-init, they want to start from scratch"

Probably requires a deeper look at other use cases of "re-initing".

dbkeitel: If this is difficult to resolve/decide, fallback from above: "drop single-IFO support" (for `plot_twoF_cumulative()`, that is).

pyfstat/core.py
```diff
@@ -1099,27 +1101,19 @@ def plot_twoF_cumulative(
...
-                taus, twoFs = self.calculate_twoF_cumulative(**kwargs)
+                taus, twoFs = self.calculate_twoF_cumulative(
+                    **calculate_twoF_cumulative_kwargs
+                )
                 ax.plot(
                     taus / 86400.0,
                     twoFs,
```
dbkeitel, line 1119:
```suggestion
                taus, twoFXs = self.calculate_twoF_cumulative(
                    **calculate_twoF_cumulative_kwargs
                )
                ax.plot(
                    taus / 86400.0,
                    twoFXs,
```
Otherwise, at the end it returns these per-IFO values instead of the arguably expected multi-IFO ones.

pyfstat/core.py
Rodrigo-Tenorio: Agree, but this is a Python code: `predict_fstat_segments` sounds clearer ?

pyfstat/core.py
Rodrigo-Tenorio: I'll try to do a single-IFO from within this method; if I don't succeed, I'll just drop it and tell the user "Look, don't give me two IFOs if you only need one: Use constraints".

Rodrigo-Tenorio: The amount of overhaul required to implement a multi-IFO is huge, and doing so would not be any different from just doing a different ComputeFstat set up (one for each of the IFO constraints).

Rodrigo-Tenorio:
Examples using the latest version:
PyFstat_example_twoF_cumulative_L1_twoFcumulative
PyFstat_example_twoF_cumulative_H1_twoFcumulative
PyFstat_example_twoF_cumulative_H1L1_twoFcumulative

Clearly there's a problem with `predict_fstat` using single IFO, even though `ComputeFStatistic` only uses a subset of the SFTs by construction.

Rodrigo-Tenorio:
Now I get what @dbkeitel  was saying:
https://github.com/PyFstat/PyFstat/blob/051b239f040b6d1d27ed1a0a909a48c861b0fd60/pyfstat/core.py#L1048-L1055

That `sftfilepattern` is the one given as an input to `ComputeFstat`, which means it's completely unaffected by sft constraints (as these apply only when reading the actual data!).

Rodrigo-Tenorio:
Fair enough for me
PyFstat_example_twoF_cumulative_H1_twoFcumulative
PyFstat_example_twoF_cumulative_L1_twoFcumulative

pyfstat/core.py
```diff
@@ -1076,108 +1075,89 @@ def plot_twoF_cumulative(
...
-        if ax is None:
-            fig, ax = plt.subplots()
-        if injectSources:
+
+        taus, twoFs = self.calculate_twoF_cumulative(**calculate_twoF_cumulative_kwargs)
+        taus_days = taus / 86400.0
+
+        axis_kwargs = {
+            "xlabel": r"Days from $t_{{\rm start}}={:.0f}$".format(
+                calculate_twoF_cumulative_kwargs["tstart"]
```
dbkeitel, line 1101: If they're optional kwargs, for safety this should be something like
```suggestion
                calculate_twoF_cumulative_kwargs.get("tstart",0)
```
though I'm not sure if the previous `self.calculate_twoF_cumulative` won't have failed anyway if that argument is missing...?

tests.py
```diff
@@ -748,6 +748,44 @@ def test_get_fully_coherent_BSGL(self):
...
+        taus, twoF_cumulative = search.calculate_twoF_cumulative(
+            self.Writer.F0,
+            self.Writer.F1,
+            self.Writer.F2,
+            self.Writer.Alpha,
+            self.Writer.Delta,
+            tstart=self.Writer.tstart,
+            tend=self.Writer.tend(),
+        )
+        twoF_cumulative = twoF_cumulative[-1]
```
dbkeitel, line 767: Should we also test that the `[0]` element is close to 0 and the middle element close to `twoF/2`? Otherwise we're not testing against another regression of the "always takes full data stretch" type. Might require too much tolerancde fiddling, though.

pyfstat/core.py
```diff
@@ -1009,20 +1010,22 @@ def calculate_twoF_cumulative(
         return taus, np.array(twoFs)
 
     def _calculate_predict_fstat_cumulative(
-        self, N, label=None, outdir=None, IFO=None, pfs_input=None
+        self, N, label=None, outdir=None, pfs_input=None
     ):
         """Calculates the predicted 2F and standard deviation cumulatively
```
dbkeitel, line 1015: This explanation (and for the other related functions) should be extended, since it wasn't enough for us to understand what they do without detailed code vivisection. ;) But I can take a turn on that.

dbkeitel:
So if I read correctly, the main function now just uses whatever data it's given, and doesn't try to do single/multi-IFO comparison at all? And the latest plots above are from the example, where it loops over `ifo_constraints = ["L1", "H1", None]`, right? If so, sounds good, and end users can just do something similar if they want single-IFO output.

Rodrigo-Tenorio:
>So if I read correctly, the main function now just uses whatever data it's given, and doesn't try to do single/multi-IFO comparison at all? And the latest plots above are from the example, where it loops over ifo_constraints = ["L1", "H1", None], right? If so, sounds good, and end users can just do something similar if they want single-IFO output.

That's it, yes.

pyfstat/core.py
Rodrigo-Tenorio:
```
>>> np.max([1, None])
  File "<__array_function__ internals>", line 5, in amax
  File "/home/rodrigo/gravity/venv/pyfstat/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 2705, in amax
    return _wrapreduction(a, np.maximum, 'max', axis, None, out,
  File "/home/rodrigo/gravity/venv/pyfstat/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 87, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
TypeError: '>=' not supported between instances of 'int' and 'NoneType'
```
I usually don't like this type of coupling (it's lurking bug, it's going to fail whenever `calculate_twoF_cumulative` changes a bit), so we should do something sane about it (I can't conclude whether your proposal is the one that makes sense or not right now).

tests.py
Rodrigo-Tenorio: Maybe we can compare ranges on top of it?

Like `twoF_cumulative[-1] - twoF_cumulative[0]` and compare that to `twoF` on the whole stream minus `twoF` on the first two ours?

examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
```diff
@@ -90,3 +90,8 @@
 mcmc.plot_corner(add_prior=True, truths=signal_parameters)
 mcmc.plot_prior_posterior(injection_parameters=signal_parameters)
 mcmc.plot_chainconsumer(truth=signal_parameters)
+mcmc.plot_cumulative_max(
+    plot_label="MCMC Example",
+    # title="Cumulative 2F for the best MCMC candidate",
+    custom_axis_kwargs={"title": "Cumulative 2F for the best MCMC candidate"},
```
dbkeitel, line 96: As demonstrated by this previously broken example, from the user perspective it makes no sense to have a direct kwarg for the `plot_label` but have to put the overall title into an esoteric `custom_axis_kwargs`...

pyfstat/core.py
```diff
@@ -955,26 +955,39 @@ def calculate_twoF_cumulative(
...
             Parameters at which to compute the cumulative twoF
         asini, period, ecc, tp, argp: float, optional
-            Binary parameters at which to compute the cumulative 2F
+            Optional: Binary parameters at which to compute the cumulative 2F
         tstart, tend: int
             GPS times to restrict the range of data used - automatically
             truncated to the span of data available
-        npoints: int
-            Number of points to compute twoF along the span
+            FIXME: tend currently unused
```
dbkeitel, line 976: !!!

pyfstat/mcmc_based_searches.py
```diff
@@ -1233,6 +1233,10 @@ def plot_cumulative_max(self, **kwargs):
 
         if hasattr(self, "search") is False:
             self._initiate_search_object()
+        # Predicted 2F values will be calculated for the parameter space point
+        # in the ".loudest" file generated above,
+        # while the parameters that are explicitly passed here are for the
+        # cumulative 2F calculation on the actual data.
```
dbkeitel, line 1239: Problem: the core `plot_twoF_cumulative()` function does not actually currently work as I've documented it here and as this MCMC method seems to assume. Since we're not giving it "signal_parameters", it will just skip the whole PFS block. However, the lowest-level `_calculate_predict_fstat_cumulative()` itself is willing to work with a ".loudest" file instead, so I think just the `if` block in `plot_twoF_cumulative()` needs to be disassembled, and then this method here will get what it wants.

pyfstat/core.py
```diff
@@ -1059,128 +1084,118 @@ def plot_twoF_cumulative(
...
+            f" ({(taus_days[1] - taus_days[0]):.2g} days per segment)"
+        )
+
+        fig, ax = plt.subplots()
+        ax.grid()
+        ax.set(**axis_kwargs)
+
+        ax.plot(taus_days, twoFs, label=plot_label, color="k")
+
+        if signal_parameters:
```
dbkeitel, line 1162: This is what needs to be broken up to make `MCMCSearch.plot_cumulative_max()` happy, I think.

Note I've renamed the argument from `inejction_parameters` to `signal_parameters` because it can just as well be the loudest from a search on real data.

dbkeitel:
This is on a good track, but trying to document it better and run the 3 currently enabled examples exposed a few more issues, see inline comments.

Also @Rodrigo-Tenorio do you have any idea why / opinion on why it is `_calculate_predict_fstat_cumulative()` with underscore and `calculate_twoF_cumulative()` without?

examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
Rodrigo-Tenorio: That's right: I just followed the inertia of the code. I'll include it inside `custom_axis_kwargs` and rename it to `custom_axes_kwargs`.

Rodrigo-Tenorio: Actually, `custom_ax_kwargs` makes more sense. Done in db86df05aea35c6a769c708683f0b2ba4a5cb565.

pyfstat/core.py
Rodrigo-Tenorio: I made #189 to solve this.

Rodrigo-Tenorio:
> This is on a good track, but trying to document it better and run the 3 currently enabled examples exposed a few more issues, see inline comments.
> 
> Also @Rodrigo-Tenorio do you have any idea why it is / opinion on whether it should be `_calculate_predict_fstat_cumulative()` with underscore and `calculate_twoF_cumulative()` without?

Looking into the input format seems like `_calculate_predict_fstat_cumulative` is plumbing, while `calculate_twoF_cumulative` is porcelain...? 

I don't understand why someone would program these two methods using such a different input structure. May be because one needs to call a process and the other does not?

Rodrigo-Tenorio:
[![](https://mermaid.ink/img/eyJjb2RlIjoiZ3JhcGggTFJcbnN1YmdyYXBoIENvbXB1dGVGU3RhdGlzdGljXG5BKGNhbGN1bGF0ZV90d29GX2N1bXVsYXRpdmUpXG5CKF9jYWxjdWxhdGVfcHJlZGljdF9mc3RhdF9jdW11bGF0aXZlKVxuQyhwbG90X3R3b0ZfY3VtdWxhdGl2ZSkgLS0-IHxjYWxjdWxhdGVfdHdvRl9jdW11bGF0aXZlX2t3YXJncz1zdXBlci5nZXRfbWF4X3R3b0Z8QVxuQyhwbG90X3R3b0ZfY3VtdWxhdGl2ZSkgLS0-IHxzaWduYWxfcGFyYW1ldGVycz0qKmt3YXJnc3xCXG5lbmRcblxuc3ViZ3JhcGggTUNNQ1NlYXJjaFxuRChwbG90X2N1bXVsYXRpdmVfbWF4KSAtLT4gIHxzZWxmLmdldF9tYXhfdHdvRnwgQ1xuZW5kXG5cbnN1YmdyYXBoIE1DTUNHbGl0Y2hTZWFyY2hcbkUocGxvdF9jdW11bGF0aXZlX21heCkgLS0-IHxzZWxmLmdldF9tYXhfdHdvRnwgQVxuZW5kIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQiLCJ0aGVtZVZhcmlhYmxlcyI6eyJiYWNrZ3JvdW5kIjoid2hpdGUiLCJwcmltYXJ5Q29sb3IiOiIjRUNFQ0ZGIiwic2Vjb25kYXJ5Q29sb3IiOiIjZmZmZmRlIiwidGVydGlhcnlDb2xvciI6ImhzbCg4MCwgMTAwJSwgOTYuMjc0NTA5ODAzOSUpIiwicHJpbWFyeUJvcmRlckNvbG9yIjoiaHNsKDI0MCwgNjAlLCA4Ni4yNzQ1MDk4MDM5JSkiLCJzZWNvbmRhcnlCb3JkZXJDb2xvciI6ImhzbCg2MCwgNjAlLCA4My41Mjk0MTE3NjQ3JSkiLCJ0ZXJ0aWFyeUJvcmRlckNvbG9yIjoiaHNsKDgwLCA2MCUsIDg2LjI3NDUwOTgwMzklKSIsInByaW1hcnlUZXh0Q29sb3IiOiIjMTMxMzAwIiwic2Vjb25kYXJ5VGV4dENvbG9yIjoiIzAwMDAyMSIsInRlcnRpYXJ5VGV4dENvbG9yIjoicmdiKDkuNTAwMDAwMDAwMSwgOS41MDAwMDAwMDAxLCA5LjUwMDAwMDAwMDEpIiwibGluZUNvbG9yIjoiIzMzMzMzMyIsInRleHRDb2xvciI6IiMzMzMiLCJtYWluQmtnIjoiI0VDRUNGRiIsInNlY29uZEJrZyI6IiNmZmZmZGUiLCJib3JkZXIxIjoiIzkzNzBEQiIsImJvcmRlcjIiOiIjYWFhYTMzIiwiYXJyb3doZWFkQ29sb3IiOiIjMzMzMzMzIiwiZm9udEZhbWlseSI6IlwidHJlYnVjaGV0IG1zXCIsIHZlcmRhbmEsIGFyaWFsIiwiZm9udFNpemUiOiIxNnB4IiwibGFiZWxCYWNrZ3JvdW5kIjoiI2U4ZThlOCIsIm5vZGVCa2ciOiIjRUNFQ0ZGIiwibm9kZUJvcmRlciI6IiM5MzcwREIiLCJjbHVzdGVyQmtnIjoiI2ZmZmZkZSIsImNsdXN0ZXJCb3JkZXIiOiIjYWFhYTMzIiwiZGVmYXVsdExpbmtDb2xvciI6IiMzMzMzMzMiLCJ0aXRsZUNvbG9yIjoiIzMzMyIsImVkZ2VMYWJlbEJhY2tncm91bmQiOiIjZThlOGU4IiwiYWN0b3JCb3JkZXIiOiJoc2woMjU5LjYyNjE2ODIyNDMsIDU5Ljc3NjUzNjMxMjglLCA4Ny45MDE5NjA3ODQzJSkiLCJhY3RvckJrZyI6IiNFQ0VDRkYiLCJhY3RvclRleHRDb2xvciI6ImJsYWNrIiwiYWN0b3JMaW5lQ29sb3IiOiJncmV5Iiwic2lnbmFsQ29sb3IiOiIjMzMzIiwic2lnbmFsVGV4dENvbG9yIjoiIzMzMyIsImxhYmVsQm94QmtnQ29sb3IiOiIjRUNFQ0ZGIiwibGFiZWxCb3hCb3JkZXJDb2xvciI6ImhzbCgyNTkuNjI2MTY4MjI0MywgNTkuNzc2NTM2MzEyOCUsIDg3LjkwMTk2MDc4NDMlKSIsImxhYmVsVGV4dENvbG9yIjoiYmxhY2siLCJsb29wVGV4dENvbG9yIjoiYmxhY2siLCJub3RlQm9yZGVyQ29sb3IiOiIjYWFhYTMzIiwibm90ZUJrZ0NvbG9yIjoiI2ZmZjVhZCIsIm5vdGVUZXh0Q29sb3IiOiJibGFjayIsImFjdGl2YXRpb25Cb3JkZXJDb2xvciI6IiM2NjYiLCJhY3RpdmF0aW9uQmtnQ29sb3IiOiIjZjRmNGY0Iiwic2VxdWVuY2VOdW1iZXJDb2xvciI6IndoaXRlIiwic2VjdGlvbkJrZ0NvbG9yIjoicmdiYSgxMDIsIDEwMiwgMjU1LCAwLjQ5KSIsImFsdFNlY3Rpb25Ca2dDb2xvciI6IndoaXRlIiwic2VjdGlvbkJrZ0NvbG9yMiI6IiNmZmY0MDAiLCJ0YXNrQm9yZGVyQ29sb3IiOiIjNTM0ZmJjIiwidGFza0JrZ0NvbG9yIjoiIzhhOTBkZCIsInRhc2tUZXh0TGlnaHRDb2xvciI6IndoaXRlIiwidGFza1RleHRDb2xvciI6IndoaXRlIiwidGFza1RleHREYXJrQ29sb3IiOiJibGFjayIsInRhc2tUZXh0T3V0c2lkZUNvbG9yIjoiYmxhY2siLCJ0YXNrVGV4dENsaWNrYWJsZUNvbG9yIjoiIzAwMzE2MyIsImFjdGl2ZVRhc2tCb3JkZXJDb2xvciI6IiM1MzRmYmMiLCJhY3RpdmVUYXNrQmtnQ29sb3IiOiIjYmZjN2ZmIiwiZ3JpZENvbG9yIjoibGlnaHRncmV5IiwiZG9uZVRhc2tCa2dDb2xvciI6ImxpZ2h0Z3JleSIsImRvbmVUYXNrQm9yZGVyQ29sb3IiOiJncmV5IiwiY3JpdEJvcmRlckNvbG9yIjoiI2ZmODg4OCIsImNyaXRCa2dDb2xvciI6InJlZCIsInRvZGF5TGluZUNvbG9yIjoicmVkIiwibGFiZWxDb2xvciI6ImJsYWNrIiwiZXJyb3JCa2dDb2xvciI6IiM1NTIyMjIiLCJlcnJvclRleHRDb2xvciI6IiM1NTIyMjIiLCJjbGFzc1RleHQiOiIjMTMxMzAwIiwiZmlsbFR5cGUwIjoiI0VDRUNGRiIsImZpbGxUeXBlMSI6IiNmZmZmZGUiLCJmaWxsVHlwZTIiOiJoc2woMzA0LCAxMDAlLCA5Ni4yNzQ1MDk4MDM5JSkiLCJmaWxsVHlwZTMiOiJoc2woMTI0LCAxMDAlLCA5My41Mjk0MTE3NjQ3JSkiLCJmaWxsVHlwZTQiOiJoc2woMTc2LCAxMDAlLCA5Ni4yNzQ1MDk4MDM5JSkiLCJmaWxsVHlwZTUiOiJoc2woLTQsIDEwMCUsIDkzLjUyOTQxMTc2NDclKSIsImZpbGxUeXBlNiI6ImhzbCg4LCAxMDAlLCA5Ni4yNzQ1MDk4MDM5JSkiLCJmaWxsVHlwZTciOiJoc2woMTg4LCAxMDAlLCA5My41Mjk0MTE3NjQ3JSkifX0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggTFJcbnN1YmdyYXBoIENvbXB1dGVGU3RhdGlzdGljXG5BKGNhbGN1bGF0ZV90d29GX2N1bXVsYXRpdmUpXG5CKF9jYWxjdWxhdGVfcHJlZGljdF9mc3RhdF9jdW11bGF0aXZlKVxuQyhwbG90X3R3b0ZfY3VtdWxhdGl2ZSkgLS0-IHxjYWxjdWxhdGVfdHdvRl9jdW11bGF0aXZlX2t3YXJncz1zdXBlci5nZXRfbWF4X3R3b0Z8QVxuQyhwbG90X3R3b0ZfY3VtdWxhdGl2ZSkgLS0-IHxzaWduYWxfcGFyYW1ldGVycz0qKmt3YXJnc3xCXG5lbmRcblxuc3ViZ3JhcGggTUNNQ1NlYXJjaFxuRChwbG90X2N1bXVsYXRpdmVfbWF4KSAtLT4gIHxzZWxmLmdldF9tYXhfdHdvRnwgQ1xuZW5kXG5cbnN1YmdyYXBoIE1DTUNHbGl0Y2hTZWFyY2hcbkUocGxvdF9jdW11bGF0aXZlX21heCkgLS0-IHxzZWxmLmdldF9tYXhfdHdvRnwgQVxuZW5kIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQiLCJ0aGVtZVZhcmlhYmxlcyI6eyJiYWNrZ3JvdW5kIjoid2hpdGUiLCJwcmltYXJ5Q29sb3IiOiIjRUNFQ0ZGIiwic2Vjb25kYXJ5Q29sb3IiOiIjZmZmZmRlIiwidGVydGlhcnlDb2xvciI6ImhzbCg4MCwgMTAwJSwgOTYuMjc0NTA5ODAzOSUpIiwicHJpbWFyeUJvcmRlckNvbG9yIjoiaHNsKDI0MCwgNjAlLCA4Ni4yNzQ1MDk4MDM5JSkiLCJzZWNvbmRhcnlCb3JkZXJDb2xvciI6ImhzbCg2MCwgNjAlLCA4My41Mjk0MTE3NjQ3JSkiLCJ0ZXJ0aWFyeUJvcmRlckNvbG9yIjoiaHNsKDgwLCA2MCUsIDg2LjI3NDUwOTgwMzklKSIsInByaW1hcnlUZXh0Q29sb3IiOiIjMTMxMzAwIiwic2Vjb25kYXJ5VGV4dENvbG9yIjoiIzAwMDAyMSIsInRlcnRpYXJ5VGV4dENvbG9yIjoicmdiKDkuNTAwMDAwMDAwMSwgOS41MDAwMDAwMDAxLCA5LjUwMDAwMDAwMDEpIiwibGluZUNvbG9yIjoiIzMzMzMzMyIsInRleHRDb2xvciI6IiMzMzMiLCJtYWluQmtnIjoiI0VDRUNGRiIsInNlY29uZEJrZyI6IiNmZmZmZGUiLCJib3JkZXIxIjoiIzkzNzBEQiIsImJvcmRlcjIiOiIjYWFhYTMzIiwiYXJyb3doZWFkQ29sb3IiOiIjMzMzMzMzIiwiZm9udEZhbWlseSI6IlwidHJlYnVjaGV0IG1zXCIsIHZlcmRhbmEsIGFyaWFsIiwiZm9udFNpemUiOiIxNnB4IiwibGFiZWxCYWNrZ3JvdW5kIjoiI2U4ZThlOCIsIm5vZGVCa2ciOiIjRUNFQ0ZGIiwibm9kZUJvcmRlciI6IiM5MzcwREIiLCJjbHVzdGVyQmtnIjoiI2ZmZmZkZSIsImNsdXN0ZXJCb3JkZXIiOiIjYWFhYTMzIiwiZGVmYXVsdExpbmtDb2xvciI6IiMzMzMzMzMiLCJ0aXRsZUNvbG9yIjoiIzMzMyIsImVkZ2VMYWJlbEJhY2tncm91bmQiOiIjZThlOGU4IiwiYWN0b3JCb3JkZXIiOiJoc2woMjU5LjYyNjE2ODIyNDMsIDU5Ljc3NjUzNjMxMjglLCA4Ny45MDE5NjA3ODQzJSkiLCJhY3RvckJrZyI6IiNFQ0VDRkYiLCJhY3RvclRleHRDb2xvciI6ImJsYWNrIiwiYWN0b3JMaW5lQ29sb3IiOiJncmV5Iiwic2lnbmFsQ29sb3IiOiIjMzMzIiwic2lnbmFsVGV4dENvbG9yIjoiIzMzMyIsImxhYmVsQm94QmtnQ29sb3IiOiIjRUNFQ0ZGIiwibGFiZWxCb3hCb3JkZXJDb2xvciI6ImhzbCgyNTkuNjI2MTY4MjI0MywgNTkuNzc2NTM2MzEyOCUsIDg3LjkwMTk2MDc4NDMlKSIsImxhYmVsVGV4dENvbG9yIjoiYmxhY2siLCJsb29wVGV4dENvbG9yIjoiYmxhY2siLCJub3RlQm9yZGVyQ29sb3IiOiIjYWFhYTMzIiwibm90ZUJrZ0NvbG9yIjoiI2ZmZjVhZCIsIm5vdGVUZXh0Q29sb3IiOiJibGFjayIsImFjdGl2YXRpb25Cb3JkZXJDb2xvciI6IiM2NjYiLCJhY3RpdmF0aW9uQmtnQ29sb3IiOiIjZjRmNGY0Iiwic2VxdWVuY2VOdW1iZXJDb2xvciI6IndoaXRlIiwic2VjdGlvbkJrZ0NvbG9yIjoicmdiYSgxMDIsIDEwMiwgMjU1LCAwLjQ5KSIsImFsdFNlY3Rpb25Ca2dDb2xvciI6IndoaXRlIiwic2VjdGlvbkJrZ0NvbG9yMiI6IiNmZmY0MDAiLCJ0YXNrQm9yZGVyQ29sb3IiOiIjNTM0ZmJjIiwidGFza0JrZ0NvbG9yIjoiIzhhOTBkZCIsInRhc2tUZXh0TGlnaHRDb2xvciI6IndoaXRlIiwidGFza1RleHRDb2xvciI6IndoaXRlIiwidGFza1RleHREYXJrQ29sb3IiOiJibGFjayIsInRhc2tUZXh0T3V0c2lkZUNvbG9yIjoiYmxhY2siLCJ0YXNrVGV4dENsaWNrYWJsZUNvbG9yIjoiIzAwMzE2MyIsImFjdGl2ZVRhc2tCb3JkZXJDb2xvciI6IiM1MzRmYmMiLCJhY3RpdmVUYXNrQmtnQ29sb3IiOiIjYmZjN2ZmIiwiZ3JpZENvbG9yIjoibGlnaHRncmV5IiwiZG9uZVRhc2tCa2dDb2xvciI6ImxpZ2h0Z3JleSIsImRvbmVUYXNrQm9yZGVyQ29sb3IiOiJncmV5IiwiY3JpdEJvcmRlckNvbG9yIjoiI2ZmODg4OCIsImNyaXRCa2dDb2xvciI6InJlZCIsInRvZGF5TGluZUNvbG9yIjoicmVkIiwibGFiZWxDb2xvciI6ImJsYWNrIiwiZXJyb3JCa2dDb2xvciI6IiM1NTIyMjIiLCJlcnJvclRleHRDb2xvciI6IiM1NTIyMjIiLCJjbGFzc1RleHQiOiIjMTMxMzAwIiwiZmlsbFR5cGUwIjoiI0VDRUNGRiIsImZpbGxUeXBlMSI6IiNmZmZmZGUiLCJmaWxsVHlwZTIiOiJoc2woMzA0LCAxMDAlLCA5Ni4yNzQ1MDk4MDM5JSkiLCJmaWxsVHlwZTMiOiJoc2woMTI0LCAxMDAlLCA5My41Mjk0MTE3NjQ3JSkiLCJmaWxsVHlwZTQiOiJoc2woMTc2LCAxMDAlLCA5Ni4yNzQ1MDk4MDM5JSkiLCJmaWxsVHlwZTUiOiJoc2woLTQsIDEwMCUsIDkzLjUyOTQxMTc2NDclKSIsImZpbGxUeXBlNiI6ImhzbCg4LCAxMDAlLCA5Ni4yNzQ1MDk4MDM5JSkiLCJmaWxsVHlwZTciOiJoc2woMTg4LCAxMDAlLCA5My41Mjk0MTE3NjQ3JSkifX0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)

I present a (sacrilegious) chart showing the dependencies among the different methods so we are all aware of what's going on right now.

tests.py
```diff
@@ -37,6 +37,8 @@ def tearDownClass(self):
     "F2": 0,
     "h0": 5,
     "cosi": 0,
+    "psi": 0,
+    "phi0": 0,
```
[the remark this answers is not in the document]
Rodrigo-Tenorio: It's actually `phi` :wink:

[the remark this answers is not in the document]
dbkeitel: Argh. That's a completely needless case of lal->PyFstat divergence... Future wishlist item: let's stop randomly renaming things. ;)

Fixing this fixes all tests except for the new one and one other (see below). For the new one, I guess we need to call the usual translation dict somewhere.

Could you meanwhile take a look at:
```
tests.py::TestBinaryModulatedWriter::test_tp_parsing - ValueError: 'psi' is not in list
```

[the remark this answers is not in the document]
Rodrigo-Tenorio: I'm on it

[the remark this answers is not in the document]
Rodrigo-Tenorio: Looks like the test suite does not do what I think it did, since e.g. binary arguments are not included in `BinaryModulatedWriter` and so... I'll check that after fixing this.

dbkeitel:
1 sounds definitely good. For 2 I have to reread again to see what the point of the separate arg was. 3 sounds like a huge project which we shouldn't spend time on for now.

Rodrigo-Tenorio:
This looks Ok.
PyFstat_example_twoF_cumulative_H1L1_twoFcumulative
PyFstat_example_twoF_cumulative_H1_twoFcumulative
PyFstat_example_twoF_cumulative_L1_twoFcumulative

Rodrigo-Tenorio:
I'll run all examples to see if there's anything else broken [ ]

Rodrigo-Tenorio:
For the sake of consistency, I'll run again all the examples: https://github.com/PyFstat/PyFstat/actions/runs/276893354

pyfstat/core.py
```diff
@@ -941,6 +941,33 @@ def get_fullycoherent_twoF(
 
         return log10_BSGL / np.log10(np.exp(1))
 
+    def _set_up_cumulative_times(self, tstart, tend, num_segments):
+        """Construct time arrays to be used in cumulative twoF computations.
+
+        This allows calculate_twoF_cumulative and predict_twoF_cumulative to use
+        the same convention (although the number of segments on use is generally
+        different due to the computing time required by predict_twoF_cumulative).
```
dbkeitel, line 949: good idea!

pyfstat/core.py
```diff
@@ -955,44 +982,50 @@ def calculate_twoF_cumulative(
...
-        SFTmaxStartTime = self.SFT_timestamps[-1]
-        tstart = np.max([SFTminStartTime, tstart])
-        min_tau = np.max([SFTminStartTime - tstart, 0]) + 3600 * 6
-        max_tau = SFTmaxStartTime - tstart
-        taus = np.linspace(min_tau, max_tau, npoints)
-        twoFs = []
         if not self.transientWindowType:
-            # still call the transient-Fstat-map function, but using the full range
-            self.transientWindowType = "none"
+            # FIXME: Can we move this elsewhere?
```
dbkeitel, line 1017: why?

pyfstat/core.py
```diff
@@ -1004,186 +1037,189 @@ def calculate_twoF_cumulative(
...
+        tstart=None,
+        tend=None,
+        num_segments=10,
+        **predict_fstat_kwargs,
     ):
-        """Calculates the predicted 2F and standard deviation cumulatively
+        """Calculate expected 2F, with uncertainty, over subsets of the observation span.
+
+        This yields the expected behaviour that calculate_twoF_cumulative() can
+        be compared again: 2F for CW signals increases with duration
```
dbkeitel, line 1061:
```suggestion
        be compared against: 2F for CW signals increases with duration
```

pyfstat/core.py
```diff
@@ -1004,186 +1037,189 @@ def calculate_twoF_cumulative(
...
-        injectSources : dict
-            See `ComputeFstat`
-        ax : matplotlib.axes._subplots_AxesSubplot, optional
-            Axis to add the plot to.
-        c : str
-            Colour
+        CFS_input: dict
+            self.calculate_twoF_cumulative input arguments.
+        PFS_input: dict
+            self.predict_twoF_cumulative input arguments.
```
dbkeitel, line 1129:
```suggestion
        PFS_input: dict or None
            self.predict_twoF_cumulative input arguments
            (besides [tstart,tend,num_segments]).
            if None: do not calculate predicted 2F
```

pyfstat/core.py
```diff
@@ -1004,186 +1037,189 @@ def calculate_twoF_cumulative(
...
+            if outside those: auto-truncated
+        num_segments_(CFS|PFS) : int
+            Number of time segments to (compute|predict) twoF.
+        custom_ax_kwargs : dict
+            Optional axis formatting options.
         savefig : bool
-            If true, save the figure in outdir
-        title, plt_label: str
-            Figure title and label
+            If true, save the figure in outdir and return taus, twoFs.
```
dbkeitel, line 1139: Seems like you actually removed that return behaviour...?

pyfstat/core.py
```diff
@@ -1004,186 +1037,189 @@ def calculate_twoF_cumulative(
...
-        N : int
-            Number of points to use
-        injectSources : dict
-            See `ComputeFstat`
-        ax : matplotlib.axes._subplots_AxesSubplot, optional
-            Axis to add the plot to.
-        c : str
-            Colour
+        CFS_input: dict
+            self.calculate_twoF_cumulative input arguments.
```
dbkeitel, line 1127:
```suggestion
            self.calculate_twoF_cumulative input arguments
            (besides [tstart,tend,num_segments]).
```

pyfstat/core.py
```diff
@@ -1004,186 +1037,189 @@ def calculate_twoF_cumulative(
...
+        # Compute cumulative twoF
+        actual_tstart_CFS, taus_CFS, twoFs = self.calculate_twoF_cumulative(
+            tstart=tstart,
+            tend=tend,
+            num_segments=num_segments_CFS,
+            **CFS_input,
+        )
+        taus_CFS_days = taus_CFS / 86400.0
+
+        # Set up plot-realted objects
```
dbkeitel, line 1161:
```suggestion
        # Set up plot-related objects
```

pyfstat/core.py
```diff
@@ -1004,186 +1037,189 @@ def calculate_twoF_cumulative(
...
+        axis_kwargs = {
+            "xlabel": r"Days from $t_{{\rm start}}={:.0f}$".format(actual_tstart_CFS),
+            "ylabel": r"$\log_{10}(\mathrm{BSGL})_{\rm cumulative}$"
+            if self.BSGL
+            else r"$\widetilde{2\mathcal{F}}_{\rm cumulative}$",
```
dbkeitel, line 1166: while we're rewriting all these blocks anyway, let's make strings more consistent:
```suggestion
            "xlabel": f"Days from $t_\\mathrm{{start}}={actual_tstart_CFS:.0f}$",
            "ylabel": "$\\log_{10}(\\mathrm{BSGL})_\\mathrm{cumulative}$"
            if self.BSGL
            else "$\\widetilde{2\\mathcal{F}}_\\mathrm{cumulative}$",
```
(may need a round of debugging)

pyfstat/core.py
```diff
@@ -1004,186 +1037,189 @@ def calculate_twoF_cumulative(
...
             ax.fill_between(
-                (times - self.minStartTime) / 86400.0,
+                taus_PFS_days,
                 pfs - pfs_sigma,
                 pfs + pfs_sigma,
-                color=c,
+                color="cyan",
                 label=(
                     r"Predicted $\langle 2\mathcal{F} " r"\rangle\pm $ 1-$\sigma$ band"
                 ),
```
dbkeitel, line 1212:
```suggestion
                label="Predicted $\\langle 2\\mathcal{F} \\rangle \\pm 1\\sigma$ band",
```
or something like that?

pyfstat/mcmc_based_searches.py
```diff
@@ -1233,37 +1233,8 @@ def plot_cumulative_max(self, **kwargs):
 
         if hasattr(self, "search") is False:
             self._initiate_search_object()
-        if self.binary is False:
```
dbkeitel, line 1236: I can't put comments on unchanged lines, so I'll put it here: do we still need the `self.generate_loudest()` a few lines above? I don't think the core functions know how to ingest that file any more now,

pyfstat/mcmc_based_searches.py
```diff
@@ -1233,37 +1233,8 @@ def plot_cumulative_max(self, **kwargs):
...
-                period=d["period"],
-                ecc=d["ecc"],
-                argp=d["argp"],
-                tp=d["argp"],
-                tstart=self.minStartTime,
-                tend=self.maxStartTime,
-                **kwargs,
-            )
+
+        self.search.plot_twoF_cumulative(self.label, self.outdir, **d, **kwargs)
```
dbkeitel: I don't understand how this call works with the rewritten core function - shouldn't it expect a non-optional keyword argument `CFS_input` but instead you're passing it the **ed version of the dictionary, i.e. the tuple of signal parameters...? Why does that work? :confused: 

Also, are we now never allowing PFS from this call? No, wait, I guess the end user can pass `PFS_input` through the `kwargs` of this one, right?

dbkeitel (changes requested):
Looks good overall, but I've added minor cleanup comments and am confused about how the MCMC method calls the core function.

pyfstat/mcmc_based_searches.py
dbkeitel: ...hey, github, what's going on? This was shown to me as the up-to-date diff after I refreshed, Rodrigo has not pushed any new changes since 5h ago, but now it's "outdated". :confused: :confounded: 

In fact, the new version looks much more like what I expected!

pyfstat/mcmc_based_searches.py
dbkeitel: I think this one still applies under the latest diff, unless github is hiding more from me...

pyfstat/mcmc_based_searches.py
```diff
@@ -1233,37 +1233,10 @@ def plot_cumulative_max(self, **kwargs):
...
-                argp=d["argp"],
-                tp=d["argp"],
-                tstart=self.minStartTime,
-                tend=self.maxStartTime,
-                **kwargs,
-            )
+
+        self.search.plot_twoF_cumulative(
+            CFS_input=d, label=self.label, outdir=self.outdir, **kwargs
+        )
```
dbkeitel, line 1239: This call makes sense now.

Just for confirmation  @Rodrigo-Tenorio : end user is expected to pass `PFS_input` through `kwargs` if they e.g. want to compare the loudest candidate vs injected values?

If yes, since that should be a fairly popular use case, we should mention it explicitly in the documentation of this method.

examples/mcmc_examples/PyFstat_example_semi_coherent_MCMC_search.py
```diff
@@ -90,3 +90,8 @@
 mcmc.plot_corner(add_prior=True, truths=signal_parameters)
 mcmc.plot_prior_posterior(injection_parameters=signal_parameters)
 mcmc.plot_chainconsumer(truth=signal_parameters)
+mcmc.plot_cumulative_max(
+    plot_label="MCMC Example",
+    # title="Cumulative 2F for the best MCMC candidate",
```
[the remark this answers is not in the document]
dbkeitel: Seems already gone in latest diff. Thanks again, github... ;)

pyfstat/core.py
Rodrigo-Tenorio: This method is re-initializing the whole instance with a different window.
Should we assume a "Hey user, make sure not to re-use your instances: always create a new one from scratch" way of thinking? That's fine with me, but at least we should ad a logging or something acknowledging the window change + initialization.

pyfstat/core.py
Rodrigo-Tenorio: AFAIK, you can use `f` and raw strings at once.

I'll try this one locally.

pyfstat/core.py
Rodrigo-Tenorio: idem as before. Thanks.

pyfstat/core.py
dbkeitel: Yes, but the raw-string hack for latex is ugly and was leading to weird issues in some places (flake8 sometimes dislikes it too), so I started replacing it with regular strings and double escapes where possible.

pyfstat/core.py
dbkeitel: It's ugly and a bit dangerous, yes. I think we use similar hacks in several places, but that's no real excuse.

I think the easiest solution is to _reset it at the end of this function_ (store current `self.transientWindowType` in a temporary variable to do so).

pyfstat/core.py
Rodrigo-Tenorio: Done in 25385a56453853ae8283eddc693331407750c752

pyfstat/core.py
Rodrigo-Tenorio: Done in 25385a56453853ae8283eddc693331407750c752

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: >I think this one still applies under the latest diff, unless github is hiding more from me...

Before there was an ugly (I know, that's opinionated, but so is `black` and we are happy with it) `if: else` block, the only thing it made wast to pass binary parameters to the function (or not); however, that was not necessary as the function was able to process binary parameters as `None`. Just a clarification.

Rodrigo-Tenorio: > I can't put comments on unchanged lines, so I'll put it here: do we still need the self.generate_loudest() a few lines above? I don't think the core functions know how to ingest that file any more now,

I agree. Let's kill it and push forward #191 at some point so we are still able to do it but in a more Pythonic way.

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: Yes, every other argument (including `PFS_input` or even `savefig` has to be passed through kwargs.

dbkeitel: Ok. On second thought, let's not bother with amending the doc, it points clearly enough to the core function.

Rodrigo-Tenorio: I'd say something like  57833ee is explicit enough?

dbkeitel: Oh, you're fast! Perfect.

pyfstat/core.py
Rodrigo-Tenorio: That defeats the purpose... since the reason to do it here is to make sure one can use the transient trick to compute "cumulative 2F" as "Transient 2F with increasing lenght"?

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: Tbh, that doc was pointing to a `pyfstat.core.plot_twoF_cumulative`, which doesn't exist anymore as a function (does this mean PyFstat started as a bunch of functions and then it got objectified? anyways, that's for another time).

pyfstat/core.py
Rodrigo-Tenorio: cd4b13c will do the trick until we clean up this class structure (and by "clean up" I mean "tear down and re-build re-using as much code as possible"; but that's more of a dream than a reality rn).

dbkeitel (approved):

PR #195, opened 2020-09-28
Setup dev environment

Rodrigo-Tenorio:
Automates the set up of a development environment by just running a bash script.

The script:

- Makes sure one is developing inside a virtual environment.
- Installs `pytest`, `flake8`, `black` and `pre-commit`.
- Sets up pre-commit hooks to run `flake8` and `black`.

Optionally, we could add a script to authomatically download ephemeris (which will be only needed if installed from `pip`).

@dbkeitel Opinions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ and [citing](#citing-this-work) PyFstat.
 * We also have a number of
 [examples](https://github.com/PyFstat/PyFstat/tree/master/examples),
 demonstrating different use cases.
-* Developers or contributors are encouraged to have a look into
+* New contributors are encouraged to have a look into
 [how to set up a development environment](#contributing-to-pyfstat)
 
 
@@ -281,8 +281,8 @@ Here's what you need to know:
   `black --check --diff .` to show the required style changes, or `black .` to automatically apply them.
 * `bin/setup-dev-tools.sh` gets your virtual environment ready for you. After making sure you are 
 using a virtual environment, it install `pytest`, `black` and `flake8` and uses `pre-commit` to run
-the last two using a pre-commit hoook. In this way, you will be prompted by a warning whenver you 
-forget to run `black` or `flake8` befor commit :wink:.
+the last two using a pre-commit hook. In this way, you will be prompted a warning whenever you 
+forget to run `black` or `flake8` before doing your commit :wink:.
 
 ## Contributors
 
--- a/README.md
+++ b/README.md
@@ -243,8 +243,8 @@ echo 'export LALPULSAR_DATADIR=$VIRTUAL_ENV/share/lalpulsar' >> ${VIRTUAL_ENV}/b
 deactivate
 source path/to/venv/bin/activate
 ```
-An executable version of this snippet is readily accessible by **sourcing** `bin/get_and_export_ephemeris.sh`. 
-Mind that this script does **not** include an export command anywhere, so you will have to source it every time
+An executable version of this snippet is readily accessible by **sourcing** `bin/get-and-export-ephemeris.sh`. 
+Mind that this script does **not** include an `export` command anywhere, so you will have to source it every time
 in order to properly set `LALPULSAR_DATADIR` variable.
 
 If instead you have built and installed lalsuite from source,
@@ -284,8 +284,8 @@ Here's what you need to know:
   `black --check --diff .` to show the required style changes, or `black .` to automatically apply them.
 * `bin/setup-dev-tools.sh` gets your virtual environment ready for you. After making sure you are 
 using a virtual environment (venv or conda),
-it installs `pytest`, `black` and `flake8` and uses `pre-commit` to run
-the last two using a pre-commit hook. In this way, you will be prompted a warning whenever you 
+it installs `black`, `flake8`, `pre-commit`, `pytest`, `wheel` via `pip` and uses `pre-commit` to run
+the `black` and `flake8` using a pre-commit hook. In this way, you will be prompted a warning whenever you
 forget to run `black` or `flake8` before doing your commit :wink:.
 
 ## Contributors
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ and [citing](#citing-this-work) PyFstat.
 * We also have a number of
 [examples](https://github.com/PyFstat/PyFstat/tree/master/examples),
 demonstrating different use cases.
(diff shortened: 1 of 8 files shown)


bin/get_and_export_ephemeris.sh
```diff
@@ -0,0 +1,20 @@
+# Get ephemeris files and export env variable to access them.
+# Output path is accepted as an input; otherwise, falls back
+# to ../.ephemeris.
+# This script does NOT alter any other bash script.
```
dbkeitel, line 4: What does that mean, why would it?

README.md
```diff
@@ -287,6 +282,23 @@ Here's what you need to know:
   `flake8 --count --statistics .` to find common coding errors and then fix them manually,
   and then
   `black --check --diff .` to show the required style changes, or `black .` to automatically apply them.
+* `bin/setup-dev-tools.sh` gets your virtual environment ready for you. After making sure you are 
+using a virtual environment, it install `pytest`, `black` and `flake8` and uses `pre-commit` to run
```
dbkeitel, line 286:
```suggestion
using a virtual environment (venv or conda),
it installs `pytest`, `black` and `flake8` and uses `pre-commit` to run
```

dbkeitel (commented):
Great idea! Besides the minor documentation comments, the only thing I'll have to double-check is if the new `bin` directory leads to any problems in PyPI and Zenodo releases...

* Should be fine for Zenodo I think, as it just does a tarball of everything.

* For PyPI I should probably do an upload to the test instance, let me see if that's possible from a branch.

bin/get_and_export_ephemeris.sh
Rodrigo-Tenorio: That has to do with the current suggestion on how to get ephemeris in the [README](README.md), which includes the following statement:
```
echo 'export LALPULSAR_DATADIR=$VIRTUAL_ENV/share/lalpulsar' >> ${VIRTUAL_ENV}/bin/activate
```

This makes downloading ephemeris a once-in-a-lifetime process; however, if you opt to use `bin/get_and_export_ephemeris.sh`, you have to do it every single time you start the virtual environment, as otherwise there would be no `LALPULSAR_DATADIR` variable in the environment (and hence the `-nc` option in `wget` is there).

README.md
Rodrigo-Tenorio: Yes, I always mess up the actual name.

dbkeitel (changes requested):
@Rodrigo-Tenorio could you add bash/python shebangs to the three `bin/` scripts please, and also `check_if_virtual_environment.py` doesn't have the executable flag yet.

dbkeitel:
Testing the packaging/distribution and a test.pypi.org upload on this branch, the `bin` directory is not included automatically and so there are no problems. But if in the future we want to include actual executables in releases, that will then require some exclusion rules. If you know a good standard directory name for "executables that are useful for devs but not to be distributed", it would be ideal to rename it now; otherwise, we'll leave it for future us to figure out.

requirements-dev.txt
```diff
@@ -0,0 +1,4 @@
+black
+flake8
+pre-commit
```
dbkeitel, line 3:
```suggestion
wheel
pre-commit
```

1. devs might need to build wheels of PyFstat itself anyway
1. It seems like `pre-commit` pulls in `pyyaml` which then, at least in the virgin venv I tried this out in, can't be built without having the `wheels` package at hand:
```
Building wheels for collected packages: black, pyyaml
  Running setup.py bdist_wheel for black ... done
  Stored in directory: /home/dkeitel/.cache/pip/wheels/6e/10/b5/edf7359c2edd0305cce7e3f96e07daf7ce55dceac9d3ce3373
  Running setup.py bdist_wheel for pyyaml ... error
  Complete output from command /home/dkeitel/git/sources/PyFstat/devtest/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-duh50s2i/pyyaml/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/pip-wheel-ndjjohth --python-tag cp37:
  usage: -c [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
     or: -c --help [cmd1 cmd2 ...]
     or: -c --help-commands
     or: -c cmd --help
  
  error: invalid command 'bdist_wheel'
  
  ----------------------------------------
  Failed building wheel for pyyaml
  Running setup.py clean for pyyaml
Successfully built black
Failed to build pyyaml
```

Rodrigo-Tenorio: > Testing the packaging/distribution and a test.pypi.org upload on this branch, the `bin` directory is not included automatically and so there are no problems. But if in the future we want to include actual executables in releases, that will then require some exclusion rules. If you know a good standard directory name for "executables that are useful for devs but not to be distributed", it would be ideal to rename it now; otherwise, we'll leave it for future us to figure out.

I'm not aware of a standard name for that. We can leave it as `bin` and use `pyfstat_cli` for those "future executables" as done by [bilby](https://git.ligo.org/lscsoft/bilby/-/tree/master/cli_bilby) (this way we also avoid to use the word `dev`, which stands for `device` as well).

dbkeitel: Ok let's leave it at that.

dbkeitel (dismissed):

bin/get-and-export-ephemeris.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
dbkeitel, line 1:
```suggestion
#!/bin/bash
```

bin/setup-dev-tools.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
dbkeitel, line 1:
```suggestion
#!/bin/bash
```

Otherwise this gives a `bad substitution` on line 10.

dbkeitel (changes requested):

dbkeitel (approved):

dbkeitel:
This may not work entirely robustly when switching between different environments, but as a purely optional helper for developers, seems fine to merge as it is now.

Issue #193, opened 2020-09-25
Simplify CI/Test Suite


Rodrigo-Tenorio:
Currently it takes O(5min) to go through all the CI action (mostly because of tests). Can we get it down to O(1min)?

dbkeitel:
On my laptop:
```
pytest --durations=0 tests.py
[...]
===================================================================================== slowest durations =====================================================================================
33.68s call     tests.py::TestMCMCSearch::test_fully_coherent_MCMC
24.77s call     tests.py::TestMCMCFollowUpSearch::test_MCMC_followup_search
6.20s call     tests.py::TestSemiCoherentGlitchSearch::test_get_semicoherent_nglitch_twoF_with_glitch
5.09s call     tests.py::TestSemiCoherentGlitchSearch::test_get_semicoherent_nglitch_twoF_no_glitch
4.93s call     tests.py::TestMCMCTransientSearch::test_transient_MCMC
4.34s call     tests.py::TestMCMCSemiCoherentSearch::test_semi_coherent_MCMC
4.16s call     tests.py::TestSemiCoherentSearch::test_get_semicoherent_twoF
4.11s call     tests.py::TestWriter::test_noise_sfts
3.91s call     tests.py::TestGlitchWriter::test_noise_sfts
3.63s call     tests.py::TestBinaryModulatedWriter::test_tp_parsing
3.62s call     tests.py::TestBinaryModulatedWriter::test_noise_sfts
3.61s call     tests.py::TestComputeFstat::test_run_computefstatistic_single_point_injectSources
3.56s call     tests.py::TestComputeFstat::test_run_computefstatistic_single_point_with_SFTs
3.04s call     tests.py::TestGlitchWriter::test_glitch_injection
3.01s call     tests.py::TestGlitchWriter::test_makefakedata_usecached
2.95s call     tests.py::TestBinaryModulatedWriter::test_data_with_gaps
2.91s call     tests.py::TestBinaryModulatedWriter::test_makefakedata_usecached
2.88s call     tests.py::TestWriter::test_makefakedata_usecached
2.74s call     tests.py::TestGlitchWriter::test_data_with_gaps
2.68s call     tests.py::TestWriter::test_data_with_gaps
2.47s call     tests.py::TestComputeFstat::test_get_fully_coherent_BSGL
2.41s call     tests.py::TestComputeFstatNoNoise::test_run_computefstatistic_single_point_no_noise
[...]

PR #196, opened 2020-09-28
Make tests faster

Rodrigo-Tenorio:
Address #193 .

It won't go faster than this, as the rest of the time seems to be spent in set up commands. 
We should gain 30s and a `--durations=0` flag to keep an eye on what we do.


dbkeitel (approved):

dbkeitel:
Looks fine. Ok to merge before #186?

Rodrigo-Tenorio:
Sure!

Issue #194, opened 2020-09-25
Overhaul: Make consistent use of PyFstat's naming convention


Rodrigo-Tenorio:
PyFstat and LAL conventions on CW arguments are different (`F0` v.s. `Freq` or `phi` v.s. `phi0`).

That shouldn't be a problem, as 1) we have a consistent way to translate PyFstat to LAL keys via `BaseSearchClass.translate_keys_to_lal` and 2) LAL calls do *not* return Python objects (since they run as separated processes) so no inverse conversion is required.

However, as seen in #186 , some PyFstat methods expect LAL keys (e.g. `phi0` in `ComputeFstat`'s `injectParams`). 

It would be a better UX not to use two conventios at once; rather, each class/method should handle its own attributes/arguments to make sure they follow whatever convention required.

Rodrigo-Tenorio:
This should be fixed after merging #186 , as the following bits of code are a consequence of this problem:

- [ ]  `pyfstat.helper_functions.get_predict_fstat_parameters_from_dict`
It manually changes `F0` to `Freq`.

- [ ] `pyfstat.helper_functions.get_predict_fstat_parameters_from_dict`

Rodrigo-Tenorio:
The strategy will be as follows: 

1) Every single PyFstat callable will follow the PyFstat naming conventions.
2) Python objects shall be addressed by the PyFstat naming conventions.
3) Translation to LAL keys will either be done at the moment of calling a LAL function, or not done at all if avoidable.

Let's gauge how big of an overhaul will this be...

Rodrigo-Tenorio:
## core.py
- Affected through `helper_functions.predict_fstat`
- `ComputeFstat.init_computefstatistic`'s `injectSources` is expected to contain `phi0`.

## make_sfts.py
- `Writer.predict_fstat`, which calls `helper_functions.predict_fstat`

## mcmc_based_searches.py
- `MCMCSearch.generate_loudest`
There's an explicit check for `["transient-t0Epoch", "transient-t0Offset", "transient-tau"]` and another
call right before calling to a lal function, so this should be fine.

## helper_functions.py
- `get_predict_fstat_parameters_from_dict`, which is already labeled as `FIXME`.
- `predict_fstat`

Rodrigo-Tenorio:
It's useful to keep `translate_keys_to_lal`, as it provides a dictionary which then can be used to construct arguments for LALSuite using a loop.

PR #197, opened 2020-09-28
consistent parameter naming convention

Rodrigo-Tenorio:
Implement suggestions in #194 

(To be merged after #186 )

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -487,7 +487,7 @@ class ComputeFstat(BaseSearchClass):
                 PP.Amp.h0 = h0
                 PP.Amp.cosi = cosi
 
-            PP.Amp.phi0 = self.injectSources["phi0"]
+            PP.Amp.phi0 = self.injectSources["phi"]
             PP.Amp.psi = self.injectSources["psi"]
             PP.Doppler.Alpha = self.injectSources["Alpha"]
             PP.Doppler.Delta = self.injectSources["Delta"]
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1103,7 +1103,7 @@ class ComputeFstat(BaseSearchClass):
                 psi=psi,
                 Alpha=Alpha,
                 Delta=Delta,
-                Freq=F0,
+                F0=F0,
                 **predict_fstat_kwargs,
             )
             for duration in cumulative_durations
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -150,7 +150,8 @@ class BaseSearchClass:
         )
         return params_dict
 
-    def translate_keys_to_lal(self, dictionary):
+    @staticmethod
+    def translate_keys_to_lal(dictionary):
         """Convert input keys into lal input keys
 
         Input keys are F0, F1, F2, ..., while LAL functions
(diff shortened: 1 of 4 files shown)


Rodrigo-Tenorio:
`injectSources` will be weird: 

- If given as a dictionary, PyFstat naming convention will be on use. Moreover, the user could be able to use `fkdot` style keys as an alternative. One could argue this breaks point 2 in  [the original issue](https://github.com/PyFstat/PyFstat/issues/194#issuecomment-699959489), as well as the Python Zen since there are two ways to do the same thing.  @dbkeitel I'll let you decide.

- If given as a string, it will look for a `cff` and use a LAL function to read it. This is consistent with point 4 in [the original issue](https://github.com/PyFstat/PyFstat/issues/194#issuecomment-699959489), as that happens behind the curtains and the user never needs to interact using non-PyFstat keys (because the creation of said `cff` can be done from `make_sfts`, which uses PyFstat conventions as well).

Rodrigo-Tenorio:
Still have to

- [ ] Fix tests
- [ ] Fix examples

Rodrigo-Tenorio:
There are still some issues with the examples related to #186

pyfstat/helper_functions.py
```diff
@@ -724,19 +724,12 @@ def get_predict_fstat_parameters_from_dict(signal_parameters, return_lal_keys=Fa
     ----------
     signal_parameters: dict
         Dictionary containing at least the signal parameters required by
-        helper_functions.predict_fstat. This dictionary's keys must follow
-        the PyFstat convention (F0 instead of Freq).
-    return_lal_keys: bool
-        If True, the returned dictionary will contain keys according to the
-        lal convention (Freq instead of F0).
+        helper_functions.predict_fstat.
```
dbkeitel, line 727:
```suggestion
        helper_functions.predict_fstat. This dictionary's keys must follow
        the PyFstat convention (F0 instead of Freq).
```

Let's keep these two lines for clarity, even if it's just "yes, global policy isn't violated here" - due to the close connection of the function with a lalapps code, it's worth saying explicitly.

dbkeitel (changes requested):
Just one partial revert suggestion on a help string, otherwise looks great (and much simpler than initially suspected).

pyfstat/core.py
```diff
@@ -150,7 +150,8 @@ def read_par(
         )
         return params_dict
 
-    def translate_keys_to_lal(self, dictionary):
+    @staticmethod
+    def translate_keys_to_lal(dictionary):
```
Rodrigo-Tenorio, line 154: @dbkeitel This may be controversial.
I just think about those times in which `lal` functions have to be invoked via `subprocess` when `PyFstat` is already on use: Then it makes sense to have this method available (otherwise you'd have to code it yourself *every single time*).

dbkeitel: Just so I remember what this does, after chat clarification: The idea is that you can do
```
from pyfstat import BaseSearchClass
translated_dict = BaseSearchClass.translate_keys_to_lal(mydict)
```
without having init-ed an object.

The alternative would be to make it a helper_function, but since as you say it won't be a commonly needed feature, this should suffice.

dbkeitel (approved):

Issue #190, opened 2020-09-25
Check BinaryModulatedWriter tests 


Rodrigo-Tenorio:
After some big changes, we should check whether BinaryModulatedWriter is actually being tested using binary parameters or not.

PR #198, opened 2020-09-28
Implement tests for BinaryModulatedWriter

Rodrigo-Tenorio:
- Solve #190 
- Further inspection required, as it fails when using noise SFTs.
- And, of course, another one after #186


Rodrigo-Tenorio:
I found #199 while playing around. I'd say it should have more priority than this one.

PR #200, opened 2020-09-28
Test linear polarization

Rodrigo-Tenorio:
Expose bug in #199 

(Issue/PR 200! :cookie: )

Edit: Rebased on top of master after latest PRs.

PR #201, opened 2020-09-30
core.py/Add more sigmas to predict Fstat plot

Rodrigo-Tenorio:
Could it be the case that we are simply using `predict_fstat` in a very conservative way?
PyFstat_example_twoF_cumulative_L1_twoFcumulative
PyFstat_example_twoF_cumulative_H1_twoFcumulative
PyFstat_example_twoF_cumulative_H1L1_twoFcumulative


Rodrigo-Tenorio:
Not really and explanation for #200.

Issue #202, opened 2020-10-05
Wrong expected SFT filenames when duration is not multiple of Tsft (e.g. for Tsft=1024s in many examples)


Rodrigo-Tenorio:
The following happens when `Tsft` is set to `1024` instead of `1800` in the `twoF_cumulative` example.
```
(pyfstat) [rodrigo@spinning-quadrupole other_examples]$ python PyFstat_example_twoF_cumulative.py
11:05 INFO    : Creating BinaryModulatedWriter object...
11:05 INFO    : Generating SFTs with fmin=28.0, Band=4
11:05 INFO    : Checking if we can re-use injection config file...
11:05 INFO    : ...no config file PyFstat_example_data/PyFstat_example_twoF_cumulative/PyFstat_example_twoF_cumulative.cff found.
11:05 INFO    : Writing config file: PyFstat_example_data/PyFstat_example_twoF_cumulative/PyFstat_example_twoF_cumulative.cff
11:05 INFO    : Checking if we can re-use existing SFT data file(s)...
11:05 INFO    : ...no SFT file matching 'PyFstat_example_data/PyFstat_example_twoF_cumulative/H-843_H1_1024SFT_PyFstat_example_twoF_cumulative-1000000000-864000.sft' found. Will create new SFT file(s).
11:05 INFO    : Now executing: lalapps_Makefakedata_v5 --outSingleSFT=TRUE --outSFTdir="PyFstat_example_data/PyFstat_example_twoF_cumulative" --outLabel="PyFstat_example_twoF_cumulative" --IFOs="H1","L1" --sqrtSX="1e-23" --startTime=1000000000 --duration=864000 --fmin=28 --Band=4 --Tsft=1024 --injectionSources="PyFstat_example_data/PyFstat_example_twoF_cumulative/PyFstat_example_twoF_cumulative.cff" --ephemEarth="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/earth00-40-DE405.dat.gz" --ephemSun="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/sun00-40-DE405.dat.gz"
[...]

PR #203, opened 2020-10-05
Predict 2F: Make an explicit error if duration is not a multiple of Tsft

Rodrigo-Tenorio:
Exposes #202.


dbkeitel:
Thanks, this was useful for tracking down the actual `Writer`-side issue causing #202. (Though actually I ended up using `examples/grid_examples/PyFstat_example_grid_search_F0.py` as a faster check.) Closing here, but should add a proper test case to #205.

PR #204, opened 2020-10-07
improve sqrtSX handling

dbkeitel:
Several classes/functions take sqrtSX-type inputs which can be single values or per-IFO lists, and their type can in principle be either a float, list of floats, list of strings or comma-separated string. These are then variously passed to lalapps commandlines (requiring simple floats or strings), passed to swig-wrapped lal functions (requiring lists) or checked PyFstat-internally for being > or < 0 (currently failing if a list or string).

We could somehow curb this chaos at the highest-level user-facing input stage by only accepting one format, but here's a semi-elegant attempt to be robust in parsing this.

One uncovered failure mode is if the user input is a list and a command-line gets constructed from it with plain `format`, but at least the error message from lalapps should be clear in that case.

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -286,12 +286,19 @@ class ComputeFstat(BaseSearchClass):
         injectSources : dict or str
             Either a dictionary of the values to inject, or a string pointing
             to the .cff file to inject
-        injectSqrtSX :
-            Per-IFO single-sided PSD values for generating fake Gaussian noise on the fly
-        assumeSqrtSX : float
-            Don't estimate noise-floors but assume (stationary) per-IFO
-            sqrt{SX} (if single value: use for all IFOs). If signal only,
-            set sqrtSX=1
+        injectSqrtSX : float or list or str
+            Single-sided PSD values for generating fake Gaussian noise on the fly.
+            Single float or str value: use same for all IFOs.
+            List or comma-separated string: must match len(detectors)
+            and/or the data in sftfilepattern.
+            Detectors will be paired to list elements following alphabetical order.
+        assumeSqrtSX : float or list or str
+            Don't estimate noise-floors but assume this (stationary) single-sided PSD.
+            Single float or str value: use same for all IFOs.
+            List or comma-separated string: must match len(detectors)
+            and/or the data in sftfilepattern.
+            Detectors will be paired to list elements following alphabetical order.
+            If working with signal-only data, please set assumeSqrtSX=1 .
         SSBprec : int
             Flag to set the SSB calculation: 0=Newtonian, 1=relativistic,
             2=relativisitic optimised, 3=DMoff, 4=NO_SPIN
@@ -466,7 +473,7 @@ class ComputeFstat(BaseSearchClass):
             FstatOAs.assumeSqrtSX = lalpulsar.FstatOptionalArgsDefaults.assumeSqrtSX
         else:
             mnf = lalpulsar.MultiNoiseFloor()
-            assumeSqrtSX = np.atleast_1d(self.assumeSqrtSX)
+            assumeSqrtSX = helper_functions.parse_list_of_numbers(self.assumeSqrtSX)
             mnf.sqrtSn[: len(assumeSqrtSX)] = assumeSqrtSX
             mnf.length = len(assumeSqrtSX)
             FstatOAs.assumeSqrtSX = mnf
@@ -511,7 +518,9 @@ class ComputeFstat(BaseSearchClass):
         else:
             FstatOAs.injectSources = lalpulsar.FstatOptionalArgsDefaults.injectSources
         if hasattr(self, "injectSqrtSX") and self.injectSqrtSX is not None:
-            self.injectSqrtSX = np.atleast_1d(self.injectSqrtSX)
+            self.injectSqrtSX = helper_functions.parse_list_of_numbers(
+                self.injectSqrtSX
+            )
             if len(self.injectSqrtSX) != len(self.detector_names):
                 raise ValueError(
                     "injectSqrtSX must be of same length as detector_names ({}!={})".format(
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -776,9 +776,12 @@ def predict_fstat(
         Comma-separated list of detectors.
         Ignored if sftfilepattern is given,
         required if it is not.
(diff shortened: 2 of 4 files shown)


dbkeitel:
Example action running: https://github.com/PyFstat/PyFstat/runs/1220736799?check_suite_focus=true

pyfstat/core.py
```diff
@@ -286,12 +286,17 @@ def __init__(
...
-            Per-IFO single-sided PSD values for generating fake Gaussian noise on the fly
-        assumeSqrtSX : float
-            Don't estimate noise-floors but assume (stationary) per-IFO
-            sqrt{SX} (if single value: use for all IFOs). If signal only,
-            set sqrtSX=1
+        injectSqrtSX : float or list or str
+            Single-sided PSD values for generating fake Gaussian noise on the fly.
+            Single float or str value: use same for all IFOs.
+            List or comma-separated string: must match len(detectors)
+            and/or the data in sftfilepattern.
```
Rodrigo-Tenorio, line 293: Let's be explicit:
```suggestion
            and/or the data in sftfilepattern. Detectors will be paired to list elements following alphabetical order.
```

pyfstat/core.py
```diff
@@ -286,12 +286,17 @@ def __init__(
...
+        assumeSqrtSX : float or list or str
+            Don't estimate noise-floors but assume this (stationary) single-sided PSD.
+            Single float or str value: use same for all IFOs.
+            List or comma-separated string: must match len(detectors)
+            and/or the data in sftfilepattern.
```
Rodrigo-Tenorio, line 298: Same proposal:
```suggestion
            and/or the data in sftfilepattern. Detectors will be paired to list elements following alphabetical order.
```

So the difference is `injectSqrtSX` adds Gaussian noise to the data and estimates PSD from the data itself, while `assumeSqrtSX` just says "PSD value is this one."?

pyfstat/core.py
```diff
@@ -286,12 +286,17 @@ def __init__(
...
+            If working with signal-only data, please set sqrtSX=1 .
```
Rodrigo-Tenorio, line 299: Shouldn't this be
```suggestion
            If working with signal-only data, please set assumeSqrtSX=1 .
```
?

pyfstat/helper_functions.py
```diff
@@ -776,9 +776,11 @@ def predict_fstat(
         Comma-separated list of detectors.
         Ignored if sftfilepattern is given,
         required if it is not.
-    assumeSqrtSX : float or str or None
+    assumeSqrtSX : float or str
         Assume stationary per-detector noise-floor instead of estimating from SFTs.
-        For multiple detectors: comma-separated list
+        Single float or str value: use same for all IFOs.
+        Comma-separated string: must match len(detectors)
+        and/or the data in sftfilepattern.
```
Rodrigo-Tenorio, line 783:
```suggestion
        and/or the data in sftfilepattern. Detectors will be paired to list elements following alphabetical order.
```

pyfstat/helper_functions.py
```diff
@@ -880,3 +882,19 @@ def predict_fstat(
...
+    to try either type of input.
+    """
+    try:
+        out = list(map(float, val.split(",")))
+    except Exception:
+        try:
+            out = list(map(float, np.atleast_1d(val)))
+        except Exception:
+            raise ValueError(f"Could not parse '{val}' as a number or list of numbers.")
+    return out
```
Rodrigo-Tenorio, line 900: I feel these `try`s should be more explicit, but I don't know how...
Maybe 
```suggestion
    try:
        out = list(map(float, val.split(",") 
        if hassattr(val, "split")
        else np.atleast_1d(val)) 
    except Exception as e:
        raise ValueError(f"Could not parse '{val}' as a number or list of numbers: {e.message}")
```

pyfstat/mcmc_based_searches.py
```diff
@@ -76,7 +76,7 @@ class MCMCSearch(BaseSearchClass):
     injectSources: dict, optional
         If given, inject these properties into the SFT files before running
         the search
-    assumeSqrtSX: float, optional
+    assumeSqrtSX: float or list or str
         Don't estimate noise-floors, but assume (stationary) per-IFO sqrt{SX}
```
Rodrigo-Tenorio, line 80:
```suggestion
        Don't estimate noise-floors, but assume (stationary) per-IFO sqrt{SX}. See `core.ComputeFstat`.
```

pyfstat/mcmc_based_searches.py
```diff
@@ -2555,7 +2555,7 @@ class MCMCSemiCoherentSearch(MCMCSearch):
     injectSources: dict, optional
         If given, inject these properties into the SFT files before running
         the search
-    assumeSqrtSX: float, optional
+    assumeSqrtSX: float or list or str, optional
         Don't estimate noise-floors, but assume (stationary) per-IFO sqrt{SX}
```
Rodrigo-Tenorio, line 2559:
```suggestion
    Don't estimate noise-floors, but assume (stationary) per-IFO sqrt{SX}. See `core.ComputeFstat`.
```

Rodrigo-Tenorio (changes requested):
Two things:

1) Improve documentation to be more explicit about the IFO order (which is alphabetical according to LAL).
2) Implement parsing function in a simpler way.

pyfstat/core.py
dbkeitel: I think/hope the estimation is completely independent of this option, one should be able to use it with/without `assumeSqrtSX".

pyfstat/core.py
dbkeitel: Yes!

pyfstat/helper_functions.py
dbkeitel: That doesn't look black-legal, but I like the if-else simplification. Will try it locally.

Rodrigo-Tenorio: I'd bet it's not (I'm actually quite surprised it came out with the "right" identation :upside_down_face: )

dbkeitel: Resolved, but just returning `e` not `e.message`.

Rodrigo-Tenorio (approved):

Rodrigo-Tenorio (approved):

Rodrigo-Tenorio:
Looks right to me.

PR #205, opened 2020-10-07
Fix expected SFT file names when duration is not multiple of Tsft

dbkeitel:
I think this is sufficient to fix #202. The alternative `_get_setup_from_noiseSFTs()` case should be fine already, since there we are just taking the SFT file names as we've been given them, but I haven't dug explicitly into it.

- [ ] TODO: check MFDv5 code/documentation to confirm we now actually _always_ match its behaviour, and I haven't left out weird corner cases.
- [x] TODO: add a test case

PR #207, opened 2020-10-08
tests/Add BinaryModulatedWriter tests

### LINKED ISSUE #190: Check BinaryModulatedWriter tests

Rodrigo-Tenorio:
Uncouples Writer and signal parameters in order to add binary parameters through BinaryModulatedWriter through a class attribute. Fixes #190 .

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -196,7 +196,8 @@ class TestWriter(BaseForTestsWithData):
         max_values_noise_and_signal = np.max(data, axis=0)
         max_freqs_noise_and_signal = freqs[np.argmax(data, axis=0)]
         self.assertTrue(len(times) == int(np.ceil(self.duration / self.Tsft)))
-        # with signal: all SFTs should peak at same freq
+        # FIXME: CW signals don't have to peak at the same frequency, but there
+        # are some consistency criteria which may be useful to implement here.
         # self.assertTrue(len(np.unique(max_freqs_noise_and_signal)) == 1)
 
         # create noise-only SFTs
--- a/tests.py
+++ b/tests.py
@@ -31,6 +31,19 @@ class BaseForTestsWithOutdir(unittest.TestCase):
                 logging.warning("{} not removed after tests".format(self.outdir))
 
 
+default_Writer_params = {
+    "label": "test",
+    "sqrtSX": 1,
+    "Tsft": 1800,
+    "tstart": 700000000,
+    "duration": 4 * 1800,
+    "detectors": "H1",
+    "SFTWindowType": "tukey",
+    "SFTWindowBeta": 0.001,
+    "randSeed": 42,
+    "Band": None,
+}
+
 default_signal_params = {
     "F0": 30,
     "F1": -1e-10,
@@ -43,19 +56,6 @@ default_signal_params = {
     "Delta": 1.2,
 }
 
-default_Writer_params = {
-    "label": "test",
-    "sqrtSX": 1,
-    "Tsft": 1800,
-    "tstart": 700000000,
-    "duration": 4 * 1800,
-    "detectors": "H1",
-    "SFTWindowType": "tukey",
-    "SFTWindowBeta": 0.001,
-    "randSeed": 42,
-    "Band": None,
-    **default_signal_params,
-}
 
 default_binary_params = {
     "period": 45 * 24 * 3600.0,
@@ -84,7 +84,7 @@ class BaseForTestsWithData(BaseForTestsWithOutdir):
         # if we directly set any options as self.xy = 1 here,
         # then values set for derived classes may get overwritten,
         # so use a default dict and only insert if no value previous set
-        for key, val in default_Writer_params.items():
+        for key, val in {**default_Writer_params, **default_signal_params}.items():
             if not hasattr(self, key):
                 setattr(self, key, val)
         self.tref = self.tstart
@@ -117,6 +117,7 @@ class BaseForTestsWithData(BaseForTestsWithOutdir):
 class TestWriter(BaseForTestsWithData):
     label = "TestWriter"
     writer_class_to_test = pyfstat.Writer
+    signal_parameters = default_signal_params
 
     def test_make_cff(self):
         self.Writer.make_cff(verbose=True)
@@ -182,7 +183,7 @@ class TestWriter(BaseForTestsWithData):
             SFTWindowType=self.SFTWindowType,
             SFTWindowBeta=self.SFTWindowBeta,
             sqrtSX=self.sqrtSX,
-            **default_signal_params,
+            **self.signal_parameters,
         )
         noise_and_signal_writer.make_data(verbose=True)
         # FIXME: here and everywhere below,
@@ -233,7 +234,7 @@ class TestWriter(BaseForTestsWithData):
             SFTWindowBeta=self.SFTWindowBeta,
             sqrtSX=0,
             noiseSFTs=noise_writer.sftfilepath,
-            **default_signal_params,
+            **self.signal_parameters,
         )
         add_signal_writer.make_data(verbose=True)
         times, freqs, data = pyfstat.helper_functions.get_sft_array(
@@ -262,7 +263,7 @@ class TestWriter(BaseForTestsWithData):
             SFTWindowBeta=self.SFTWindowBeta,
             sqrtSX=0,
             noiseSFTs=noise_writer.sftfilepath,
-            **default_signal_params,
+            **self.signal_parameters,
         )
         add_signal_writer_constr.make_data(verbose=True)
         times, freqs, data = pyfstat.helper_functions.get_sft_array(
@@ -364,12 +365,11 @@ class TestWriterOtherTsft(TestWriter):
 class TestBinaryModulatedWriter(TestWriter):
     label = "TestBinaryModulatedWriter"
     writer_class_to_test = pyfstat.BinaryModulatedWriter
+    signal_parameters = {**default_signal_params, **default_binary_params}
 
     def test_tp_parsing(self):
         this_writer = self.writer_class_to_test(
-            outdir=self.outdir,
-            **default_Writer_params,
-            **default_binary_params,
+            outdir=self.outdir, **default_Writer_params, **self.signal_parameters
         )
         this_writer.make_data()
 
--- a/tests.py
+++ b/tests.py
@@ -48,11 +48,11 @@ default_signal_params = {
     "F0": 30,
     "F1": -1e-10,
     "F2": 0,
-    "h0": 5,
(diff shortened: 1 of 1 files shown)


tests.py
```diff
@@ -192,9 +192,9 @@ def test_noise_sfts(self):
         )
         max_values_noise_and_signal = np.max(data, axis=0)
         max_freqs_noise_and_signal = freqs[np.argmax(data, axis=0)]
-        self.assertTrue(len(times) == int(np.ceil(self.duration / self.Tsft)))
+        self.assertTrue(len(times) == self.duration / self.Tsft)
         # with signal: all SFTs should peak at same freq
-        self.assertTrue(len(np.unique(max_freqs_noise_and_signal)) == 1)
+        # self.assertTrue(len(np.unique(max_freqs_noise_and_signal)) == 1)
```
Rodrigo-Tenorio, line 197: I don't think agree with this statement @dbkeitel : The whole point of CW is that signals suffer from a frequency modulation, hence this does *not* have to be true in general (It's true we assume `Tsft` such that the signal is "reasonably well contained", but longer time scale modulations can be present).

Rodrigo-Tenorio:
This is going to be fun:

We could keep the same strategy as for isolated by choosing`asini` and `period` parameters such that the binary modulation resembles a very faint spindown; this would make tests pass; but that also implies `tp` (which, remember, is a *phase*) becomes degenerated, breaking the second test.

tests.py
```diff
@@ -245,11 +245,11 @@ def test_noise_sfts(self):
...
-        self.assertTrue(
-            np.allclose(
-                max_values_added_signal, max_values_noise_and_signal, rtol=1e-6, atol=0
-            )
-        )
+        # self.assertTrue(
+        #    np.allclose(
+        #        max_values_added_signal, max_values_noise_and_signal, rtol=1e-6, atol=0
+        #    )
+        # )
```
Rodrigo-Tenorio, line 252: Regarding the actual value of the peaks, we are somehow digging our own grave (again): Noise is random, that doesn't have to be true.

I understand that the relative difference should "generally" between well understood intervals, so I'm more willing to make this snippet work.

tests.py
dbkeitel: Absolutely correct, this won't be true in general. Only for very short signals without extra (e.g. binary) modulation.

We could check the expected modulation in +- bin counts and require the max to be within that?

tests.py
dbkeitel: This part explicitly compares the two paths of "create noise and signal together" vs "first create noise, then inject signal" with the same random seed for the noise on both paths. So I think this one actually should always be able to pass at good precision.

dbkeitel:
> We could keep the same strategy as for isolated by choosingasini and period parameters such that the binary modulation resembles a very faint spindown; this would make tests pass; but that also implies tp (which, remember, is a phase) becomes degenerated, breaking the second test.

"The second test" = which one?

Rodrigo-Tenorio:
Ups: I mean the test for `tp` being properly parsed.

tests.py
Rodrigo-Tenorio: I agree: Nuke the "unique frequency peak", make this one work.

Rodrigo-Tenorio: Interesting, if I use `rtol=1e-6` and print relevant quantities to [numpy.allclose](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html):
```
print(f"Absolute diff: {max_values_added_signal-max_values_noise_and_signal} , rtols: {1e-6*max_values_added_signal}, {1e-6 * max_values_noise_and_signal}")

Absolute diff: [  7.199463    8.605072  -18.658844   -1.1877441] , rtols: [0.00051476 0.00016379 0.00031445 0.00068457], [0.00050756 0.00015519 0.00033311 0.00068576]
```

Rodrigo-Tenorio: And it persists if we just scale `SqrtSX` and `h0` down to `1e-22`:
```
Absolute diff: [ 7.1995086e-22  8.6050694e-22 -1.8658836e-21 -1.1877797e-22] , rtols: [5.1476252e-26 1.6379310e-26 3.1444997e-26 6.8456936e-26], [5.0756303e-26 1.5518803e-26 3.3310881e-26 6.8575715e-26]
```

Rodrigo-Tenorio:
Noise and data:
```
15:04 INFO    : Creating BinaryModulatedWriter object...
15:04 INFO    : Estimating required SFT frequency range from properties of signal to inject plus 59 extra bins either side (corresponding to default F-statistic settings).
15:04 INFO    : Generating SFTs with fmin=29.963073443256597, Band=0.07385239358636084
```
Noise only:
```
15:04 INFO    : Creating BinaryModulatedWriter object...
15:04 INFO    : Estimating required SFT frequency range from properties of signal to inject plus 59 extra bins either side (corresponding to default F-statistic settings).
15:04 INFO    : Generating SFTs with fmin=29.964043790519263, Band=0.07191241896147502
```

tests.py
Rodrigo-Tenorio: Solved in https://github.com/PyFstat/PyFstat/pull/207#issuecomment-705555667

tests.py
```diff
@@ -126,17 +127,16 @@ def test_make_cff(self):
 
     def test_run_makefakedata(self):
         self.Writer.make_data(verbose=True)
-        numSFTs = int(np.ceil(self.duration / self.Tsft))
```
dbkeitel, line 129: This and the next 2 diffs looks like a rebase accident...? This was just introducede in #205.

tests.py
```diff
@@ -182,19 +182,19 @@ def test_noise_sfts(self):
...
         times, freqs, data = pyfstat.helper_functions.get_sft_array(
             noise_and_signal_writer.sftfilepath
         )
+        # Store the maximum SFT values for later comparison
         max_values_noise_and_signal = np.max(data, axis=0)
         max_freqs_noise_and_signal = freqs[np.argmax(data, axis=0)]
-        self.assertTrue(len(times) == int(np.ceil(self.duration / self.Tsft)))
-        # with signal: all SFTs should peak at same freq
-        self.assertTrue(len(np.unique(max_freqs_noise_and_signal)) == 1)
+        self.assertTrue(len(times) == self.duration / self.Tsft)
```
dbkeitel, line 197: Can we keep the same-bin test commented out, with a FIXME that it should account for bin drift to be properly useful?
Also again the `np.ceil` should stay I think.

tests.py
```diff
@@ -226,14 +226,10 @@ def test_noise_sfts(self):
         add_signal_writer = self.writer_class_to_test(
             label="test_noiseSFTs_add_signal",
             outdir=self.outdir,
-            duration=None,
-            Tsft=self.Tsft,
-            tstart=None,
             SFTWindowType=self.SFTWindowType,
             SFTWindowBeta=self.SFTWindowBeta,
-            sqrtSX=0,
```
dbkeitel, line 234: Seems to make sense to remove these, with `noiseSFTs` I hope they're ignored anyway.

tests.py
```diff
@@ -208,8 +208,8 @@ def test_noise_sfts(self):
             SFTWindowType=self.SFTWindowType,
             SFTWindowBeta=self.SFTWindowBeta,
             sqrtSX=self.sqrtSX,
-            h0=0,
-            F0=self.F0,
+            Band=0.5,
```
dbkeitel, line 211: Does this solve the band estimation issue you mentioned in a comment?

tests.py
```diff
@@ -355,21 +351,14 @@ def test_data_with_gaps(self):
         self.assertTrue(os.path.isfile(expected_SFT_filepath))
 
 
-class TestWriterOtherTsft(TestWriter):
-    label = "TestWriterOtherTsft"
-    writer_class_to_test = pyfstat.Writer
-    Tsft = 1024
-
-
```
dbkeitel, line 363: Also seems to have died in rebase, please reinstate.

dbkeitel (changes requested):
Some rebase accidents to fix, otherwise mostly looks good.

tests.py
dbkeitel: I mean this one: https://github.com/PyFstat/PyFstat/pull/207#issuecomment-705555667

tests.py
```diff
@@ -54,9 +42,21 @@ def tearDownClass(self):
...
+    "h0": 5,
+    "cosi": 0,
+    "psi": 0,
+    "phi": 0,
+    "Alpha": 5e-3,
+    "Delta": 1.2,
+}
+
+
 default_binary_params = {
```
dbkeitel, line 60: Putting this here to get a nice thread view:

> Rodrigo: We could keep the same strategy as for isolated by choosingasini and period parameters such that the binary modulation resembles a very faint spindown; this would make tests pass; but that also implies tp (which, remember, is a phase) becomes degenerated, breaking the second test.

> David: "The second test" = which one?

> Rodrigo: Ups: I mean the test for tp being properly parsed.

I don't quite follow - are you saying some different choice of these parameters (specifically tp) would make the `TestBinaryModulatedWriter::test_tp_parsing` fail? But _not_ because the parser itself fails, but just because that test actually runs a MCMC and that won't recover the value as expected if there isn't actually sufficient binary modulation to resolve it?

tests.py
Rodrigo-Tenorio: I'm equally surprised. I'll fix it.

tests.py
Rodrigo-Tenorio: They are. The explanation is "This test precedes the overhaul of `noiseSFTs`".

tests.py
Rodrigo-Tenorio: Yes. Using `F0` and `Band` takes preference over anything else.

tests.py
Rodrigo-Tenorio: If `asini` and `period` are such that the signal resembles an isolated signal over the observing time (big `asini`, low `period`), then `tp` becomes degenerated and the MCMC will not recover the injected value.

dbkeitel: Ok. In that sense the test is overkill, but it seems to pass at the moment, so all good I guess.

Rodrigo-Tenorio:
That was unexpected. 
- [ ] Does MFD injecta signal for `all` the SFTs if `duration < SFT duration`?

Rodrigo-Tenorio:
The problem was related to a bad choice of sky positon: Projected amplitude (i.e. after taking into account detector antenna function) was close to a minimum in one of the SFTs.

Changing the sky position of the generated signal was enough to make tests pass.

dbkeitel (approved):

PR #211, opened 2020-10-26
Implement injection parameter generator

Rodrigo-Tenorio:
Implement a class to generate injection parameters. 

The idea would be the same one one has been doing so far: Pick a rng and draw injection parameters from it. But, instead
of re-inventing the wheel every time, now we could just pass those arguments to this new class an query injection parameters on demand.

- [x] Finish up the implementation.
- [x] Deal with "constant parameters" (just make sure you don't have to code them too explicitly).
- [x] Fill up tests cases.

More desirable things:
- ~[ ] Make this class an iterable?~ Somehow I feel a callable makes more sense.
- [x] Add a subclass which already includes all sky sampling.

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -44,9 +44,9 @@ class InjectionParametersGenerator:
         seed:
             `seed` argument to be feed to numpy.random.default_rng.
         """
-        self._rng = np.random.default_rng(seed)
-        if priors is not None:
-            self.priors = priors
+        self.seed = seed
+        self._rng = np.random.default_rng(self.seed)
+        self.priors = priors or {}
 
     @property
     def seed(self):
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -72,7 +72,7 @@ class InjectionParametersGenerator:
         return injection_parameters
 
     def __call__(self):
-        self.return_injection_parameters()
+        return self.return_injection_parameters()
 
 
 class Writer(BaseSearchClass):
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -40,45 +40,77 @@ class InjectionParametersGenerator:
             while each value must contain a dictionary with its key as one of the available random
             number generators of numpy and its values as kwargs.
         """
-        self.set_outdir(outdir)
-        self.set_label(label)
-        self.set_parameter_priors(parameter_priors)
+        self.outdir = outdir
+        self.label = label
+        self.parameter_priors = parameter_priors
 
         self._rng = np.random.default_rng()
 
         self.injection_parameters = None
 
-    def set_outdir(self, outdir):
-        """Create output folder"""
-        self.outdir = outdir + ("/" if outdir[-1] != "/" else "")
-        os.makedirs(outdir, exist_ok=True)
+    @property
+    def outdir(self):
+        return self._outdir
 
-    def set_label(self, label):
+    @property
+    def label(self):
+        return self._label
+
+    @property
+    def parameter_priors(self):
+        return self._parameter_priors
+
+    def check_out_files(self):
+        """
+        Check if output files already exist to prevent
+        accidental overwritting
+        """
+        pass
+
+    @outdir.setter
+    def outdir(self, new_outdir):
+        """Set output dir name.
+
+        Output dir(s) will not be created until needed.
+
+        Parameters
+        ----------
+        new_outdir: str
+            Output directory name. Trailing slash is irrelevant, as
+            os.path.join will be used to construct the actual filenames.
+        """
+        self._outdir = new_outdir
+
+    @label.setter
+    def label(self, new_label):
         """Set injection set's label
 
         A warning wil be raised if there exists a file in the same folder with
         the same name.
         """
-        self.label = label + ".mdfparameters"
+        self._label = new_label
         if os.path.isfile(self.outdir + self.label):
             raise FileExistsError(
                 "Injection file {} already exists in {}".format(self.label, self.outdir)
             )
 
-    def set_parameter_priors(self, parameter_priors):
-        """Check parameter_priors' format"""
-        for parameter in parameter_priors.keys():
-            if len(parameter_priors[parameter].keys()) != 1:
+    @parameter_priors.setter
+    def parameter_priors(self, new_parameter_priors):
+        """Set priors to drawn parameter space points from """
+
+        # Check parameter_priors' format
+        for parameter in new_parameter_priors.keys():
+            if len(new_parameter_priors[parameter].keys()) != 1:
                 raise ValueError(
                     "Some paramer ranges contain more than one"
                     "random number generator per parameter."
                     "Please, check your parameter ranges dict."
                 )
 
-        self.parameter_priors = parameter_priors
+        self._parameter_priors = new_parameter_priors
         logging.info(
             "Updating parameters: "
-            + " ".join(["{}".format(key) for key in parameter_priors.keys()])
+            + " ".join(["{}".format(key) for key in new_parameter_priors.keys()])
         )
 
     def generate_injection_parameters(self, number_of_injections):
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -38,8 +38,17 @@ class InjectionParametersGenerator:
         self.parameter_priors = parameter_priors
 
     @property
+    def seed(self):
+        return self._seed
+
+    @property
     def parameter_priors(self):
         return self._parameter_priors
+
+    @seed.setter
+    def seed(self, new_seed):
+        self._seed = new_seed
+        self._rng = np.random.default_rng(self._seed)
 
     @parameter_priors.setter
     def parameter_priors(self, new_parameter_priors):
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -105,13 +105,16 @@ class AllSkyInjectionParametersGenerator(InjectionParametersGenerator):
 
     def __init__(self, priors=None, seed=None):
 
-        self.restricted_priors_set = False
-        super().__init__(None, seed)
         self.restricted_priors = {
             "Alpha": lambda: self._rng.uniform(low=0.0, high=2 * np.pi),
             "Delta": lambda: 2 * np.arcsin(self._rng.uniform(low=-1.0, high=1.0)),
         }
-        InjectionParametersGenerator.priors.fset(self, self.restricted_priors)
+        self.restricted_priors_set = False
+
+        super().__init__(None, seed)
+        self.priors = (priors or {}).copy()
+        self.priors.update(self.restricted_priors)
+        self.restricted_priors_set = True
 
     def _check_if_updating_sky_priors(self, new_priors):
         if self.restricted_priors_set and any(
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -2,6 +2,7 @@
 
 
 import numpy as np
+import functools
 import logging
 import os
 import glob
@@ -52,7 +53,7 @@ class InjectionParametersGenerator:
 
     @parameter_priors.setter
     def parameter_priors(self, new_parameter_priors):
-        """Set priors to drawn parameter space points from """
+        """Set priors to draw parameter space points from """
 
         current_priors = getattr(self, "parameter_priors", {})
         for parameter_name, parameter_prior in new_parameter_priors.items():
@@ -63,7 +64,7 @@ class InjectionParametersGenerator:
                 rng_function = getattr(self._rng, rng_function_name)
                 rng_kwargs = parameter_prior[rng_function_name]
                 current_priors.update(
-                    {parameter_name: (lambda: rng_function(**rng_kwargs))}
+                    {parameter_name: functools.partial(rng_function, **rng_kwargs)}
                 )
 
         self._parameter_priors = current_priors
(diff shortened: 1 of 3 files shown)


Rodrigo-Tenorio:
Current state is related to `parameter_priors` and `generate_injection_parameters`.

The idea would be to return a *list of dictionaries*: Each dictionary corresponds to a software injection, parameters are labeled using dictionary keys, so one can easily unpack them into e.g. `make_sfts.Writer`.

However, we should support joint priors (i.e. functions returning *multiple* parameters because they are prescribed to be correlated), and that's not that straighforwards. Current idea is to use a list of keys as key in such a way that `len(keys) == len(prior output)`, but this may change.

Rodrigo-Tenorio:
Discarded joint priors idea: Use case is mostly based on uncorrelated parameters so go for it.

Rodrigo-Tenorio:
Current problem:
```
>>> a = pyfstat.InjectionParametersGenerator({"A": {"uniform": {"low": 1.0, "high": 1.0}}, "B": {"uniform": {"low": 0., "high": 0.0}}})
17:47 INFO    : Updated parameters: A B
>>> a()
{'A': 0.0, 'B': 0.0}
```
Seriously?

Rodrigo-Tenorio:
Let's dig deeper:
```
>>> import pyfstat
>>> ipg = pyfstat.InjectionParametersGenerator
>>> a = ipg({"ParamA": {"uniform": {"low": 0, "high": 0}}})
18:38 INFO    : Updated parameters: ParamA
>>> F1 = lambda :getattr(a._rng, "uniform")(**{"low": 0, "high": 0})
>>> F2 = lambda :getattr(a._rng, "uniform")(**{"low": 1, "high": 1})
>>> F1
<function <lambda> at 0x7f9f691ce0d0>
>>> F2
<function <lambda> at 0x7f9f4dfdfca0>
>>> F1()
0.0
>>> F2()
1.0
```

Rodrigo-Tenorio:
Solved using `functools.partial`. Here some further info https://stackoverflow.com/questions/19837486/python-lambda-in-a-loop .

pyfstat/make_sfts.py
```diff
@@ -23,6 +24,102 @@ class KeyboardInterruptError(Exception):
...
+        super().__init__(parameter_priors, seed)
+
+    def _check_if_updating_sky_priors(self, new_parameter_priors):
+        if any(
+            restricted_key in new_parameter_priors
+            for restricted_key in self.restricted_priors.keys()
+        ):
+            raise ValueError(
+                "New parameter priors would overwrite sky priors (Alpha, Delta)."
+                "This class is explicitiely coded to prevent that from happening. Please, restore "
```
dbkeitel, line 113: explicitly

pyfstat/make_sfts.py
```diff
@@ -23,6 +24,102 @@ class KeyboardInterruptError(Exception):
...
+        self._rng = np.random.default_rng(self._seed)
+
+    @parameter_priors.setter
+    def parameter_priors(self, new_parameter_priors):
+        """Set priors to draw parameter space points from """
+
+        current_priors = getattr(self, "parameter_priors", {})
+        for parameter_name, parameter_prior in new_parameter_priors.items():
+            if callable(parameter_prior):
+                current_priors.update({parameter_name: parameter_prior})
```
dbkeitel, line 62: here, 67 and 72 the plain `dict[key]=` syntax would probably be a bit cleaner

dbkeitel (changes requested):
Minor comments only. But I'll also give this a quick manual test before approving.

dbkeitel:
Problem:
```
>>> from pyfstat import InjectionParametersGenerator
>>> ipg = InjectionParametersGenerator()
>>> ipg.
ipg.priors                        ipg.return_injection_parameters(  ipg.seed                          
>>> ipg.priors
  File "/home/dkeitel/git/sources/PyFstat/pyfstat/make_sfts.py", line 57, in priors
    return self._priors
AttributeError: 'InjectionParametersGenerator' object has no attribute '_priors'
>>> ipg.seed
  File "/home/dkeitel/git/sources/PyFstat/pyfstat/make_sfts.py", line 53, in seed
    return self._seed
AttributeError: 'InjectionParametersGenerator' object has no attribute '_seed'
>>> ipg.return_injection_parameters()
  File "/home/dkeitel/git/sources/PyFstat/pyfstat/make_sfts.py", line 94, in return_injection_parameters
    for parameter_name, parameter_prior in self._priors.items()
AttributeError: 'InjectionParametersGenerator' object has no attribute '_priors'
```
 * I think `priors` should just not have a `None` default, and be checked to be one of the understood input types before assigning to ` self.priors`.
 * and the `seed` may be `None` (np default behaviour), but should also be stored as `self.seed = seed` before setting `self._rng`.

(Either might need to be modified for the "setter" magic that I'm still only passively familiar with by now.)

pyfstat/make_sfts.py
```diff
@@ -23,6 +24,111 @@ class KeyboardInterruptError(Exception):
...
+                )
+
+        self._priors = current_priors
+
+        logging.info(
+            "Updated parameters: "
+            + " ".join(["{}".format(key) for key in new_priors.keys()])
+        )
+
+    def return_injection_parameters(self):
```
dbkeitel, line 91: For convenience, this could be shortened to just `draw`.

pyfstat/make_sfts.py
```diff
@@ -23,6 +24,111 @@ class KeyboardInterruptError(Exception):
...
+
+    def __init__(self, priors=None, seed=None):
+        """
+        priors: dict
+            Each key refers to one of the signal's parameters (following the PyFstat convention).
+            Priors can be given as values in three formats (by order of evaluation):
+                1) Callable without required arguments
+                    {"ParameterA": np.random.uniform}
+                2) Dict containing numpy.random distribution as key and kwargs in a dict as value
+                    {"ParameterA": {"uniform": {low: 0, high:1}}}
```
dbkeitel, line 41:
```suggestion
                    {"ParameterA": {"uniform": {"low": 0, "high": 1}}}
```

dbkeitel (changes requested):
Some more testing, let's make it a checklist

- [ ] fix default `None`s as per above
- [ ] explicitly check that `priors` is a dict, for a cleaner error message than
```
>>> ipg = InjectionParametersGenerator(priors="x")
  File "/home/dkeitel/git/sources/PyFstat/pyfstat/make_sfts.py", line 49, in __init__
    self.priors = priors
  File "/home/dkeitel/git/sources/PyFstat/pyfstat/make_sfts.py", line 69, in priors
    for parameter_name, parameter_prior in new_priors.items():
AttributeError: 'str' object has no attribute 'items'
````
- [ ] check for duplicate keys in `priors` (test case: `ipg = InjectionParametersGenerator(priors={"ParameterA": np.random.uniform, "ParameterA": np.random.normal})` - it seems the second always wins, but as dicts are generally unordered, I think the behaviour will be essentially undefined)
- [ ] I'm not sure setting the `seed` actually has an effect, or at least I couldn't reproduce any sequences. I'm afraid we need a test for that.

**Comment by Rodrigo-Tenorio** [comment]
> check for duplicate keys in priors (test case: ipg = InjectionParametersGenerator(priors={"ParameterA": np.random.uniform, "ParameterA": np.random.normal}) - it seems the second always wins, but as dicts are generally unordered, I think the behaviour will be essentially undefined)

I'm not sure if that's an actual problem? Python by default allows dictionaries to be defined with duplicated keys, then it takes (apparently) the last one:
```
>>> a = {"A": 1, "A":2}
>>> a
{'A': 2}
```
Btw, python3 preserves dict key ordering!

**Comment by Rodrigo-Tenorio** [comment]
More:
> I think priors should just not have a None default, and be checked to be one of the understood input types before assigning to self.priors.
I'll put an empty dictionary as a default. The problem right now is that `priors` is not set if `None` is given, so there's no attribute `self._prior` (because the setter was not called) and hence the `There's no such a thing as self._priors` error.

>and the seed may be None (np default behaviour), but should also be stored as self.seed = seed before setting self._rng.
Yes. I agree.

**Comment by dbkeitel** [comment]
Regarding dict order: not in python3 in general, but apparently since 3.6 by accident and 3.7+ by design, see [here](https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6). That explains why I didn't catch this improvement when I made the 2.7->3.5 jump. ;)

Regarding `seed`: It seems to work for `InjectionParametersGenerator` but not `AllSkyInjectionParametersGenerator`:
```

```>>> from pyfstat import AllSkyInjectionParametersGenerator
>>> asipg=AllSkyInjectionParametersGenerator(seed=42)
21:43 INFO    : Updated parameters: Alpha Delta
>>> asipg.draw()
{'Alpha': 3.3363807863428123, 'Delta': -0.333397259181978}
>>> asipg=AllSkyInjectionParametersGenerator(seed=42)
21:43 INFO    : Updated parameters: Alpha Delta
>>> asipg.draw()
{'Alpha': 2.4071382784266357, 'Delta': 0.9804089807057452}

**Comment by dbkeitel** [comment]
> Regarding dict order: not in python3 in general, but apparently since 3.6 by accident and 3.7+ by design, see here. That explains why I didn't catch this improvement when I made the 2.7->3.5 jump. ;)

PS: and sanity check feature retracted, of course python doesn't store duplicate keys in the dict to begin with.

**Comment by Rodrigo-Tenorio** [comment]
> Regarding seed: It seems to work for InjectionParametersGenerator but not AllSkyInjectionParametersGenerator:

I think the tests disagree with you. Can you pull and run `pytest tests.py::TestAllSkyInjectionParametersGenerator`?

**Comment by dbkeitel** [comment]
> > Regarding seed: It seems to work for InjectionParametersGenerator but not AllSkyInjectionParametersGenerator:
> 
> I think the tests disagree with you. Can you pull and run `pytest tests.py::TestAllSkyInjectionParametersGenerator`?

That passes, including the `test_rng_seed` which looks like it should test the right thing. Then what's wrong with my snippet above, which still fails?

**Comment by Rodrigo-Tenorio** [comment]
```
>>> asipg=AllSkyInjectionParametersGenerator(seed=42)
22:07 INFO    : Updated parameters: Alpha Delta
22:07 INFO    : Updated parameters:
>>> asipg.draw()
{'Alpha': 2.2841446626171242, 'Delta': 1.1313231406560278}
>>> asipg=AllSkyInjectionParametersGenerator(seed=42
... )
22:07 INFO    : Updated parameters: Alpha Delta
22:07 INFO    : Updated parameters:
>>> asipg.draw()
{'Alpha': 2.6886436313722952, 'Delta': -1.2737975323465385}
```
Good question

File: tests.py [hunk 5]
```diff
@@ -114,6 +114,75 @@ def setUpClass(self):
...
+    def test_rng_seed(self):
+        a_seed = 420
+
+        samples = []
+        for i in range(2):
+            self.InjectionGenerator = self.class_to_test(
+                priors={"ParameterA": {"normal": {"loc": 0, "scale": 1}}}, seed=a_seed
+            )
+            samples.append(self.InjectionGenerator.draw())
+        self.assertTrue(samples[0]["ParameterA"] == samples[1]["ParameterA"])
```
dbkeitel, line 158: When called from `TestAllSkyInjectionParametersGenerator`, this still only checks `ParameterA`, which is actually ok, but Alpha and Delta ignore the seed:
```
>>> from pyfstat import AllSkyInjectionParametersGenerator
>>> asipg=AllSkyInjectionParametersGenerator(priors={"ParameterA": {"normal": {"loc": 0, "scale": 1}}}, seed=42)
22:06 INFO    : Updated parameters: Alpha Delta
22:06 INFO    : Updated parameters: ParameterA
>>> asipg.draw()
{'Alpha': 1.8296262331751116, 'Delta': -0.7278825693693345, 'ParameterA': 0.30471707975443135}
>>> asipg=AllSkyInjectionParametersGenerator(priors={"ParameterA": {"normal": {"loc": 0, "scale": 1}}}, seed=42)
22:06 INFO    : Updated parameters: Alpha Delta
22:06 INFO    : Updated parameters: ParameterA
>>> asipg.draw()
{'Alpha': 5.337303292736909, 'Delta': -1.3868702724508961, 'ParameterA': 0.30471707975443135}
```

pyfstat/make_sfts.py
```diff
@@ -23,6 +24,111 @@ class KeyboardInterruptError(Exception):
...
+
+    def __call__(self):
+        return self.draw()
+
+
+class AllSkyInjectionParametersGenerator(InjectionParametersGenerator):
+    """Like InjectionParametersGenerator, but with hardcoded priors to perform
+    all sky searches. It assumes 1) PyFstat notation and 2) Equatorial coordinates"""
+
+    restricted_priors = {
```
Rodrigo-Tenorio, line 106: Priors are given as numpy functions before setting the seed!

Rodrigo-Tenorio (commented):
Of course it fails

Rodrigo-Tenorio:
And this should be it, but I just realized my prior specification differs from that of PyFstat...

Rodrigo-Tenorio:
Summary of changes after getting rid of setters:

> fix default Nones as per above

I implemented a method to explicitely `set` priors.

> explicitly check that priors is a dict, for a cleaner error message than

Done, it literally spits the docstring out. 

> check for duplicate keys in priors 

Already discussed, that's some Python stuff.

> I'm not sure setting the seed actually has an effect...

It does. Fixed and explicitly tested for generic priors *and sky priors*.

dbkeitel (changes requested):
Last 3 issues:

- [x] behaviour without setting priors:
```
>>> ipg = InjectionParametersGenerator()
>>> ipg.seed
>>> ipg.priors
{}
>>> ipg.draw()
{}
```
As discussed, we'll not make it throw an error here, but leave as useless but reasonable behaviour ("You asked me to draw values for an empty set of parameters, so I give you an empty dict!").

- [x] spurious factor 2 in `Delta` from `AllSkyInjectionParametersGenerator`

- [ ] `AllSkyInjectionParametersGenerator` does not seem to inherit the nice "check this class' docstring" message from `set_priors` when called with e.g. a `str` input (if nontrivial to fix, I'm ok to approve without this)

Rodrigo-Tenorio (commented):
I'd say that finishes it.

dbkeitel (approved):

PR #215, opened 2020-10-30
remove deprecated specialist grid search classes

### LINKED ISSUE #213: Deprecate and then remove various special classes
Hi @ReinhardPrix : Given that @GregoryAshton has almost no time left to maintain any parts of this package and @Rodrigo-Tenorio and I can only effectively maintain the parts we're actively using, we have decided we'd like to remove various specialised classes, all from the grid_based_searches.py module, to make it easier to do some other necessary internal changes. These classes have so far been kept technically alive (tests pass where they exist), but we haven't checked in detail in a long time if they really still do what they're supposed to (and not in all cases do I even really know what they're supposed to do), and some don't have tests at all.

Specifically, these classes are:
 - SliceGridSearch
 - GridUniformPriorSearch
 - SlidingWindow
 - FrequencySlidingWindow
 - EarthTest
 - DMoff_NO_SPIN

I'll do a first PR shortly to _**deprecate**_ them and we'll include that already in an imminent 1.9.0 release. Then I'll make a separate PR for _**removing**_ them and assign that one to you; then it would be great if you can reply, by sometime next week, whether you are still interested in using any subset of them, and we would then keep those and try together to make sure they're correct and properly tested.

dbkeitel:
- using DefunctClass to still throw an informative error
 - closes #213


ReinhardPrix (approved):
no objections from, as I said. I can approve this on this basis, but I'm afraid I can't really 'review' the changes in detail as I'm not familiar enough with pyFstat internals.

dbkeitel:
Thanks and no problem, there isn't much to review here as it really just removes all the classes and puts stubs in their place. They can still be imported but throw the following error when called, a second harder deprecation step instead of a full removal so to say:
```
>>> from pyfstat import FrequencySlidingWindow
>>> fsw=FrequencySlidingWindow()
  File "/home/dkeitel/git/sources/PyFstat/pyfstat/core.py", line 1771, in __new__
    raise NotImplementedError(defunct_message)
NotImplementedError: The FrequencySlidingWindow class is no longer included in PyFstat! Last supported version was 1.9.0. Pull requests to reinstate the class are welcome.
```

PR #216, opened 2020-10-30
tests.py: remove spurious file copy

dbkeitel:
- finally I figured out why I always had some random SFTs in my root directory...
 - leftover from some debugging obviously


Rodrigo-Tenorio (approved):
Same

PR #220, opened 2020-12-04
Add example for tracks

Rodrigo-Tenorio:
Let's have a common place to generate spectrogram plots containing CW signals, shall we?


examples/other_examples/spectrogram_example.py
```diff
@@ -0,0 +1,60 @@
...
+import matplotlib.pyplot as plt
+
+import pyfstat
+
+plt.rcParams["font.family"] = "serif"
+plt.rcParams["font.size"] = 18
+plt.rcParams["text.usetex"] = True
+
+label = "track_example"
+outdir = os.path.join(sys.path[0], "tmp")
```
dbkeitel, line 14:
```suggestion
```

examples/other_examples/spectrogram_example.py
```diff
@@ -0,0 +1,60 @@
...
+    "h0": data_parameters["sqrtSX"] / depth,
+    "cosi": 1.0,
+}
+
+data = pyfstat.BinaryModulatedWriter(
+    label=label, outdir=outdir, **data_parameters, **signal_parameters
+)
+data.make_data()
+
+times, freqs, sft_data = pyfstat.helper_functions.get_sft_array(data.sftfilepath)
```
dbkeitel, line 45:
```suggestion
with tempfile.TemporaryDirectory() as tmpdir:
    data = pyfstat.BinaryModulatedWriter(
        label=label, outdir=tmpdir, **data_parameters, **signal_parameters
    )
   data.make_data()

   times, freqs, sft_data = pyfstat.helper_functions.get_sft_array(data.sftfilepath)
```

examples/other_examples/spectrogram_example.py
```diff
@@ -0,0 +1,60 @@
...
+    (times - times[0]) / 86400,
+    freqs,
+    (sft_data / data_parameters["sqrtSX"])[:-1, :-1],
+    cmap="inferno_r",
+)
+fig.colorbar(c, label="Normalized Power")
+plt.tight_layout()
+fig.savefig("spectrogram_example.png")
+
+shutil.rmtree(outdir)
```
dbkeitel, line 60:
```suggestion
```

examples/other_examples/spectrogram_example.py
```diff
@@ -0,0 +1,60 @@
...
+
+times, freqs, sft_data = pyfstat.helper_functions.get_sft_array(data.sftfilepath)
+
+fig, ax = plt.subplots(figsize=(0.8 * 16, 0.8 * 9))
+ax.grid(which="both")
+ax.set(xlabel="Time [days]", ylabel="Frequency [Hz]", ylim=(99.98, 100.02))
+c = ax.pcolormesh(
+    (times - times[0]) / 86400,
+    freqs,
+    (sft_data / data_parameters["sqrtSX"])[:-1, :-1],
```
dbkeitel, line 53: It doesn't really matter for an illustrative example plot, but I think _technically_ here you're dropping one data point (SFT) and the "proper" solution would be to add end points to the `times` and `freqs` instead...?
(Just recently ran into this annoying "n+1,n+1,n*n" convention of `pcolormesh` elsewhere...)

pyfstat/helper_functions.py
```diff
@@ -357,7 +357,7 @@ def get_sft_array(sftfilepattern, data_duration=None, F0=None, dF0=None):
         )
 
     SFTs = MultiSFTs.data[0]
-    times = [sft.epoch.gpsSeconds for sft in SFTs.data]
+    times = np.array([sft.epoch.gpsSeconds for sft in SFTs.data])
```
dbkeitel, line 360: Good catch, it really made no sense to have `data` numpyfied but `times` not.

tests.py
```diff
@@ -555,8 +555,8 @@ def test_glitch_injection(self):
         max_freq_noglitch = freqs_noglitch[np.argmax(data_noglitch, axis=0)]
         max_freq_glitch = freqs_glitch[np.argmax(data_glitch, axis=0)]
         print([max_freq_vanilla, max_freq_noglitch, max_freq_glitch])
-        self.assertEqual(times_noglitch, times_vanilla)
-        self.assertEqual(times_glitch, times_vanilla)
+        self.assertTrue(np.allclose(times_noglitch, times_vanilla))
+        self.assertTrue(np.allclose(times_glitch, times_vanilla))
```
dbkeitel, line 559: Our times still are integers, aren't they? So shouldn't we still check for exact equality? I guess you could either cast it back to a list for `assertEqual` to digest, or there's also an `numpy.testing.assert_array_equal`.

examples/other_examples/spectrogram_example.py
Rodrigo-Tenorio: Yes. I'll think about this.

tests.py
Rodrigo-Tenorio: That's true. I'll look into `numpy.testing.assert_array_equal`

examples/other_examples/spectrogram_example.py
```diff
@@ -0,0 +1,60 @@
+import os
+import shutil
```
Rodrigo-Tenorio, line 2:
```suggestion
```

examples/other_examples/spectrogram_example.py
Rodrigo-Tenorio: `shading="nearest"` does exactly what I intended to do!

tests.py
Rodrigo-Tenorio: Why not just `np.all(a == b)`, since `a` and `b` are integer arrays.

Rodrigo-Tenorio: a65eafb

dbkeitel (approved):

Issue #221, opened 2020-12-29
html documentation via sphinx/autodoc & readthedocs


dbkeitel:
I've had a look at using [sphinx/autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) and [readthedocs](https://readthedocs.org/) to generate and host html documentation pages, and at least the generation part seems straightforward enough.

- [x] basic sphinx setup --> #222
- [x] publish on readthedocs
- [x] go through warnings and obvious issues
- [x] include examples --> #226 
- [x] extend and improve documentation where lacking

dbkeitel:
--> https://pyfstat.readthedocs.io/en/latest/

dbkeitel:
So far I've taken care of cleaning up and extending

- core.py: #232
- make_sfts.py: #234 
- grid_based_searches.py: #238
- helper_functions.py: #241
- tcw_fstat_map_functions.py: #242

That just leaves mcmc_based_searches.py, arguably our most important submodule... @Rodrigo-Tenorio would you be willing to take that one on? Notes:
- Don't worry too much about whitespace/newlines formatting in the doc strings. Apparently the current `black 20.8b1` does not have opinions on them, but the next release will.
- If you find any undocumented (or very sparsely documented) class methods which aren't used directly yet anywhere else (including examples) and don't have obvious end-user appeal, feel free to change them to `_` internals.

dbkeitel:
It should be good enough for now.

PR #222, opened 2020-12-29
Sphinx

dbkeitel:
See #221. This first PR only adds a very basic sphinx setup.

-  The basic set of files was generated with `sphinx-quickstart`.
-  Then the `.rst` files (which yield the actual contents) generated with `sphinx-apidoc -f -o source ../ ../versioneer.py ../setup.py` (the last arguments after the first "../" are exclude patterns); this only needs to be redone if completely new module files are added.
-  Further top-level documentation can be manually added to `docs/source/index.rst`.
-  The pages can be generated with `make html` from within the `docs` directory.

I guess I'll get this merged first as-is (or with minimal further fixes) and then try to set up auto-build and publication actions before tackling the remaining tasks from #221.

Example output attached:
[pyfstat-docs-dca1ed63d71b7016fd26a719676418833fdb7d4a.tar.gz](https://github.com/PyFstat/PyFstat/files/5751540/pyfstat-docs-dca1ed63d71b7016fd26a719676418833fdb7d4a.tar.gz)


dbkeitel:
This seems to work already, getting it merged to do further improvements later.

PR #223, opened 2020-12-29
Include examples in sphinx html documentation

dbkeitel:
See #221. Sphinx doesn't play well with executable scripts out-of-the-box, so here are two possible hacky solutions. Both will require a simple helper script to auto-generate the necessary `.rst` files so they can be easily updated as the examples suite evolves. I think solution 2 is better.

Either way, it requires properly restructuring all examples with `if __name__ == "__main__":` to be import-safe, and some adding of help strings describing what the example does (a good idea anyway). So this will be a "draft" for now; it also should wait until after #171 is merged to avoid conflicts.

Example output: 
[pyfstat-docs-7cda9931241c6a682a4a1ba48f789e6ba342597c.tar.gz](https://github.com/PyFstat/PyFstat/files/5751557/pyfstat-docs-7cda9931241c6a682a4a1ba48f789e6ba342597c.tar.gz)

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,13 +178,11 @@ For a general introduction to installing modules, see
   and [black](https://black.readthedocs.io) style checker
   locally,
   as these checks are required to pass by the online integration pipeline.
-* Some optional plotting methods depend on two additional packages,
-  and some of the [examples](./examples) require these to run:
- [gridcorner](https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner) 
-  and [chainconsumer](https://github.com/Samreay/ChainConsumer);
-for `pip` users this is most conveniently installed by
+* Some optional plotting methods depend on the additional package
+  [chainconsumer](https://github.com/Samreay/ChainConsumer)
+  and some of the [examples](./examples) require this to run.
+  Gor `pip` users, this is most conveniently installed by
 ```
-pip install git+https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner
 pip install chainconsumer
 ```
 * If you prefer to make your own LALSuite installation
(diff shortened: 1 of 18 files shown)


dbkeitel:
It can even auto-build from PRs, nice: https://pyfstat--223.org.readthedocs.build/en/223/examples.html

@Rodrigo-Tenorio opinions on what is a useful format for the examples?

Rodrigo-Tenorio:
> opinions on

I would tend to reuse as much code as possible.

A first idea would be to clean up "redundant" examples, meaning we put everything (`Writer`, `MCMCSearch` and the like) before
the `if __name__ == "__main__"` and import them to the rest of examples whenever possible. I think something like this is already done for the glitching examples...?

Folding that a bit more, we could create a dedicated module in each example folder and import everything from there, but this somehow ignores the initial purpose.

Finally, we could re-work examples into some sort of jupyter notebook and use [binder](https://mybinder.org/)? I like this approach, it allows one to tweak with the examples in a clean environment.

dbkeitel:
> EDIT: Looks like binder indeed plays well with sphinx: https://sphinx-gallery.github.io/stable/index.html

That sounds really nice. May require restructuring of the examples directory, but the only showstopper would be if it _requires_ any of the fancy rst-markup inside the files themselves - I'd like to keep them maximally readbable as plain python files too. But from a first look through the docs, it sounds like everything besides an extra top-level file is optional.

dbkeitel:
Now fails with the missing gridcorner dependency. I guess I could learn how to install-from-gitlab in a requirements.txt file, but since we had discussed the issue before, we should just include the package's functionality directly. --> #225

Rodrigo-Tenorio:
I think `sphinx-gallery` already covers the issue tackled in this PR (see #226 whenever #228 gets solved)?

dbkeitel:
Dropped in favour of #226.

PR #224, opened 2020-12-29
add .readthedocs.yml to allow using both requirements.txt files

dbkeitel:
The [first readthedocs builds off master](https://readthedocs.org/projects/pyfstat/builds/12655554/) actually didn't include the autodoc-generated contents as they failed to import dependencies. The web interface only allows to load in one `requirements.txt` file; let's see if through a yml we can use both the one in `docs` and the root-level one of PyFstat itself.

PS: as per https://docs.readthedocs.io/en/stable/config-file/v2.html#requirements-file


dbkeitel:
Problem solved.

PR #227, opened 2020-12-29
add gridcorner code 

### LINKED ISSUE #225: include gridcorner
Some of the examples use @GregoryAshton's [gridcorner package](https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner) which is not released on pypi nor conda. As discussed a while ago with him, it would be best to directly include (copy&paste) this functionality into pyfstat.

dbkeitel:
taken from https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner with @GregoryAshton's permission

 - based on code from https://github.com/dfm/corner.py
 - closes #225

@GregoryAshton do you think this is the right way to include the corner.py copyright licence? (Edit: full-text paste in the top-level docstring of the gridcorner.py file.) It seems compatible with our MIT licence, but probably safer to copy it over explicitly like this, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,11 +269,14 @@ Maintainers:
 * Greg Ashton
 * David Keitel
 
-Other contributors:
+Active contributors:
 * Reinhard Prix
 * Rodrigo Tenorio
+
+Other contributors:
 * Karl Wette
 * Sylvia Zhu
+* Dan Foreman-Mackey (`pyfstat.gridcorner` is based on DFM's [corner.py](https://github.com/dfm/corner.py))
 
 
 ## Citing this work
--- a/README.md
+++ b/README.md
@@ -178,13 +178,11 @@ For a general introduction to installing modules, see
   and [black](https://black.readthedocs.io) style checker
   locally,
   as these checks are required to pass by the online integration pipeline.
-* Some optional plotting methods depend on two additional packages,
-  and some of the [examples](./examples) require these to run:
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -178,13 +178,11 @@ For a general introduction to installing modules, see
...
   as these checks are required to pass by the online integration pipeline.
-* Some optional plotting methods depend on two additional packages,
-  and some of the [examples](./examples) require these to run:
- [gridcorner](https://gitlab.aei.uni-hannover.de/GregAshton/gridcorner) 
-  and [chainconsumer](https://github.com/Samreay/ChainConsumer);
-for `pip` users this is most conveniently installed by
+* Some optional plotting methods depend on the additional package
+  [chainconsumer](https://github.com/Samreay/ChainConsumer)
+  and some of the [examples](./examples) require this to run.
+  Gor `pip` users, this is most conveniently installed by
```
Rodrigo-Tenorio, line 184:
```suggestion
  For `pip` users, this is most conveniently installed by
```

GregoryAshton:
Yes, I agree this seems the most sensible way to do it. Do you think we should email Dan to double check?

dbkeitel:
> Yes, I agree this seems the most sensible way to do it. Do you think we should email Dan to double check?

It would probably be a fair gesture. Do you remember in more detail which (if any) pieces of the code are direct copies?

dbkeitel:
Got confirmation from @dfm by e-mail about the corner.py license terms, so this can now be merged. Thanks!

I've now also updated the "contributors" section in the README.

PR #230, opened 2020-12-30
sphinx: autoclass_content = 'both'

dbkeitel:
- includes documentation of each __init__() method into the documentation of its class


dbkeitel:
I apparently forgot to merge this, but also had the same commit already in !232 so this is no longer needed.

PR #235, opened 2020-12-30
Writer.run_makefakedata(): remove dysfunctional code for removing old files

dbkeitel:
- It seems this wasn't actually doing anything,
   probably because unlink doesn't understand * wildcards or something.
 - It was contrary to the whole check-if-safe-to-reuse idea
   the code actually uses.

@Rodrigo-Tenorio quick sanity check please

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -753,12 +753,6 @@ class Writer(BaseSearchClass):
         and else runs the actual generation code.
         """
 
-        # Remove old data:
-        try:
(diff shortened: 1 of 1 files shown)


Rodrigo-Tenorio (approved):
Never seen this particular exception rise.

dbkeitel:
> Never seen this particular exception rise.

Well the trick was that it's excepted with `pass`, so doesn't say anything. But I put in some debug output and it's clear that it was not doing anything even if the files existed. Let's just get rid of it.

PR #239, opened 2020-12-31
helper_functions: remove compute_P_twoFstarcheck() and compute_pstar()

dbkeitel:
-  These were introduced in 975f1f5c4494036c3db3ec39fb4e0dfe6b9f8628
    in the context of the `DMoff` classes which have been removed
    due to being unmaintained.


dbkeitel:
cc @GregoryAshton : I'm removing these for now as I don't understand them enough to document or maintain - but of course we'd always be happy for you to come out of retirement and revive any of those parts I'm killing. ;)

PR #243, opened 2021-01-04
mcmc_based_searches docs

Rodrigo-Tenorio:
Improve docs of `mcmc_based_searches.py` module.


pyfstat/mcmc_based_searches.py
```diff
@@ -1,5 +1,4 @@
-""" Searches using MCMC-based methods """
-
+""" PyFstat search & follow-up classes using MCMC-based methods """
```
dbkeitel, line 1: Let's already but it here centrally:
```suggestion
"""PyFstat search & follow-up classes using MCMC-based methods

The general approach is described in
Ashton & Prix (PRD 97, 103020, 2018):
https://arxiv.org/abs/1802.05450
and we use the `ptemcee` sampler
described in Vousden et al. (MNRAS 455, 1919-1937, 2016):
https://arxiv.org/abs/1501.05823
and based on Foreman-Mackey et al. (PASP 125, 306, 2013):
https://arxiv.org/abs/1202.3665
"""
```

pyfstat/mcmc_based_searches.py
```diff
@@ -2728,62 +2626,15 @@ def logl(self, theta, search):
...
-        over the SFT time span as passed to CreateFstatInput
-    injectSources: dict, optional
-        If given, inject these properties into the SFT files before running
-        the search
-    assumeSqrtSX: float, optional
-        Don't estimate noise-floors, but assume (stationary) per-IFO sqrt{SX}
+    Executes MCMC runs with increasing coherence times in order to follow up a parameter space
+    region. The main idea is to use an MCMC run to identify an interesting parameter space region
+    to then zoom-in said region using a finer "effective resolution" by increasing the coherence time.
+    See G. Ashton, R. Prix (2018) arXiv:1802.05450 [astro-ph.IM].
```
dbkeitel, line 2634: For consistency with the citation style I've used elsewhere:
```suggestion
    See Ashton & Prix (PRD 97, 103020, 2018): https://arxiv.org/abs/1802.05450
```
(There's probably a proper way to maintain a reference library, but that can wait.)

pyfstat/mcmc_based_searches.py
```diff
@@ -22,72 +21,11 @@
...
-    transientWindowType: str
-        If 'rect' or 'exp',
-        compute atoms so that a transient (t0,tau) map can later be computed.
-        ('none' instead of None explicitly calls the transient-window function,
-        but with the full range, for debugging)
-        Currently only supported for nsegs=1.
-    tCWFstatMapVersion: str
-        Choose between standard 'lal' implementation,
-        'pycuda' for gpu, and some others for devel/debug.
+    Evalutates the coherent F-statistic across a parameter space region
```
dbkeitel, line 27:
```suggestion
    Evaluates the coherent F-statistic across a parameter space region
```

pyfstat/mcmc_based_searches.py
```diff
@@ -22,72 +21,11 @@
...
+    corresponding to an isolated/binary-modulated CW signal.
 
     Attributes
```
dbkeitel, line 30: These show up twice in the documentation - once these comments near the top of the class, then additional un-documented blocks below the init parameters. I think it'll look cleaner if this section is omitted and the explanations go into `"""text"""` immediately after each attribute instead.

pyfstat/mcmc_based_searches.py
```diff
@@ -589,23 +486,28 @@ def run(
...
-            The proposal scale factor used by the sampler, see Goodman & Weare
-            (2010). If the acceptance fraction is too low, you can raise it by
-            decreasing the a parameter; and if it is too high, you can reduce
-            it by increasing the a parameter [Foreman-Mackay (2013)].
+            The proposal scale factor `a > 1` used by the sampler, see
+            [Goodman & Weare (2010) 10.2140/camcos.2010.5.65].
+            The bigger the value, the wider the range to draw proposals from.
+            If the acceptance fraction is too low, you can raise it by
+            decreasing the `a` parameter; and if it is too high, you can reduce
+            it by increasing the `a` parameter [Foreman-Mackay arXiv:1202.3665 [astro-ph.IM] (2013)].
```
dbkeitel, line 494: should also harmonize reference format here

pyfstat/mcmc_based_searches.py
```diff
@@ -1151,7 +1049,22 @@ def plot_prior_posterior(
...
+           `norm`, `halfnorm`, `neghalfnorm` or `lognorm` priors are given, otherwise ignored.
+        injection_parameters: dict
+            Dictionary containing the parameters of a signal. All parameters being searched must be
+            present as dictionary keys, otherwise this option is ignored.
+        fig_and_axes: tuple
+            (fig, axes) tuple to plot on.
+        save_fig: bool
+            If true, save the figure, else return the fig, axes.
+
+        """
```
dbkeitel, line 1067: missing `Returns`

pyfstat/mcmc_based_searches.py
```diff
@@ -1664,7 +1577,7 @@ def _pickle_data(self, samples, lnprobs, lnlikes, all_lnlikelihood):
             pickle.dump(d, File)
 
     def get_saved_data_dictionary(self):
-        """ Returns dictionary of the data saved in the pickle """
+        """ Returns dictionary of the data saved in the pickle `self.pickle_path`. """
```
dbkeitel, line 1580: missing `Returns`

pyfstat/mcmc_based_searches.py
```diff
@@ -1767,14 +1683,8 @@ def _get_twoF_from_loglikelihood(self, idx=None):
...
-    def get_max_twoF(self, threshold=0.05):
-        """Returns the max likelihood sample and the corresponding 2F value
-
-        Note: the sample is returned as a dictionary along with an estimate of
-        the standard deviation calculated from the std of all samples with a
-        twoF within `threshold` (relative) to the max twoF
-
-        """
+    def get_max_twoF(self):
+        """Returns the max likelihood sample and the corresponding 2F value"""
```
dbkeitel, line 1687: missing explicit `Returns`

pyfstat/mcmc_based_searches.py
```diff
@@ -1815,7 +1725,11 @@ def get_max_twoF(self, threshold=0.05):
         return d, maxtwoF
 
     def get_summary_stats(self):
-        """ Returns a dict of point estimates for all production samples """
+        """Returns a dict of point estimates for all production samples.
+
+        Point estimates are computed on the MCMC samples using `numpy.mean`,
+        `numpy.std` and `numpy.quantiles` with q=[0.005, 0.05, 0.25, 0.5, 0.75, 0.95, 0.995].
+        """
```
dbkeitel, line 1732: again better to use explicit `Returns" label

pyfstat/mcmc_based_searches.py
```diff
@@ -2084,8 +2004,16 @@ def _p_val_twoFhat(self, twoFhat, ntrials, twoFhatmax=500, Npoints=1000):
         pdf = self._pdf_twoFhat(twoFhats, self.nglitch, ntrials)
         return np.trapz(pdf, twoFhats)
 
-    def get_p_value(self, delta_F0, time_trials=0):
-        """ Get's the p-value for the maximum twoFhat value """
+    def get_p_value(self, delta_F0=0, time_trials=0):
+        """Get's the p-value for the maximum twoFhat value assuming Gaussian noise
```
dbkeitel, line 2008:
```suggestion
        """Gets the p-value for the maximum twoFhat value assuming Gaussian noise
```

(It should ideally also be clearly documented what this actually means, but can be done another time.)

pyfstat/mcmc_based_searches.py
```diff
@@ -2158,6 +2086,13 @@ def compute_evidence(self, make_plots=False, write_to_file=None):
 
     @staticmethod
     def read_evidence_file_to_dict(evidence_file_name="Evidences.txt"):
+        """Read evidence file and put it into an OrderedDict
+
+        Parameters
+        ----------
+        evidence_file_name: str
+            Filename to read.
+        """
```
dbkeitel, line 2095: missing `Returns`

dbkeitel: and both missing `Parameters` and `Returns` on `compute_evidence()` above this one

pyfstat/mcmc_based_searches.py
```diff
@@ -2177,19 +2121,6 @@ class MCMCGlitchSearch(MCMCSearch):
...
-        The minimum duration (in seconds) of a segment between two glitches
-        or a glitch and the start/end of the data
-    theta0_idx, int
-        Index (zero-based) of which segment the theta refers to - useful
-        if providing a tight prior on theta to allow the signal to jump
-        too theta (and not just from)
-
     """
 
     symbol_dictionary = dict(
```
dbkeitel, line 2126: can copy docstrings for all these from parent class and modify if necessary

pyfstat/mcmc_based_searches.py
```diff
@@ -2728,62 +2626,15 @@ def logl(self, theta, search):
...
-        Don't estimate noise-floors, but assume (stationary) per-IFO sqrt{SX}
+    Executes MCMC runs with increasing coherence times in order to follow up a parameter space
+    region. The main idea is to use an MCMC run to identify an interesting parameter space region
+    to then zoom-in said region using a finer "effective resolution" by increasing the coherence time.
+    See G. Ashton, R. Prix (2018) arXiv:1802.05450 [astro-ph.IM].
+
+    See MCMCSemiCoherentSearch for a list of additional parameters, here we list only the additional
+    init parameters of this class.
 
     Attributes
```
dbkeitel, line 2639: can be dropped here I think?

pyfstat/mcmc_based_searches.py
```diff
@@ -2949,7 +2912,22 @@ def init_run_setup(
         log_table=True,
         gen_tex_table=True,
     ):
-
+        """
```
dbkeitel, line 2915: not properly formatted: missing opening sentence, `Parameters` and `Returns`

pyfstat/mcmc_based_searches.py
```diff
@@ -3100,181 +3113,6 @@ def _get_p0_per_stage(self, reset_p0=False):
...
-                    )
-                    plt.close(walkers_fig)
-                except Exception as e:
-                    logging.warning(
-                        "Failed to save walker plots due to Error {}".format(e)
-                    )
-
 
 class MCMCTransientSearch(MCMCSearch):
     """MCMC search for a transient signal using ComputeFstat
```
dbkeitel, line 3118: this has all the parameters duplicated from basic `MCMCSearch`, and the attributes are missing documentation again

dbkeitel (changes requested):
Looks good overall. Long-term we should make it a lot more verbose, but for now I only put in comments about inconsistent formatting, references etc.

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: Done in d847f5e, how do you like it? @dbkeitel

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: I haven't found any arXiv link, are you Ok keeping the doi? 93217b9c5198507488c3a4db05d02f996b1a2ce2

(I still have to update the Formean-Mackay one)

dbkeitel: :+1:

pyfstat/mcmc_based_searches.py
dbkeitel: Yes, that looks better! :+1:

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: I like to make clear that this particular follow-up approach is documented in detail in a paper.

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: That's odd... I'm not sure why it happens in this class but not the other ones.

Rodrigo-Tenorio:
Question (just to be clear): Do we want to repeat the documentation of inherited arguments in child classes or not?
It seems like both options can be done (e.g. see `MCMCGlitchSearch` and `MCMCTransientSearch`), which one do you prefer?

I'd tend to say not to repeat them, since child classes have an hyperlink to parent class, but I'm open to the other option as well.

pyfstat/mcmc_based_searches.py
dbkeitel: That comment was supposed to point to the "attributes", which are the same as above.

pyfstat/mcmc_based_searches.py
dbkeitel: Ah, because it doesn't have an overriden `__init__()` and hence inherits it from the main `MCMCSearch` class, full documentation included. That's good then, no duplicate text to maintain!

dbkeitel: ...then the two lines below about "See parent" can be removed, though.

dbkeitel:
> Question: Do we want to repeat the documentation of inherited arguments in child classes or not?

In principle it's nicer to explicitly repeat them, but where we'd need to do it manually, it'll add maintenance burden.

> It seems like both options can be done (e.g. see `MCMCGlitchSearch` and `MCMCTransientSearch`), which one do you prefer?

As commented above, the difference is that one overrides the `__init__()` and the other inherits it.

> I'd tend to say not to repeat them, since child classes have an hyperlink to parent class, but I'm open to the other option as well.

I agree. If a child fully inherits from its parents, that's a lucky case and we leave it as is; but if `__init__()` is overridden and hence we'd need to manually duplicate the parameters documentation, just leave a comment like "See parent".

pyfstat/mcmc_based_searches.py
```diff
@@ -116,6 +50,11 @@ class MCMCSearch(BaseSearchClass):
         tp=r"tp",
         argp=r"argp",
     )
+    """
+    symbol_dictionary
```
dbkeitel, line 54: This and the others don't need their name repeated:
Screenshot_20210107_130923

pyfstat/mcmc_based_searches.py
```diff
@@ -128,7 +67,18 @@ class MCMCSearch(BaseSearchClass):
         tp="",
         argp="",
     )
+    """
+    unit_dictionary
```
dbkeitel, line 71:
```suggestion
```

pyfstat/mcmc_based_searches.py
```diff
@@ -128,7 +67,18 @@ class MCMCSearch(BaseSearchClass):
...
+        Key, val pairs of the parameters (i.e. `F0`, `F1`), and the
+        units (i.e. `Hz`)
+    """
     transform_dictionary = {}
+    """
+    transform_dictionary
```
dbkeitel, line 77:
```suggestion
```

pyfstat/mcmc_based_searches.py
```diff
@@ -2199,11 +2196,21 @@ class MCMCGlitchSearch(MCMCSearch):
         Alpha=r"$\alpha$",
         Delta=r"$\delta$",
     )
+    """
+    symbol_dictionary
```
dbkeitel, line 2200:
```suggestion
```

pyfstat/mcmc_based_searches.py
```diff
@@ -2199,11 +2196,21 @@ class MCMCGlitchSearch(MCMCSearch):
...
+        Key, val pairs of the parameters (`F0`, `F1`, ...), to LaTeX math
+        symbols for plots
+    """
     glitch_symbol_dictionary = dict(
         delta_F0=r"$\delta f$",
         delta_F1=r"$\delta \dot{f}$",
         tglitch=r"$t_\mathrm{glitch}$",
     )
+    """
+    glitch_symbol_dictionary
```
dbkeitel, line 2210:
```suggestion
```

pyfstat/mcmc_based_searches.py
```diff
@@ -2215,6 +2222,11 @@ class MCMCGlitchSearch(MCMCSearch):
         delta_F1=r"Hz/s",
         tglitch=r"s",
     )
+    """
+    unit_dictionary
```
dbkeitel, line 2226:
```suggestion
```

pyfstat/mcmc_based_searches.py
```diff
@@ -2223,6 +2235,12 @@ class MCMCGlitchSearch(MCMCSearch):
             "label": r"$t^{g}_0$ \n [d]",
         }
     )
+    """
+    transform_dictionary
```
dbkeitel, line 2239:
```suggestion
```

pyfstat/mcmc_based_searches.py
```diff
@@ -3293,6 +3207,11 @@ class MCMCTransientSearch(MCMCSearch):
         transient_tstart=r"$t_\mathrm{start}$",
         transient_duration=r"$\Delta T$",
     )
+    """
+    symbol_dictionary
```
dbkeitel, line 3211:
```suggestion
```

pyfstat/mcmc_based_searches.py
```diff
@@ -3302,7 +3221,11 @@ class MCMCTransientSearch(MCMCSearch):
         transient_tstart=r"s",
         transient_duration=r"s",
     )
-
+    """
+    unit_dictionary
```
dbkeitel, line 3225:
```suggestion
```

pyfstat/mcmc_based_searches.py
```diff
@@ -3316,6 +3239,12 @@ class MCMCTransientSearch(MCMCSearch):
             "label": "Transient start-time \n days after minStartTime",
         },
     )
+    """
+    transform_dictionary
```
dbkeitel, line 3243:
```suggestion
```

dbkeitel (changes requested):

dbkeitel (approved):

PR #226, opened 2020-12-29
Sphinx binder

Rodrigo-Tenorio:
Playing on top of #223  to include a gallery of examples and `binder` support, if possible.

PR #245, opened 2021-01-06
BSGL readiness and internal cleanup

dbkeitel:
Here's another big messy PR changing lots of internals, with the main goal to fix the BSGL implementation, but along the way cleaning up the way the likelihood and detection statistic functions are called and outputs are stored. It also adds some basic test cases and examples for BSGL usage in both grid and MCMC classes, resolving #99.

Some highlights from the commits:
---
- fixed basic twoFX and BSGL computation logic for transient and semicoherent cases, generalized to Ndet>2
- split up the various detection statistic methods into generic and specific ones
- introduce generic `_get_detstat_from_loglikelihood()`
- fix missing logic for BSGL case in various places
- for MCMC classes:
  - clarify likelihood-detstat relation, `_set_likelihoodcoef()` now also takes care of twoF->F and of BSGL case
  - new `_set_point_for_evaluation` method to centrally merge fixed and samples parameters
  - in BSGL case, export both twoF and BSGL, but not yet twoFX
- for grid classes:
  - in BSGL case, export all of twoF, twoFX and BSGL
  - now call `_initiate_search_object()` from `__init__()` instead of from `run()`
- for transients: explicitly call the standard detection statistic `maxTwoF`
- new test cases and examples
- lots of documentation and cleanup changes

_Not_ done in this PR:
---
- enable BSGL for glitch search classes; put in an explicit error message instead
- adding explicit user options for the BSGL tuning parameters (note that MCMC classes already have a `rhohatmax` option which could be reused, as that happens to be an equivalent parameter to BSGL's `Fstar0`)
- adding strict quantitative tests
- changing the choices for how "transient BSGL" is done, which may not be the best currently
- finishing up the implementation and output of single-IFO F-stats on multi-IFO searches - currently these are only computed when `BSGL` option is set, and only exported from grid- not from MCMC searches. A followup should compute and export them always when Ndet>=2 (as they're essentially for free).
- further work on transient detection statistic choices - we should also wrap the PGM11 transient marginalized Bayes factor

Review / status:
---
@GregoryAshton @Rodrigo-Tenorio this will have to wait to be rebased on top of #226 #227 #243 anyway, but comments are welcome already. As announced, the diff is messy though. @ReinhardPrix several of the changes here should have a lot of nostalgic value for you too. ;)

### DIFF
--- a/examples/transient_examples/PyFstat_example_short_transient_grid_search.py
+++ b/examples/transient_examples/PyFstat_example_short_transient_grid_search.py
@@ -32,9 +32,11 @@ if __name__ == "__main__":
     Alphas = [data.Alpha]
     Deltas = [data.Delta]
 
+    BSGL = False
+
     print("Standard CW search:")
     search1 = pyfstat.GridSearch(
-        label="CW",
+        label="CW" + ("-BSGL" if BSGL else ""),
         outdir=data.outdir,
         sftfilepattern=os.path.join(data.outdir, "*simulated_transient_signal*sft"),
         F0s=F0s,
@@ -43,16 +45,15 @@ if __name__ == "__main__":
         Alphas=Alphas,
         Deltas=Deltas,
         tref=data.tref,
-        BSGL=False,
+        BSGL=BSGL,
     )
     search1.run()
     search1.print_max_twoF()
-
     search1.plot_1D(xkey="F0", xlabel="freq [Hz]", ylabel="$2\\mathcal{F}$")
 
     print("with t0,tau bands:")
     search2 = pyfstat.TransientGridSearch(
-        label="tCW",
+        label="tCW" + ("-BSGL" if BSGL else ""),
         outdir=data.outdir,
         sftfilepattern=os.path.join(data.outdir, "*simulated_transient_signal*sft"),
         F0s=F0s,
@@ -64,11 +65,10 @@ if __name__ == "__main__":
         transientWindowType="rect",
         t0Band=data.duration - 2 * data.Tsft,
         tauBand=data.duration,
-        BSGL=False,
         outputTransientFstatMap=True,
         tCWFstatMapVersion="lal",
+        BSGL=BSGL,
     )
     search2.run()
     search2.print_max_twoF()
-
     search2.plot_1D(xkey="F0", xlabel="freq [Hz]", ylabel="$2\\mathcal{F}$")
--- a/examples/transient_examples/PyFstat_example_short_transient_grid_search.py
+++ b/examples/transient_examples/PyFstat_example_short_transient_grid_search.py
@@ -36,7 +36,7 @@ if __name__ == "__main__":
 
     print("Standard CW search:")
     search1 = pyfstat.GridSearch(
-        label="CW" + ("-BSGL" if BSGL else ""),
+        label="CW" + ("_BSGL" if BSGL else ""),
         outdir=data.outdir,
         sftfilepattern=os.path.join(data.outdir, "*simulated_transient_signal*sft"),
         F0s=F0s,
@@ -53,7 +53,7 @@ if __name__ == "__main__":
 
     print("with t0,tau bands:")
     search2 = pyfstat.TransientGridSearch(
-        label="tCW" + ("-BSGL" if BSGL else ""),
+        label="tCW" + ("_BSGL" if BSGL else ""),
         outdir=data.outdir,
         sftfilepattern=os.path.join(data.outdir, "*simulated_transient_signal*sft"),
         F0s=F0s,
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1623,7 +1623,6 @@ class SemiCoherentSearch(ComputeFstat):
             to obtain the per-segment coherence time `Tcoh`.
         """
 
-        self.fs_file_name = os.path.join(self.outdir, self.label + "_FS.dat")
         self.set_ephemeris_files(earth_ephem, sun_ephem)
         self.transientWindowType = None  # will use semicoherentWindowRange instead
         self.computeAtoms = True  # for semicoh 2F from ComputeTransientFstatMap()
@@ -2080,7 +2079,6 @@ class SemiCoherentGlitchSearch(SearchForSignalWithJumps, ComputeFstat):
             signal to jump to theta (and not just from).
         """
 
-        self.fs_file_name = os.path.join(self.outdir, self.label + "_FS.dat")
         self.set_ephemeris_files(earth_ephem, sun_ephem)
         self.transientWindowType = "rect"
         self.t0Band = None
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1157,14 +1157,10 @@ class ComputeFstat(BaseSearchClass):
             self.FstatResults.multiFatoms[0],  # 0 index: single frequency bin
             self.windowRange,
         )
-        if self.tCWFstatMapVersion == "lal":
-            F_mn = self.FstatMap.F_mn.data
-        else:
-            F_mn = self.FstatMap.F_mn
 
         # Now instead of the overall twoF,
         # we get the maximum twoF over the transient window range.
-        self.maxTwoF = 2 * np.max(F_mn)
+        self.maxTwoF = 2 * self.FstatMap.maxF
         if np.isnan(self.maxTwoF):
             self.maxTwoF = 0
         return self.maxTwoF
@@ -1186,16 +1182,14 @@ class ComputeFstat(BaseSearchClass):
             at the input parameter values.
             Also stored as `self.log10BSGL`.
         """
-        if self.tCWFstatMapVersion == "lal":
-            F_mn = self.FstatMap.F_mn.data
-        else:
-            F_mn = self.FstatMap.F_mn
         # First, we need to also compute per-detector F_mn maps.
         # For now, we use the t0,tau index that maximises the multi-detector F
         # to return BSGL for a signal with those parameters.
         # FIXME: should we instead compute BSGL over the whole F_mn
         # and return the maximum of that?
-        idx_maxTwoF = np.unravel_index(np.argmax(F_mn), F_mn.shape)
+        idx_maxTwoF = np.unravel_index(
+            np.argmax(self.FstatMap.F_mn), self.FstatMap.F_mn.shape
+        )
         for X in range(self.FstatResults.numDetectors):
             # For each detector, we need to build a MultiFstatAtomVector
             # because that's what the Fstat map function expects.
@@ -1217,11 +1211,7 @@ class ComputeFstat(BaseSearchClass):
                 singleIFOmultiFatoms,
                 self.windowRange,
             )
-            if self.tCWFstatMapVersion == "lal":
-                F_mn_X = FXstatMap.F_mn.data
-            else:
-                F_mn_X = FXstatMap.F_mn
-            self.twoFXatMaxTwoF[X] = 2 * F_mn_X[idx_maxTwoF]
+            self.twoFXatMaxTwoF[X] = 2 * FXstatMap.F_mn[idx_maxTwoF]
         self.log10BSGL = lalpulsar.ComputeBSGL(
             self.maxTwoF, self.twoFXatMaxTwoF, self.BSGLSetup
         )
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -3,7 +3,6 @@
 
 import os
 import logging
-import copy
 from pprint import pformat
 
 import glob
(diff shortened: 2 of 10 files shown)


examples/transient_examples/PyFstat_example_short_transient_grid_search.py
```diff
@@ -32,9 +32,11 @@
     Alphas = [data.Alpha]
     Deltas = [data.Delta]
 
+    BSGL = False
+
     print("Standard CW search:")
     search1 = pyfstat.GridSearch(
-        label="CW",
+        label="CW" + ("-BSGL" if BSGL else ""),
```
Rodrigo-Tenorio, line 39:
```suggestion
        label="CW" + ("_BSGL" if BSGL else ""),
```

examples/transient_examples/PyFstat_example_short_transient_grid_search.py
```diff
@@ -43,16 +45,15 @@
...
     )
     search1.run()
     search1.print_max_twoF()
-
     search1.plot_1D(xkey="F0", xlabel="freq [Hz]", ylabel="$2\\mathcal{F}$")
 
     print("with t0,tau bands:")
     search2 = pyfstat.TransientGridSearch(
-        label="tCW",
+        label="tCW" + ("-BSGL" if BSGL else ""),
```
Rodrigo-Tenorio, line 56:
```suggestion
        label="tCW" + ("_BSGL" if BSGL else ""),
```
Any particular reason for your choice?

Rodrigo-Tenorio (commented):
Examples look fine, except for a `-`

examples/transient_examples/PyFstat_example_short_transient_grid_search.py
dbkeitel: no

pyfstat/core.py
```diff
@@ -415,6 +427,7 @@ def _get_SFTCatalog(self):
             SFTCatalog = lalpulsar.SFTCatalog()
             Toverlap = 0
             self.detector_names = self.detectors.split(",")
+            self.numDetectors = len(self.detector_names)
```
Rodrigo-Tenorio, line 430: Should we make this more robust with https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tutils__h.html#gafd482631e24c3af4a5458412c99cccf7 ?

pyfstat/core.py
```diff
@@ -962,13 +992,66 @@ def get_fullycoherent_twoF(
...
+            if self.BSGL is False:
+                return self.twoF
+            self.get_fullycoherent_single_IFO_twoFs()
+            self.get_fullycoherent_log10BSGL()
+            return self.log10BSGL
+        self.get_transient_maxTwoFstat(tstart, tend)
+        if self.BSGL is False:
+            return self.maxTwoF
+        else:
+            return self.get_transient_log10BSGL()
```
Rodrigo-Tenorio, line 1018: I find this block kind of convoluted. Reminder to check futher.

pyfstat/core.py
```diff
@@ -982,23 +1065,82 @@ def get_fullycoherent_twoF(
...
+        -------
+        twoFX: list
+            A list of the single-detector detection statistics twoF.
+            Also stored as `self.twoFX`.
+        """
+        self.twoFX[: self.FstatResults.numDetectors] = [
+            self.FstatResults.twoFPerDet(X)
+            for X in range(self.FstatResults.numDetectors)
+        ]
+        return self.twoFX
```
Rodrigo-Tenorio, line 1094: Suggestions:

1) Make it private
2) raise meaningful exception if `self.get_fullycoherent_twoF` was not run.
3) Make it run `self.get_fullycoherent_twoF` (problems?)

pyfstat/core.py
```diff
@@ -1007,51 +1149,83 @@ def get_fullycoherent_twoF(
...
+        -------
+        log10BSGL: float
+            A single value of the detection statistic log10BSGL
+            at the input parameter values.
+            Also stored as `self.log10BSGL`.
+        """
+        if self.tCWFstatMapVersion == "lal":
+            F_mn = self.FstatMap.F_mn.data
+        else:
+            F_mn = self.FstatMap.F_mn
```
Rodrigo-Tenorio, line 1192: suggestion:

```
def __init__(self, yada yada yada, tCWFstatMapVersion="whatever"):
    # ... initi stuff ...
    self.get_tCWFstatMap = {"lal":_get_lal_tCWFstatMap, "pycuda": _get_cuda_tCWFstatMap}

def _get_lal_tCWFstatMap(FstatMap):
    return FstatMap.F_mn.data
    
def _get_cuda_tCWFstatMap(FstatMap):
    return FstatMap.F_mn
```

An now these blocks can be simplified (once for all!) to
```
F_mn = self.get_tCWFstatMap[self.tCWFstatMapVersion](self.FstatMap.F_mn)
```
Instead of a dictionary we can just assign the suitable method and simplify even further.

pyfstat/core.py
dbkeitel: [MultiAddToFakeSFTCatalog](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tutils__h.html#ga47d5e1fa610345c239ba9f0418e4048d) a few lines down already calls [XLALGetChannelPrefix](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tutils__h.html#gadd0ea28e0308a22f53bbabaa3b2896be) which _should_ also fail on invalid names, so I think we _should_ be good here...?

pyfstat/core.py
dbkeitel: It's the "return early" philosophy which I also personally find harder to read than a fully hierarchical if-else construct, but I decided to keep it logically in that same structure as it was before.

pyfstat/core.py
dbkeitel: >    Make it private

I've thought about it, but there may be use cases where one wants to compute these post-hoc. On the other hand, it will always just work on the last grid point, so such cases would be fairly limited, and maybe we should just decide to discourage users from single-point operations on search classes in general?

>   raise meaningful exception if self.get_fullycoherent_twoF was not run.

Good idea, but complicated by the fact that "not run" could mean both "never run" and "only run on previous grid points, not the current one". Which would be an argument in favour of "discourage users from single-point operations on search classes", because then one could safely assume that the latter doesn't happen.

>    Make it run self.get_fullycoherent_twoF (problems?)

Then the structure of `get_fullycoherent_detstat` would need to be changed to not double-call it, and would be less intuitive.

Rodrigo-Tenorio (changes requested):
First review of core. Some suggestions.

pyfstat/core.py
dbkeitel: I'm longer-term planning a proper object-based rewrite for the whole "map" stuff anyway.

But I think you're right that just those few lines should work without extra effort.

pyfstat/grid_based_searches.py
```diff
@@ -722,6 +738,20 @@ def plot_2D(
         else:
             return ax
 
+    def get_max_det_stat(self):
+        """Get the maximum detection statistic over the grid.
+
+        This requires the `run()` method to have been called before.
```
Rodrigo-Tenorio, line 744: Same: `raise` something meaningful. I'll think about it.

pyfstat/grid_based_searches.py
```diff
@@ -722,6 +738,20 @@ def plot_2D(
...
+
+        Returns
+        -------
+        d: dict
+            Dictionary containing parameters and detection statistic at the maximum.
+        """
+        idx = np.argmax(self.data[self.detstat])
+        d = OrderedDict([(key, self.data[key][idx]) for key in self.output_keys])
```
Rodrigo-Tenorio, line 752: Do we need `OrderedDict`?  I though since 3.6 dictionaries preserved the order.

pyfstat/grid_based_searches.py
```diff
@@ -992,11 +1031,19 @@ def run(self, return_data=False):
...
             self.timingFstatMap += getattr(self.search, "timingFstatMap", 0.0)
-            thisCand = list(vals) + [detstat]
+            thisCand.append(self.search.twoF)
+            if hasattr(self.search, "twoFX"):
+                thisCand += list(self.search.twoFX[: self.search.numDetectors])
+            thisCand.append(self.search.maxTwoF)
+            if hasattr(self.search, "twoFXatMaxTwoF"):
+                thisCand += list(self.search.twoFXatMaxTwoF[: self.search.numDetectors])
+            if self.detstat != "maxTwoF":
+                thisCand.append(detstat)
```
Rodrigo-Tenorio, line 1046: Reminder: I've seen this block too many times

pyfstat/mcmc_based_searches.py
```diff
@@ -1698,11 +1758,16 @@ def export_samples_to_disk(self):
...
         if idx is None:
-            return (self.lnlikes - self.likelihoodcoef) * 2
+            return (
+                self.lnlikes - self.likelihoodcoef
+            ) / self.likelihooddetstatmultiplier
         else:
-            return (self.lnlikes[idx] - self.likelihoodcoef) * 2
+            return (
+                self.lnlikes[idx] - self.likelihoodcoef
+            ) / self.likelihooddetstatmultiplier
```
Rodrigo-Tenorio, line 1770: Reminder: I'm not sure, but maybe this can be simplified

pyfstat/grid_based_searches.py
dbkeitel: Only since 3.7, apparently: https://docs.python.org/3.7/whatsnew/3.7.html

It also seems to retain a few extra methods, but we don't need those.

pyfstat/mcmc_based_searches.py
dbkeitel: As in, get rid of the if-else?

pyfstat/grid_based_searches.py
Rodrigo-Tenorio: Those willing to deprecate PyFstat's Python 3.6 support say `Yea` :drum: ...

pyfstat/mcmc_based_searches.py
Rodrigo-Tenorio: Ja! Behold!
```
return (self.lnlikes[idx or ...] - self.likelihoodcoef) / self.likelihooddetstatmultiplier
```
Where `...` is the `Ellipsis` object which takes everything.

pyfstat/grid_based_searches.py
dbkeitel: Let's make that a separate PR after lalsuite finally supports 3.9. I like having 3 supported versions. ;)

pyfstat/mcmc_based_searches.py
dbkeitel: Nice! Want to add it as a suggestion or your own commit?

Rodrigo-Tenorio: f96d742

Rodrigo-Tenorio: Let me eat my words: `idx or ...` fails if `idx == 0`.

dbkeitel: `[... if idx is None else idx]` should work, at least on np arrays, which I think `self.nlikes` should be...?

Rodrigo-Tenorio: a10981d

pyfstat/core.py
Rodrigo-Tenorio: Looks like so

pyfstat/core.py
Rodrigo-Tenorio: a9d3bb70f538f9a251a413e73df3b0436380eeda

pyfstat/grid_based_searches.py
Rodrigo-Tenorio: This is consistent with the way in which `mcmc` methods operate.

pyfstat/grid_based_searches.py
Rodrigo-Tenorio: No, I didn't. Sorry for the noise.

pyfstat/mcmc_based_searches.py
```diff
@@ -1677,8 +1722,23 @@ def export_samples_to_disk(self):
...
+        detstat = np.atleast_2d(self._get_detstat_from_loglikelihood()).T
+        if self.BSGL:
+            twoF = np.zeros_like(detstat)
+            self.search.BSGL = False
+            for idx, samp in enumerate(self.samples):
+                p = self._set_point_for_evaluation(samp)
+                if isinstance(p, dict):
+                    twoF[idx] = self.search.get_det_stat(**p)
+                else:
+                    twoF[idx] = self.search.get_det_stat(*p)
```
Rodrigo-Tenorio, line 1737: Check this further: Isn't it risky?

Rodrigo-Tenorio: Not necessarily: Everything is "private".

Rodrigo-Tenorio (approved):
Nice!

pyfstat/core.py
dbkeitel: @Rodrigo-Tenorio any further thoughts on this aspect?

Rodrigo-Tenorio: Read-the-docs philosophy: Leave as is

dbkeitel: issue for future consideration -> #262

PR #246, opened 2021-01-07
Doc fixes

dbkeitel:
- [x] add an internal method docstring suggested but then forgotten in #171
- [x] try to enforce `fail_on_warning`
- [x] fix warnings once exposed

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,6 +8,7 @@ version: 2
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/source/conf.py
+  fail_on_warning: true
 
 # Build documentation with MkDocs
 #mkdocs:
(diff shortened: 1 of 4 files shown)


dbkeitel:
Failed successfully: https://readthedocs.org/projects/pyfstat/builds/12711301/

dbkeitel:
MCMC formatting warnings fixed, but there's a spurious matplotlib warning I don't have locally that shows up in the CI and then triggers the failure too:
```
WARNING:matplotlib.font_manager:Matplotlib is building the font cache; this may take a moment.
```
Probably because it's always run in a clean container.

dbkeitel:
Actually that matplotlib warning doesn't seem to happen every time, but there was another justified one:
```
WARNING: html_static_path entry '_static' does not exist
```

dbkeitel:
Will merge this for now, but we have to watch if the matplotlib thing happens more often and revert the `fail_on_warning` if necessary.

PR #251, opened 2021-01-08
sphinx: use pyfstat.__version__ at least on local builds

dbkeitel:
For the actual readthedocs builds, this doesn't seem to be currently possible:
* https://github.com/readthedocs/readthedocs.org/issues/4529
But for PRs the extra banner is clear enough, "latest" is always "latest" by definition, and while it would be nice to say on "stable" which version it is, at least I can "activate" additional stable-stable tagged versions for posterity.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -22,5 +22,9 @@ formats:
 python:
   version: 3.8
   install:
-    - requirements: docs/requirements.txt
-    - requirements: requirements.txt
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
+    - method: setuptools
+      path: .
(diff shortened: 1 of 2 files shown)


dbkeitel:
I think I got the `.yml` changed correctly to install `pyfstat` itself inside the build environment, but for some reason it now cannot find the right lalsuite version: https://readthedocs.org/projects/pyfstat/builds/12718424/

Abandoning this for now to get v1.10.0 out.

Possible solutions:

- figure out why it hates lalsuite

- switch to conda

dbkeitel:
Seems the "it hates lalsuite" has ben resolved by waiting 2 months. :)

Looks like this now on local build:
image


And unchanged for PR builds:
image

Rodrigo-Tenorio (approved):

PR #255, opened 2021-01-11
bump to lalsuite>=6.80 and use lalpulsar.ValidateSFTFile()

dbkeitel:
This was introduced in https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1428 and can now be used instead of the `lalapps` equivalent. We only used that one once in the tests, though.

Using the opportunity for a general lalsuite requirement version bump too, also to exercise our release and upgrade chain when it comes to those...

@Rodrigo-Tenorio main review question: am I overlooking any other place where versions are defined for conda/docker/...?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,5 +8,5 @@ tqdm
 bashplotlib
 peakutils
 pathos
-lalsuite>=6.76
+lalsuite>=6.80
 versioneer
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup(
         "bashplotlib",
         "peakutils",
         "pathos",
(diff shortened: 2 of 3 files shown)


Rodrigo-Tenorio:
No. Docker gets the requirements from `requirements.txt`, and so does `conda` (if I'm not mistaken).

Rodrigo-Tenorio (approved):

dbkeitel:
Conda-forge is supposed to pick it up from either that or the setup.py (not sure which) through their automatic bot, or at least it asks me to review "Dependencies have been updated if changed" on each release. Let's see if it works as expected.

PR #256, opened 2021-01-11
PredictFstat: simplify input checking

### LINKED ISSUE #157: predict_fstat signal parameter handling simplifications
Once https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1430 has made it into a PyPi release, some logic in `helper_functions.predict_fstat()` can be simplified.

dbkeitel:
- resolves #157
 - removes workarounds for insufficient checks on lalapps side
   which have been fixed in https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1430

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -1001,25 +1001,8 @@ def predict_fstat(
     cl_pfs = []
     cl_pfs.append("lalapps_PredictFstat")
 
-    if h0 is not None:
-        cl_pfs.append("--h0={}".format(h0))
-    if cosi is not None:
-        cl_pfs.append("--cosi={}".format(cosi))
-    # NOTE: as of lalsuite 6.76, [psi,Alpha,Delta] are required even for
-    # noise-only calls, hence we default them to 0 if h0 is None or ==0,
-    # but fail if they're not set with h0>0.
-    # This can likely be simplified after a future lalsuite release.
-    pars = {"psi": psi, "Alpha": Alpha, "Delta": Delta}
-    for par in pars.keys():
-        if pars[par] is None:
-            if h0:
-                raise ValueError(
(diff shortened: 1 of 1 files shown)


pyfstat/helper_functions.py
```diff
@@ -1001,24 +1001,9 @@ def predict_fstat(
...
-        if pars[par] is None:
-            if h0:
-                raise ValueError(
-                    "For h0>0, {:s} needs to be set explicitly.".format(par)
-                )
-            else:
-                cl_pfs.append("--{:s}=0".format(par))
-        else:
+        if pars[par] is not None:
             cl_pfs.append("--{:s}={}".format(par, pars[par]))
```
Rodrigo-Tenorio, line 1007: Suggestion:
```
cl_pfs.extend(["--{:s}={}".format(key, value 
for key, value in pars.items() if value is not None])
```

Rodrigo-Tenorio: Also add a FIXME: The list of `predict_fstat` parameters should be written somewhere, so they can be filtered out from any dictionary without reinventing the loop.

Rodrigo-Tenorio (changes requested):
Opinion? 

The rest looks fine

pyfstat/helper_functions.py
dbkeitel: There is already `get_predict_fstat_parameters_from_dict()` for that purpose, but its internal list is _not_ identical to the one used for commandline setup here, as there's a `F0`<->`Freq` translation here. We could employ `translate_keys_to_lal`, but I'd like to keep this to a minimum maintenance edit here.

Because, long-term goal: our own PFS class which makes the lalapps call superfluous.

Rodrigo-Tenorio (approved):

Issue #263, opened 2021-01-22
move priors info from wiki to docs


dbkeitel:
@Rodrigo-Tenorio now that we have proper docs, I think the info from the page https://github.com/PyFstat/PyFstat/wiki/Priors should go there. ("Wiki for repo-related things (developers/maintainers), docs for package-related things (users)" philosophy).

The info could either go into the top-level info of `mcmc_based_searches` (where it currently says "The general approach...") or into a separate .rst file linked from there.

PR #265, opened 2021-01-22
[docs mcmc_based_searches.py] Add prior documentation

Rodrigo-Tenorio:
Address #263


dbkeitel (approved):
Thanks! It looks a bit odd as a single TOC entry, see below. Maybe we can figure out how to add entries for the classes there too. But good enough for now.

Screenshot_20210122_203028

Issue #184, opened 2020-09-18
axis label auto-scaling issues


dbkeitel:
Known issue: axis labels on corner plots often become unreadable, especially for frequencies with many post-decimal digits. Also affects the recently revived (#183) "chainconsumer" plots.

Since we don't necessarily promise publication-ready plots, this isn't high priority to address.

dbkeitel:
@Rodrigo-Tenorio are we happy enough with the current (vastly improved) state?

Rodrigo-Tenorio:
There's still one last thing I'd like to have a look at: Adjust the format of title values (the ones shown at the top of each histogram) not to show 0.00 +- 0.00 whenever values are too small.

PR #266, opened 2021-01-25
Change plot_corner defaults

Rodrigo-Tenorio:
Partially adressing #184 .

It's not the best solution, but does the trick and took me way less than I expected, so let's rejoice I say.


Rodrigo-Tenorio:
Here's an example of the new defaults

Rodrigo-Tenorio:
@dbkeitel Examples look actually quite nice

dbkeitel (approved):

PR #267, opened 2021-01-26
fix TestMCMCSearchBSGL.test_MCMC_search_on_data_with_line

dbkeitel:
- signal and line need to be separated well enough in freq


dbkeitel:
This setup should now be robust enough:

test
TestMCMCSearch-F_walkers
TestMCMCSearch-BSGL_walkers

dbkeitel:
Further need to disable the quantiles check for this one, as the cheap MCMC setup isn't good enough for the wider frequency prior, and I don't want to make it more expensive just to re-cover the basic convergence already covered in other cases.

PR #269, opened 2021-01-26
do NOT always set matplotlib backend to agg

### LINKED ISSUE #268: matplotlib backend is always set to agg, preventing interactive use
There are checks in core.py and helper_functions.py to set matplotlib to the non-interactive `agg` backend if the `DISPLAY` env variable isn't set. However, we actually ALWAYS use it through `set_up_matplotlib_defaults()`, even if we do have a `DISPLAY`. Observe on a local terminal:
```
>>> import os
>>> "DISPLAY" in os.environ
True
>>> import matplotlib.pyplot as plt
>>> plt.plot(0,1); plt.show();
[<matplotlib.lines.Line2D object at 0x7fa337662e48>]
>>> import pyfstat
>>> plt.plot(0,1); plt.show();
[<matplotlib.lines.Line2D object at 0x7fa33cdf2e10>]
__main__:1: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
```

dbkeitel:
- this was still done through helper_functions.set_up_matplotlib_defaults()
 - instead, we should only rely on the optional import logic
   based on whether "DISPLAY" is set in the env
 - closes #268

Works for me on my laptop and on CIT. @Rodrigo-Tenorio any concerns this may be a problem in docker containers?

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -48,7 +48,6 @@ def set_up_optional_tqdm():
 
 def set_up_matplotlib_defaults():
     """Sets some defaults for matplotlib plotting."""
-    plt.switch_backend("Agg")
     plt.rcParams["text.usetex"] = shutil.which("latex") is not None
     plt.rcParams["axes.formatter.useoffset"] = False
 


Rodrigo-Tenorio:
I just installed `matplotlib` in a Singularity container (which is what we'll be using in the end). 

There the backed is set to `TkAgg` and plots show as expected.

Is this sufficient? I'm honestly not that familiar with matplotlib's guts.

Rodrigo-Tenorio (commented):
I just installed matplotlib in a Singularity container (which is what we'll be using in the end).

There the backed is set to TkAgg and plots show as expected.

Is this sufficient? I'm honestly not that familiar with matplotlib's guts.

Rodrigo-Tenorio:
I forced `TkAgg` backend in one of the examples, nothing changed either.

Rodrigo-Tenorio (approved):

PR #277, opened 2021-02-11
FrequencyModulatedArtifactWriter: Juggle input params to circumvent a parent check

Rodrigo-Tenorio:
- Checks in `_basic_setup` clash against different options in Artifact.
- Now it works.

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -1262,7 +1262,14 @@ class FrequencyModulatedArtifactWriter(Writer):
         self.phi = 0
         self.F2 = 0
 
+        self.cosi = 0
+        self.noiseSFTs = None
+
+        h0_value = self.h0
+        self.h0 = None
         self._basic_setup()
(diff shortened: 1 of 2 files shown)


pyfstat/make_sfts.py
```diff
@@ -1262,7 +1262,14 @@ def __init__(
         self.phi = 0
         self.F2 = 0
 
+        self.cosi = 0
+        self.noiseSFTs = None
+
```
dbkeitel, line 1267:
```suggestion
# The _basic_setup() method inherited from Writer
# requires additional CW signal parameters if h0>0,
# which an artifact doesn't need to have,
# hence we temporarily unset it here as a workaround
# and restore it after the method call.
```

dbkeitel (changes requested):

dbkeitel (approved):
This won't quite make the class work again (need to track lalapps_splitSFTs changes), but merging as an improvement for now.

PR #278, opened 2021-02-11
fix and test FrequencyModulatedArtifactWriter

dbkeitel:
@Rodrigo-Tenorio followup to #277

 - needed to adapt concatenate_sft_files()
   method to changed lalapps_splitSFTs filename convention

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -1264,6 +1264,10 @@ class FrequencyModulatedArtifactWriter(Writer):
 
         self.cosi = 0
         self.noiseSFTs = None
+
+        if type(self.detectors) is not str or len(self.detectors.split(",")) > 1:
+            raise ValueError("'detectors' must be  a single-IFO string")
+
         # The _basic_setup() method inherited from Writer
         # requires additional CW signal parameters if h0>0,
         # which an artifact doesn't need to have,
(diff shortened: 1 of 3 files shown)


dbkeitel:
...now with less actions ;)

pyfstat/make_sfts.py
```diff
@@ -1378,39 +1378,40 @@ def get_h0(self, t):
...
-            self.detectors[1],
-            self.nsfts,
-            self.Tsft,
-            int(self.tstart),
-            int(self.duration),
-            self.label,
+        """Merges the individual SFT files via lalapps_splitSFTs executable."""
+
+        SFTFilename = (
+            f"{self.detectors[0]}-{self.nsfts}_{self.detectors}_{self.Tsft}SFT"
```
Rodrigo-Tenorio, line 1384: This line is just beautiful.

I guess `self.detectors` is checked to be a string somewhere?

Rodrigo-Tenorio (changes requested):

pyfstat/make_sfts.py
dbkeitel: > This line is just beautiful.

For reference, the original lalapps snprintf beauty is https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalapps/src/pulsar/SFTTools/splitSFTs.c#L688

> I guess self.detectors is checked to be a string somewhere?

Purely coincidentally and implicitly, yes:

https://github.com/PyFstat/PyFstat/blob/af6d263c1a8da07e54b3a7649a7fdca855cd2341/pyfstat/make_sfts.py#L1337

Rodrigo-Tenorio (approved):

pyfstat/make_sfts.py
dbkeitel: But actually if I change in the test to `detectors="H1,L1"`, this one does not catch it. Hmm.

dbkeitel: Ah, because it's only reached if adding fake Alpha,Delta dependent modulation.

Will add an explicit check earlier.

dbkeitel: done in d1443ddf0e106ca73ca6b44205e52452507635da

Rodrigo-Tenorio (approved):

PR #281, opened 2021-02-12
Implement line writer

Rodrigo-Tenorio:
Implement a proper class to inject (possibly transient) lines in data using Makefakedata_v4.

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -945,14 +945,14 @@ class BinaryModulatedWriter(Writer):
         )
 
 
-class TransientLineWriter(Writer):
-    """Inject <tbd> into SFT data.
+class LineWriter(Writer):
+    """Inject an line into SFT data.
 
-    WIP
+    A (transient) line is defined as a constant amplitude and constant excess power artifact in the data.
 
     In practice, it corresponds to a CW without Doppler or antenna-patern-induced amplitude modulation.
 
-    This functionality is implemented through `lalapps_MakeFakeData_v4`.
+    This functionality is implemented via `lalapps_MakeFakeData_v4`'s `lineFeature` option.
     This version of MFD only supports one interferometer at a time.
     """
 
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -1020,7 +1020,7 @@ class LineWriter(Writer):
             logging.info(
                 "Computing h0_eff from h0 and cosi. "
                 "This value will be stored in h0, "
-                "as cosi is *not* used by lineFeature"
+                "as cosi is *not* used by lineFeature."
             )
             eta = self.signal_parameters["cosi"]
             eta2 = eta * eta
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -191,6 +191,8 @@ class Writer(BaseSearchClass):
     for more detailed help with some of the parameters.
     """
 
+    mfd = "lalapps_Makefakedata_v5"
+
     signal_parameter_labels = [
         "tref",
         "F0",
@@ -747,9 +749,23 @@ class Writer(BaseSearchClass):
         to see if equivalent data files already exist,
         and else runs the actual generation code.
         """
+        cl_mfd = self._build_MFD_command_line()
 
-        mfd = "lalapps_Makefakedata_v5"
-        cl_mfd = [mfd]
+        check_ok = self.check_cached_data_okay_to_use(cl_mfd)
+        if check_ok is False:
+            helper_functions.run_commandline(cl_mfd)
+            if not np.all([os.path.isfile(f) for f in self.sftfilenames]):
+                raise IOError(
+                    f"It seems we successfully ran {self.mfd},"
+                    f" but did not get the expected SFT file path(s): {self.sftfilepath}."
+                )
+            logging.info(f"Successfully wrote SFTs to: {self.sftfilepath}")
+            logging.info("Now validating each SFT file...")
+            for sft in self.sftfilenames:
+                lalpulsar.ValidateSFTFile(sft)
+
+    def _build_MFD_command_line(self):
+        cl_mfd = [self.mfd]
         cl_mfd.append("--outSingleSFT=TRUE")
         cl_mfd.append('--outSFTdir="{}"'.format(self.outdir))
         cl_mfd.append('--outLabel="{}"'.format(self.label))
@@ -802,19 +818,7 @@ class Writer(BaseSearchClass):
         if self.randSeed:
             cl_mfd.append("--randSeed={}".format(self.randSeed))
 
-        cl_mfd = " ".join(cl_mfd)
-        check_ok = self.check_cached_data_okay_to_use(cl_mfd)
-        if check_ok is False:
-            helper_functions.run_commandline(cl_mfd)
-            if not np.all([os.path.isfile(f) for f in self.sftfilenames]):
-                raise IOError(
-                    f"It seems we successfully ran {mfd},"
-                    f" but did not get the expected SFT file path(s): {self.sftfilepath}."
-                )
-            logging.info(f"Successfully wrote SFTs to: {self.sftfilepath}")
-            logging.info("Now validating each SFT file...")
-            for sft in self.sftfilenames:
-                lalpulsar.ValidateSFTFile(sft)
+        return " ".join(cl_mfd)
 
     def predict_fstat(self, assumeSqrtSX=None):
         """Predict the expected F-statistic value for the injection parameters.
@@ -952,6 +956,8 @@ class TransientLineWriter(Writer):
     This version of MFD only supports one interferometer at a time.
     """
 
+    mfd = "lalapps_Makefakedata_v4"
+
     required_mfd_line_parameters = [
         "Freq",
         "phi0",
@@ -1036,14 +1042,8 @@ class TransientLineWriter(Writer):
         super().calculate_fmin_Band()
         self.noiseSFTs = hide_noiseSFTs
 
-    def run_makefakedata(self):
-        """Generate the SFT data calling lalapps_Makefakedata_v4.
-
-        This first builds the full commandline,
-        then calls `check_cached_data_okay_to_use()`
-        to see if equivalent data files already exist,
-        and else runs the actual generation code.
-        """
+    def _build_MFD_command_line(self):
+        """Generate the SFT data calling lalapps_Makefakedata_v4."""
 
         mfd = "lalapps_Makefakedata_v4"
         cl_mfd = [mfd]
@@ -1100,19 +1100,7 @@ class TransientLineWriter(Writer):
         if self.randSeed:
             cl_mfd.append("--randSeed={}".format(self.randSeed))
 
-        cl_mfd = " ".join(cl_mfd)
-        check_ok = self.check_cached_data_okay_to_use(cl_mfd)
-        if check_ok is False:
-            helper_functions.run_commandline(cl_mfd)
-            if not np.all([os.path.isfile(f) for f in self.sftfilenames]):
-                raise IOError(
-                    f"It seems we successfully ran {mfd},"
-                    f" but did not get the expected SFT file path(s): {self.sftfilepath}."
-                )
-            logging.info(f"Successfully wrote SFTs to: {self.sftfilepath}")
-            logging.info("Now validating each SFT file...")
-            for sft in self.sftfilenames:
-                lalpulsar.ValidateSFTFile(sft)
+        return " ".join(cl_mfd)
 
 
 class GlitchWriter(SearchForSignalWithJumps, Writer):
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -211,8 +211,8 @@ class Writer(BaseSearchClass):
     gps_time_and_string_formats_as_LAL = {
         "refTime": ":10.9f",
         "transientWindowType": ":s",
-        "transientStartTime": ":10.9f",
-        "transientTau": ":10.9f",
+        "transientStartTime": ":10.0f",
+        "transientTau": ":10.0f",
     }
     """Dictionary to ensure proper format handling for some special parameters.
 
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -942,14 +942,14 @@ class BinaryModulatedWriter(Writer):
 
 
 class TransientLineWriter(Writer):
-    """Inject a transient line into some data.
+    """Inject <tbd> into SFT data.
 
-    A transient line is defined as a constant amplitude and frequency excess power lasting
-    for a (not necesarilly) finite time span of an observing run.
+    WIP
 
     In practice, it corresponds to a CW without Doppler or antenna-patern-induced amplitude modulation.
 
-    This uses MFD_v4
+    This functionality is implemented through `lalapps_MakeFakeData_v4`.
+    This version of MFD only supports one interferometer at a time.
     """
 
     required_mfd_line_parameters = [
@@ -971,7 +971,8 @@ class TransientLineWriter(Writer):
         elif len(self.detectors.split(",")) > 1:
             raise NotImplementedError(
                 "MakeFakeData_v4 does not support more than one detector at a time. "
-                "Please, do it properly"
+                "Multi-detector behaviour can be reproduced by calling the procedure "
+                "on single-detector SFT sets once at a time."
             )
 
     def correct_line_amplitude(self, use_effective_h0, scale_by_duration):
@@ -1011,7 +1012,11 @@ class TransientLineWriter(Writer):
             params_to_purge = list(
                 set(self.signal_parameters) - set(self.required_mfd_line_parameters)
             )
-            print("Purging: {}".format(params_to_purge))
+            logging.info(
+                "Purging input parameters that are not meaningful for <tbd>: {}".format(
+                    params_to_purge
+                )
+            )
             for key in params_to_purge:
                 self.signal_parameters.pop(key)
 
@@ -1032,7 +1037,7 @@ class TransientLineWriter(Writer):
         self.noiseSFTs = hide_noiseSFTs
 
     def run_makefakedata(self):
-        """Generate the SFT data calling lalapps_Makefakedata_v5.
+        """Generate the SFT data calling lalapps_Makefakedata_v4.
 
         This first builds the full commandline,
         then calls `check_cached_data_okay_to_use()`
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -975,7 +975,22 @@ class TransientLineWriter(Writer):
             )
 
     def correct_line_amplitude(self, use_effective_h0, scale_by_duration):
-        pass
+        """
+        Modify the amplitude parameter h0 to include effects from polarization (cosi)
+        or account for shorter signals.
+
+        use_effective_h0: bool
+            If true, use h0_eff = h0 * sqrt(cosi^4 + 6 * cosi^2 + 1)
+        scale_by_duration: bool
+            If true, scale h0 up by sqrt(t_obs/tau)
+        """
+
+        if use_effective_h0:
+            eta = self.signal_parameters["cosi"]
+            eta2 = eta * eta
+            self.signal_parameters["h0"] *= np.sqrt(eta2 * eta2 + 6 * eta2 + 1)
+        if scale_by_duration:
+            self.signal_parameters["h0"] *= np.sqrt(self.duration / self.transientTau)
 
     def _parse_args_consistent_with_mfd(self):
         """
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -513,12 +513,6 @@ class Writer(BaseSearchClass):
             for key in self.signal_parameters
         }
 
-        if self.signal_parameters["transientWindowType"] != "none":
-            self.signal_parameters["transientStartTime"] = self.transientStartTime
-            self.signal_formats["transientStartTime"] = ":10.0f"
-            self.signal_parameters["transientTau"] = self.transientTau
-            self.signal_formats["transientTau"] = ":10.0f"
-
     def calculate_fmin_Band(self):
         """Set fmin and Band for the output SFTs to cover.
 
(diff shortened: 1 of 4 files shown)


tests.py
```diff
@@ -430,6 +438,53 @@ def test_data_with_gaps(self):
...
+            times < (writer.transientStartTime + writer.transientTau)
+        )
+        max_power_freq_index_with_line = max_power_freq_index[line_active_mask]
+
+        # Maximum power should be a the transient line whenever that's on
+        print(max_power_freq_index_with_line)
+        self.assertTrue(
+            np.all(max_power_freq_index_with_line == max_power_freq_index_with_line[0])
+        )
+        print(freqs[max_power_freq_index_with_line])
```
Rodrigo-Tenorio, line 482: This shows a line injected 50 bins away from the actual value.

Rodrigo-Tenorio: Solved, I was just parsing the keys wrongly.

tests.py
```diff
@@ -65,6 +66,13 @@ def tearDownClass(self):
     "argp": 0.3,
 }
 
+default_transient_params = {
```
dbkeitel, line 69: Note there's already several classes that use transient parameters, but not these default choices (usually, it seems:
```
self.transientStartTime = int(self.tstart + 0.25 * self.duration)
self.transientTau = int(0.5 * self.duration)
```
).  Not really a problem though.

tests.py
```diff
@@ -430,6 +438,121 @@ def test_data_with_gaps(self):
...
+                duration=self.duration,
+                Tsft=self.Tsft,
+                tstart=self.tstart,
+                detectors=detectors,
+                sqrtSX=self.sqrtSX,
+                Band=0.5,
+                **self.signal_parameters,
+            )
+
+    def _check_maximum_power_consistency(self, writer, return_line_power=False):
```
dbkeitel, line 461: note to self: not checked the rest below here in detail yet

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
         )
 
 
+class TransientLineWriter(Writer):
```
dbkeitel, line 944: As discussed, I'd prefer a more general name. `LineWriter` or `UnmodulatedArtifactWriter`.

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
...
+    """Inject a transient line into some data.
+
+    A transient line is defined as a constant amplitude and frequency excess power lasting
+    for a (not necesarilly) finite time span of an observing run.
+
+    In practice, it corresponds to a CW without Doppler or antenna-patern-induced amplitude modulation.
+
+    This uses MFD_v4
```
dbkeitel, line 952: add here that it's single-IFO

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
         )
 
 
+class TransientLineWriter(Writer):
+    """Inject a transient line into some data.
+
+    A transient line is defined as a constant amplitude and frequency excess power lasting
+    for a (not necesarilly) finite time span of an observing run.
```
dbkeitel, line 948:
```suggestion
    A (transient) line is defined as a constant amplitude and constant frequency excess power artifact,
    lasting for a finite time span or the whole data duration.
```

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
...
+    def __init__(self, *args, **kwargs):
+
+        super().__init__(*args, **kwargs)
+
+        if self.detectors is None:
+            raise ValueError("MakeFakeData_v4 requires detector name to be given")
+        elif len(self.detectors.split(",")) > 1:
+            raise NotImplementedError(
+                "MakeFakeData_v4 does not support more than one detector at a time. "
+                "Please, do it properly"
```
dbkeitel, line 974:
```suggestion
```

don't admonish the user too much ;)

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
...
+            )
+
+    def correct_line_amplitude(self, use_effective_h0, scale_by_duration):
```
dbkeitel, line 977: I don't like that name, there's nothing "incorrect" about just a fixed h0... `rescale_amplitude`?

dbkeitel: Also this seems a bit clunky as an interface, doing two completely disjoint operations, and in both cases assuming the user had already made a "good" guess of h0 in some other sense. It's also not called automatically by the class apparently?

Suggestions:
1. drop this and make it the user's responsibility to know which h0 they want
2. only drop the more trivial duration path, and add an optional `h0_eff` argument to the class itself which will at init time use that one and cosi to override h0

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
...
+                set(self.signal_parameters) - set(self.required_mfd_line_parameters)
+            )
+            print("Purging: {}".format(params_to_purge))
+            for key in params_to_purge:
+                self.signal_parameters.pop(key)
+
+        if "transientTau" in self.signal_parameters:
+            self.signal_parameters["transientTauDays"] = (
+                self.signal_parameters.pop("transientTau") / 86400.0
+            )
```
dbkeitel, line 1021: Probably ok if tests pass, but shouldn't this be an int?

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
...
+            key not in self.required_mfd_line_parameters
+            for key in self.signal_parameters
+        ):
+            logging.warning(
+                "Transient lines only require the following parameters:\n"
+                f"{self.required_mfd_line_parameters}.\n"
+                "Any other parameter will be purged from this class now"
+            )
+            params_to_purge = list(
+                set(self.signal_parameters) - set(self.required_mfd_line_parameters)
```
dbkeitel, line 1012: Oh, another idiom for the non-intersection of two lists... I've used `np.setdiff1d` so far, but since there's nothing numpylicious about these simple lists of things, yours is fine.

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
...
+            )
+            print("Purging: {}".format(params_to_purge))
```
dbkeitel, line 1014:
```suggestion
            logging.info("Purging input parameters that are not meaningful for an unmodulated line injection: {}".format(params_to_purge))
```

(black won't like this)

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
...
+
+        if "transientTau" in self.signal_parameters:
+            self.signal_parameters["transientTauDays"] = (
+                self.signal_parameters.pop("transientTau") / 86400.0
+            )
+            self.signal_formats["transientTauDays"] = self.signal_formats.pop(
+                "transientTau"
+            )
+
+    def calculate_fmin_Band(self):
```
dbkeitel, line 1026:
```suggestion
    def _calculate_fmin_Band(self):
```
(and update where it's called below)

pyfstat/make_sfts.py
```diff
@@ -943,6 +941,175 @@ def __init__(
...
+        # MakeFakeData_v4 *requires* fmin and Band to be given.
+        # This overriding prevents Writer.calculate_fmin_Band from
+        # not setting them due to self.noiseSFTs being truthy.
+        hide_noiseSFTs, self.noiseSFTs = self.noiseSFTs, None
+        super().calculate_fmin_Band()
+        self.noiseSFTs = hide_noiseSFTs
+
+    def run_makefakedata(self):
+        """Generate the SFT data calling lalapps_Makefakedata_v5.
```
dbkeitel, line 1035:
```suggestion
        """Generate the SFT data calling lalapps_Makefakedata_v4.
```

dbkeitel (changes requested):
Looks good overall, but some polishing still needed.

- [ ] see inline comments
- [ ] add to `__init__.py`
- [x] Optional: could reduce code duplication by introducing `_build_MFD_base_commandline` and `_execute_MFD` methods to the main `Writer` class, which are both called from `run_makefakedata`, with only the differences remaining in that one and needing to be overridden

tests.py
Rodrigo-Tenorio: I'll let you decide what to do. I can change the name to `default_transient_line_params` if you prefere.

pyfstat/make_sfts.py
Rodrigo-Tenorio: What about
```
MonochromaticArtifactWriter
```
?

We are technically injecting a signal
```
h(t) = h0 * sin(Omega * t + Phi0)
```
where `h0 = ct.`, meaning it's as monochromatic as it can get.

Of course, that doesn't hold when we include transient lines (i.e. windowed lines), but `Unmodulated` doesn't either.

About the `Artifact` term... I don't have a strong opinion.

pyfstat/make_sfts.py
Rodrigo-Tenorio: a) `correct` should be understood in the sense of  "add extra terms to whatever you already have". I agree is not the best term, I just put it there to have it done, I'll change it once we clarify how to implement this.

b) I don't like to re-implement a function to multiply `h0` by `eta⁴ + 6 eta² + 1` every time, or another one related to the SNR scaling (since it's very easy to mess it up). I think we should have these relations somewhere (specially the h0_eff one).

The intended use case was to allow the user to apply at its own will either (or both) of these corrections. I.e. once the writer is set up, the user can choose whether to modify it or not.

Since `cosi` is completely ignored by MFDv4 in this case, I also though about using `h0eff` always if `cosi` is given.

So, a second proposal

- [ ] Use **always** `h0eff = h0 * (eta⁴ + 6 eta² + 1)`, which reduces to `h0` when `cosi=0` (default case).


As for re-scaling according to the time duration... we could keep this method only to do that?

pyfstat/make_sfts.py
Rodrigo-Tenorio: This is one of those places where v4 and v5 differ.

MFDv4 parses parameters using `XLALRegisterUvarMember`, asking for `float` (well, `double`)
```
  XLALRegisterUvarMember(transientWindowType,  STRING, 0, OPTIONAL, "Type of transient signal window to use. ('none', 'rect', 'exp').");
  XLALRegisterUvarMember(  transientStartTime,   REAL8, 0, OPTIONAL, "GPS start-time 't0' of transient signal window.");
  XLALRegisterUvarMember(  transientTauDays,     REAL8, 0, OPTIONAL, "Timescale 'tau' of transient signal window in days.");
   ```
 while we use MFDv5 in such a way that the parsing is done through a `cff` file, which ends up being read in `XLALReadPulsarParams`
 ```
    // ---------- transientWindow_t ----------
 
   // ----- type
   char *transientWindowType = NULL; BOOLEAN have_transientWindowType;
   XLAL_CHECK ( XLALReadConfigSTRINGVariable ( &transientWindowType, cfgdata, secName, "transientWindowType", &have_transientWindowType ) == XLAL_SUCCESS, XLAL_EFUNC );
   // ----- t0
   UINT4 transientStartTime = 0; BOOLEAN have_transientStartTime;
   XLAL_CHECK ( XLALReadConfigUINT4Variable ( &transientStartTime, cfgdata, secName, "transientStartTime", &have_transientStartTime ) == XLAL_SUCCESS, XLAL_EFUNC );
   // ----- tau (still keeping deprecated Days variant for backwards compatibility, for now)
   UINT4 transientTau = 0; BOOLEAN have_transientTau;
   XLAL_CHECK ( XLALReadConfigUINT4Variable ( &transientTau, cfgdata, secName, "transientTau", &have_transientTau ) == XLAL_SUCCESS, XLAL_EFUNC );
   REAL8 transientTauDays = 0; BOOLEAN have_transientTauDays;
   XLAL_CHECK ( XLALReadConfigREAL8Variable ( &transientTauDays, cfgdata, secName, "transientTauDays", &have_transientTauDays ) == XLAL_SUCCESS, XLAL_EFUNC );
 ```
where `transientTau` is `UINT4`. 

Funnily enough, `transientTauDays` is still supported, for now :wink:.

File: pyfstat/make_sfts.py (hunk 12 above)
Comment (Rodrigo-Tenorio) [reply pyfstat/make_sfts.py:1026]: Need some clarification on this: I use the same name in order to override the method.

**Review by Rodrigo-Tenorio — COMMENTED** [review]
1. Refactored `run_makefakedata` into two methods. Further changes may become futile, as they would involve keeping track of executable options in some sort of class. I propose postponing that to a future PR once we discuss how we want to approach lalapps executables.

2. Docstrings improved. Once the naming is decided, I'll finish it up.

File: tests.py (hunk 2 above)
Comment (dbkeitel) [reply tests.py:69]: let's leave it for future cleanup

File: pyfstat/make_sfts.py (hunk 4 above)
Comment (dbkeitel) [reply pyfstat/make_sfts.py:951]: I think I like `LineWriter` best, as the others are more technically precise, but this is established jargon that most active CWers will get immediately, and much shorter.

Also leaves open the possibility of generalizing it.

File: pyfstat/make_sfts.py (hunk 6 above)
Comment (dbkeitel) [reply pyfstat/make_sfts.py]: This comment now seems to be orphaned, but if we use `LineWriter` as per above, a similar sentence would fit in well again.

File: pyfstat/make_sfts.py (hunk 8 above)
Comment (dbkeitel) [reply pyfstat/make_sfts.py:987]: I find it confusing that this is supposed to be called after initialising the class. Can you clarify your intended use case? One makes a single instance, then reuses it for multiple injections with different cosi/duration but fixed effective strength...?

File: pyfstat/make_sfts.py (hunk 12 above)
Comment (dbkeitel) [reply pyfstat/make_sfts.py:1040]: You're right, it's already labeled as public on the main `Writer` class. I don't really see why, it should normally only be called once from `_basic_setup()`, so I'd say let's make it private?

Comment (Rodrigo-Tenorio) [reply pyfstat/make_sfts.py:1040]: If we let it public, the user can see how it's computed from the documentation.

That could be useful, just to have it upfront.

File: pyfstat/make_sfts.py [hunk 14]
```diff
@@ -742,6 +742,10 @@ def make_data(self, verbose=False):
         self.run_makefakedata()
 
     def run_makefakedata(self):
+        mfd_command_line = self.build_MFD_command_line()
+        self._execute_MFD(mfd_command_line)
```
dbkeitel, line 746: Actually it seems you ended up only needing to override `build_MFD_command_line` for the new class, so the contents of `_execute_MFD` could just be moved back into this one?

pyfstat/make_sfts.py
```diff
@@ -750,8 +754,7 @@ def run_makefakedata(self):
         and else runs the actual generation code.
         """
```
dbkeitel, line 755: this docstring should go back into `run_makefakedata()`

pyfstat/make_sfts.py
```diff
@@ -742,6 +742,10 @@ def make_data(self, verbose=False):
         self.run_makefakedata()
 
     def run_makefakedata(self):
+        mfd_command_line = self.build_MFD_command_line()
+        self._execute_MFD(mfd_command_line)
+
+    def build_MFD_command_line(self):
```
dbkeitel, line 748: `_build_MFD_command_line`? Or is there a good reason for users to call it directly?

pyfstat/make_sfts.py
```diff
@@ -943,6 +948,170 @@ def __init__(
...
+
+    def calculate_fmin_Band(self):
+        # MakeFakeData_v4 *requires* fmin and Band to be given.
+        # This overriding prevents Writer.calculate_fmin_Band from
+        # not setting them due to self.noiseSFTs being truthy.
+        hide_noiseSFTs, self.noiseSFTs = self.noiseSFTs, None
+        super().calculate_fmin_Band()
+        self.noiseSFTs = hide_noiseSFTs
+
+    def build_MFD_command_line(self):
```
dbkeitel, line 1048: as above, `_build_MFD_command_line`?

pyfstat/make_sfts.py
Rodrigo-Tenorio: > then reuses it for multiple injections with different cosi/duration but fixed effective strength

No, no re-use. 

One makes a single instance with the intention of, say, using h0_eff instead of h0.

Instead of computing somewhere `h0_eff`,  one gives `cosi` and calls this method to make the class handle the conversion to `h0_eff`.

pyfstat/make_sfts.py
dbkeitel: Ok, let's leave it. Needs to be a proper `"""` docstring for this overridden version then, though.

pyfstat/make_sfts.py
dbkeitel: Then let's just use your `Use always h0eff = h0 * (eta⁴ + 6 eta² + 1), which reduces to h0 when cosi=0 (default case).` in `__init__`, but best with a comment in the docstring and a warning thrown that this is happening in the actual cosi!=0 case.

The `scale_by_duration` I'd really leave to the user, this is not a well-defined input behaviour.

pyfstat/make_sfts.py
Rodrigo-Tenorio: I like the jargon argument. I'll update docs accordingly.

Rodrigo-Tenorio: 89006f7

pyfstat/make_sfts.py
Rodrigo-Tenorio: Adopted into 89006f7

pyfstat/make_sfts.py
Rodrigo-Tenorio: 64edc85

pyfstat/make_sfts.py
Rodrigo-Tenorio: See 24246e6

pyfstat/make_sfts.py
Rodrigo-Tenorio: I though for a second to leave this as a backdoor to hijack the MFD command line, but that can be done as easily without both methods, so out.

pyfstat/make_sfts.py
Rodrigo-Tenorio: Done in 2f31d5c

pyfstat/make_sfts.py
Rodrigo-Tenorio: Same as before, intended backdoor. But unnecessary, so removed in 2f31d5c.

pyfstat/make_sfts.py
Rodrigo-Tenorio: 2f31d5c

tests.py
dbkeitel: @Rodrigo-Tenorio  The `_compare_corrected_powers` could be merged back into `test_cosi_scaling`, especially since it doesn't do the actual comparison itself (L538).

Apart from that, looks fine.

pyfstat/make_sfts.py
```diff
@@ -191,6 +191,8 @@ class Writer(BaseSearchClass):
     for more detailed help with some of the parameters.
     """
 
+    mfd = "lalapps_Makefakedata_v5"
```
dbkeitel, line 194:
```suggestion
    mfd = "lalapps_Makefakedata_v5"
    """The executable; can be overridden by child classes."""
```
(I was just checking the docs built from this branch and this shows up as a class attribute, so better have a short string on it.)

pyfstat/make_sfts.py
```diff
@@ -943,6 +945,163 @@ def __init__(
         )
 
 
+class LineWriter(Writer):
+    """Inject an line into SFT data.
```
dbkeitel, line 949:
```suggestion
    """Inject a simulated line-like detector artifact into SFT data.
```

pyfstat/make_sfts.py
```diff
@@ -943,6 +945,163 @@ def __init__(
...
+
+    A (transient) line is defined as a constant amplitude and constant excess power artifact in the data.
+
+    In practice, it corresponds to a CW without Doppler or antenna-patern-induced amplitude modulation.
+
+    This functionality is implemented via `lalapps_MakeFakeData_v4`'s `lineFeature` option.
+    This version of MFD only supports one interferometer at a time.
```
dbkeitel, line 956:
```suggestion
    NOTE: This functionality is implemented via `lalapps_MakeFakeData_v4`'s `lineFeature` option.
    This version of MFD only supports one interferometer at a time.
    
    NOTE: All signal parameters except for `h0` and `cosi` will be ignored.
    `cosi` is used to rescale `h0` with the usual `sqrt(cosi**4 + 6*cosi**2 + 1)` relation.
```

pyfstat/make_sfts.py
```diff
@@ -943,6 +945,163 @@ def __init__(
...
+    """
+
+    mfd = "lalapps_Makefakedata_v4"
```
dbkeitel, line 959:
```suggestion
    mfd = "lalapps_Makefakedata_v4"
    """The executable (older version that supports the `--lineFeature` option)."""
```

pyfstat/make_sfts.py
```diff
@@ -943,6 +945,163 @@ def __init__(
...
+
+    required_mfd_line_parameters = [
+        "Freq",
+        "phi0",
+        "h0",
+        "cosi",
+        "transientWindowType",
+        "transientStartTime",
+        "transientTau",
+    ]
```
dbkeitel, line 969:
```suggestion
    ]
    """Other signal parameters will be removed before passing to MFDv4."""
```

pyfstat/make_sfts.py
```diff
@@ -943,6 +945,163 @@ def __init__(
...
+        Take care of minor inconsistencies between MFD_v4 and MFD_v5
+        """
+        super()._parse_args_consistent_with_mfd()
+
+        if any(
+            key not in self.required_mfd_line_parameters
+            for key in self.signal_parameters
+        ):
+            logging.warning(
+                "Transient lines only uses the following parameters:\n"
```
dbkeitel, line 996:
```suggestion
                "Injection of line artifacts only uses the following parameters:\n"
```

dbkeitel (changes requested):

pyfstat/make_sfts.py
dbkeitel: diff outdated

dbkeitel: ...or not...

dbkeitel (approved):

PR #285, opened 2021-02-22
integration.yml: add flake8-docstrings and flake8-executable

dbkeitel:
- [flake8-docstrings](https://pypi.org/project/flake8-docstrings/): pep257 compatibility, seems we already auto-fulfil this thanks to black
- [flake8-executable](https://pypi.org/project/flake8-executable/): checks executable flags and shebangs. We'll fail this one.


dbkeitel:
- [ ] ./setup.py:1:1: EXE001 Shebang is present but the file is not executable.
- [ ] ./examples/mcmc_vs_grid_simple_example/PyFstat_example_simple_mcmc_vs_grid_comparison.py:1:1: EXE001 Shebang is present but the file is not executable.
- [ ] ./examples/binary_examples/PyFstat_example_binary_mcmc_vs_grid_comparison.py:1:1: EXE001 Shebang is present but the file is not executable.
- [ ] ./examples/transient_examples/PyFstat_example_short_transient_grid_search.py:1:1: EXE001 Shebang is present but the file is not executable.
- [ ] ./examples/transient_examples/PyFstat_example_make_data_for_long_transient_search.py:1:1: EXE001 Shebang is present but the file is not executable.
- [ ] ./examples/transient_examples/PyFstat_example_short_transient_MCMC_search.py:1:1: EXE001 Shebang is present but the file is not executable.
- [ ] ./examples/transient_examples/PyFstat_example_make_data_for_short_transient_search.py:1:1: EXE001 Shebang is present but the file is not executable.
- [ ] ./examples/transient_examples/PyFstat_example_long_transient_MCMC_search.py:1:1: EXE001 Shebang is present but the file is not executable.
- [ ] ./pyfstat/core.py:0:1: EXE002 The file is executable but no shebang is present.

8     EXE001 Shebang is present but the file is not executable.
1     EXE002 The file is executable but no shebang is present.

dbkeitel:
@Rodrigo-Tenorio 

- [ ] any opinions on setup.py specifically?

- [ ] binder for examples doesn't mind either way, right?

- [ ] what's the right syntax to add the plugins to [.pre-commit-config.yaml](https://github.com/PyFstat/PyFstat/blob/15efc50305c84d6deb2fa0ef4247facb4163a016/.pre-commit-config.yaml)?

Rodrigo-Tenorio:
- I always execute python scripts using `python <script>`, just to make sure I use the right python.
- It shouldn't.

Rodrigo-Tenorio (approved):
Fine with this

Rodrigo-Tenorio (dismissed):
Why is `core.py` empty?

.pre-commit-config.yaml
```diff
@@ -7,3 +7,7 @@ repos:
     rev: 3.8.3
     hooks:
     - id: flake8
+      additional_dependencies: [
+          'flake8-docstrings==1.6.0',
+          'flake8-executable==2.1.1',
+      ]
```
Rodrigo-Tenorio, line 13: This should be enough, as plugins run by default after installation (according to the docstrings).

Pinning and updating the right version will be left for another PR, probably following changes in #288

dbkeitel:
> Why is `core.py` empty?

The diff says `100755 → 100644` which is fine and when I click on "view file", it's not empty: https://github.com/PyFstat/PyFstat/blob/22e9c08d6c9adc73e91fd3d179bd7eeae0c0b97b/pyfstat/core.py

Must be a weird web interface glitch.

dbkeitel:
Rebased on top of master in the hope that resets the diff view, but it still claims "empty file". I guess what it means is really **empty diff**, because I only changed the _mode_ of the file, not the content. Should be good to go.

Issue #284, opened 2021-02-21
LineWriter cache is broken


Rodrigo-Tenorio:
`LineWriter`'s cache is broken, due to the way in which `cosi` is applied.

Probably the best way of fixing this is following @dbkeitel initiall advice "Let the user take care of it", but I'll think about it for a while.

Also, `_basic_setup` still asks for `Alpha` and `Delta` keys, even though these are not used at all.

PR #286, opened 2021-02-23
Fix line writer cache

Rodrigo-Tenorio:
Here's the reason for the current test to fail: ~Executables are different~
```
/home/rodrigo/opt/miniconda3/envs/pyfstat-conda/bin/lalapps_Makefakedata_v4 --outSingleSFT=TRUE --outSFTbname="TestData/H-4_H1_1800SFT_testing_cache-700000000-7200.sft" --IFO="H1" --ephemEarth="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/earth00-40-DE405.dat.gz" --ephemSun="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/sun00-40-DE405.dat.gz" --startTime=700000000 --duration=7200 --fmin=29.96404307059555 --Band=0.07191313888519033 --Tsft=1800 --window="tukey" --tukeyBeta=0.001 --h0=5 --cosi=0 --phi0=0 --Freq=30 --noiseSqrtSh=1 --lineFeature=TRUE --randSeed=42

lalapps_Makefakedata_v4 --lineFeature=TRUE --outSingleSFT=TRUE --outSFTbname="TestData/H-4_H1_1800SFT_testing_cache-700000000-7200.sft" --IFO=H1 --noiseSqrtSh="1" --window="tukey" --tukeyBeta=0.001 --startTime=700000000 --duration=7200 --fmin=29.96404307059555 --Band=0.07191313888519033 --Tsft=1800 --h0 5.0 --Freq 30.0 --phi0 0 --cosi=0 --ephemEarth="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/earth00-40-DE405.dat.gz" --ephemSun="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/sun00-40-DE405.dat.gz" --randSeed=42
```

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -45,7 +45,7 @@ default_Writer_params = {
     "Band": None,
 }
 
-default_signal_params = {
+default_signal_params_no_sky = {
     "F0": 30.0,
     "F1": -1e-10,
     "F2": 0,
@@ -53,8 +53,11 @@ default_signal_params = {
     "cosi": 0,
     "psi": 0,
     "phi": 0,
-    "Alpha": 5e-1,
-    "Delta": 1.2,
+}
+
+default_signal_params = {
+    **default_signal_params_no_sky,
+    **{"Alpha": 5e-1, "Delta": 1.2},
 }
 
 
@@ -503,6 +506,7 @@ class TestWriter(BaseForTestsWithData):
 class TestLineWriter(TestWriter):
     label = "TestLineWriter"
     writer_class_to_test = pyfstat.make_sfts.LineWriter
+    signal_parameters = default_signal_params_no_sky
     multi_detectors = "H1"
 
     def test_multi_ifo_fails(self):
--- a/tests.py
+++ b/tests.py
@@ -520,6 +520,9 @@ class TestLineWriter(TestWriter):
                 **self.signal_parameters,
             )
 
+    def test_makefakedata_usecached(self):
+        pass
+
     def _check_maximum_power_consistency(self, writer, return_line_power=False):
         times, freqs, data = pyfstat.helper_functions.get_sft_array(writer.sftfilepath)
         max_power_freq_index = np.argmax(data, axis=0)
@@ -563,35 +566,6 @@ class TestLineWriter(TestWriter):
         writer.make_data(verbose=True)
 
         self._check_maximum_power_consistency(writer)
-
-    def test_cosi_scaling(self):
-        signal_params = default_signal_params.copy()
-        writer = self.writer_class_to_test(
-            outdir=self.outdir,
-            **default_Writer_params,
-            **signal_params,
-            **default_transient_params,
-        )
-
-        writer.make_data()
-        basic_power = self._check_maximum_power_consistency(
-            writer, return_line_power=True
-        )
-
-        signal_params["cosi"] = 1.0
-        writer = self.writer_class_to_test(
-            outdir=self.outdir,
-            **default_Writer_params,
-            **signal_params,
-            **default_transient_params,
-        )
-        writer.make_data()
-        corrected_power = self._check_maximum_power_consistency(
-            writer, return_line_power=True
-        )
-        self.assertTrue(
-            np.allclose(corrected_power / basic_power, np.sqrt(8), rtol=0, atol=1e-2)
-        )
 
 
 class TestWriterOtherTsft(TestWriter):
--- a/tests.py
+++ b/tests.py
@@ -521,7 +521,28 @@ class TestLineWriter(TestWriter):
             )
 
     def test_makefakedata_usecached(self):
-        pass
+
+        # Make everything from scratch
+        writer = self.writer_class_to_test(
+            outdir=self.outdir,
+            **default_Writer_params,
+            **default_signal_params,
+            **default_transient_params,
+        )
+        writer.make_data(verbose=True)
+        first_time = os.path.getmtime(writer.sftfilepath)
+
+        # Re-run, and should be unchanged
+        writer.make_data(verbose=True)
+        second_time = os.path.getmtime(writer.sftfilepath)
+        self.assertTrue(first_time == second_time)
+
+        # third run: touch the .cff to force regeneration
+        time.sleep(1)  # make sure timestamp is actually different!
+        os.system("touch {}".format(writer.config_file_name))
+        writer.run_makefakedata()
+        third_time = os.path.getmtime(writer.sftfilepath)
+        self.assertFalse(first_time == third_time)
 
     def _check_maximum_power_consistency(self, writer, return_line_power=False):
         times, freqs, data = pyfstat.helper_functions.get_sft_array(writer.sftfilepath)
--- a/tests.py
+++ b/tests.py
@@ -238,21 +238,29 @@ class TestWriter(BaseForTestsWithData):
             os.remove(self.Writer.config_file_name)
         if os.path.isfile(self.Writer.sftfilepath):
             os.remove(self.Writer.sftfilepath)
+
         # first run: make everything from scratch
         self.Writer.make_cff(verbose=True)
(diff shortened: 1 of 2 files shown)


Rodrigo-Tenorio:
@dbkeitel Any idea about how to circumvent that?

tests.py
```diff
@@ -458,6 +458,30 @@ def test_multi_ifo_fails(self):
...
+        second_time = os.path.getmtime(writer.sftfilepath)
+        self.assertTrue(first_time == second_time)
+
+        # third run: touch the .cff to force regeneration
+        time.sleep(1)  # make sure timestamp is actually different!
+        os.system("touch {}".format(writer.config_file_name))
+        writer.run_makefakedata()
+        third_time = os.path.getmtime(writer.sftfilepath)
+        self.assertFalse(first_time == third_time)
+
```
Rodrigo-Tenorio, line 484: This is the problematic test, where executables are different for some reason.

Rodrigo-Tenorio: Actual command line calls being compared. Seems like the order of the arguments is different...?

```
15:04 INFO    : Here is the CL from the old one:
lalapps_Makefakedata_v4 --outSingleSFT=TRUE --outSFTbname="TestData/H-4_H1_1800SFT_test-700000000-7200.sft" --IFO="H1" --ephemEarth="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/earth00-40-DE405.dat.gz" --ephemSun="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/sun00-40-DE405.dat.gz" --startTime=700000000 --duration=7200 --fmin=29.96404307059555 --Band=0.07191313888519033 --Tsft=1800 --window="tukey" --tukeyBeta=0.001 --h0=5 --cosi=0 --phi0=0 --Freq=30 --noiseSqrtSh=1 --lineFeature=TRUE --randSeed=42
15:04 INFO    : Here is the CL from the one to be created:
lalapps_Makefakedata_v4 --lineFeature=TRUE --outSingleSFT=TRUE --outSFTbname="TestData/H-4_H1_1800SFT_test-700000000-7200.sft" --IFO=H1 --noiseSqrtSh="1" --window="tukey" --tukeyBeta=0.001 --startTime=700000000 --duration=7200 --fmin=29.96404307059555 --Band=0.07191313888519033 --Tsft=1800 --h0 5.0 --Freq 30.0 --phi0 0 --cosi=0 --ephemEarth="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/earth00-40-DE405.dat.gz" --ephemSun="/home/rodrigo/gravity/venv/pyfstat/share/lalpulsar/sun00-40-DE405.dat.gz" --randSeed=42
```

Rodrigo-Tenorio: Apparently, [XLALUserVarGetLog](https://lscsoft.docs.ligo.org/lalsuite/lal/_user_input_8c_source.html#l01217) is used to get such a log before putting it into the SFT, and the order seems to be consistent with that specified in [MFDv4's XLALInitUserVars](https://lscsoft.docs.ligo.org/lalsuite/lalapps/makefakedata__v4_8c.html#a66b8a22a0d05442022448b06c66a5512).

So I think I'll simply re-sort the way in which arguments are concatenated.
Also, it looks like that was the exact approach taken in other MFDv4-based classes, such as `FrequencyModulatedArtifactWriter`.

Rodrigo-Tenorio: @dbkeitel I'll need a yes/no from you on this last point.

dbkeitel: I have no clear memory of that and would need to dig into the code. But I'm pretty sure that somewhere we already have a set-based, i.e. order-independent, solution to it...

dbkeitel: `helper_functions.match_commandlines()`, is that not used here?

Rodrigo-Tenorio: You are right, `helper_functions.match_commandlines`, thanks.

Rodrigo-Tenorio: Paying actual attention:
Old cl: `--h0=5  --Freq=30 --phi0=0 --noiseSqrtSh=1`
New cl: `--h0 5.0 --Freq 30.0 --phi0 0 --noiseSqrtSh="1"`

tests.py
```diff
@@ -564,35 +588,6 @@ def test_noise_sfts(self):
 
         self._check_maximum_power_consistency(writer)
 
-    def test_cosi_scaling(self):
```
dbkeitel, line 567: Why is this one removed?

tests.py
```diff
@@ -520,6 +520,30 @@ def test_multi_ifo_fails(self):
                 **self.signal_parameters,
             )
 
+    def test_makefakedata_usecached(self):
```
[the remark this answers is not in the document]
Rodrigo-Tenorio: I think the only difference is that it uses `self.writer` instead of `self.Writer`, so I can use LineWriter there.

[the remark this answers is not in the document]
dbkeitel: isn't that what we have `writer_class_to_test` for?

tests.py
Rodrigo-Tenorio: This is a vestige of when I considered including `cosi` into the whole mixture
in order to specify `h0eff` rather than `h0` for the line amplitude. 
But then we somehow concluded this was overly complicated, so we just removed the whole
functionality altogether.

tests.py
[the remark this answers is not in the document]
Rodrigo-Tenorio: Yes... although apparently not for `test_makefakedata_usecached`: https://github.com/PyFstat/PyFstat/blob/82fa81f30aa4db21d15558b75500245eedbf00ec/tests.py#L233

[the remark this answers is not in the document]
Rodrigo-Tenorio: Because for some reason (probably historical) this method is using the `Writer` created during start up, rather than generating a new one for this specific test.

tests.py
dbkeitel: Makes sense.

tests.py
```diff
@@ -520,6 +520,30 @@ def test_multi_ifo_fails(self):
...
+        writer.make_data(verbose=True)
+        second_time = os.path.getmtime(writer.sftfilepath)
+        self.assertTrue(first_time == second_time)
+
+        # third run: touch the .cff to force regeneration
+        time.sleep(1)  # make sure timestamp is actually different!
+        os.system("touch {}".format(writer.config_file_name))
+        writer.run_makefakedata()
+        third_time = os.path.getmtime(writer.sftfilepath)
+        self.assertFalse(first_time == third_time)
```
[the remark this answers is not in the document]
dbkeitel: Ok, so just keep overriding the method and not worry about any duplication, and it should work?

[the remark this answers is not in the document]
Rodrigo-Tenorio: Currently working on it. I'll override the method and hack a final check after it. Not beautiful, but will test properly.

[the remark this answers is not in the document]
dbkeitel: Ah sorry, I thought we were only talking about _test_ class methods. But it's about the cff logic of the main writer class.

Looks to me like the clean solution to make the main `Writer`'s behaviour more adaptable by children would be:
* in `_basic_setup` only store `self.config_file_name` if `injectionSources` is set
* change logic in `check_cached_data_okay_to_use` correspondingly

dbkeitel (approved):
Approved once the pipeline is happy.

Closes #284 when merged.

tests.py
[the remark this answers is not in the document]
dbkeitel: Your solution is equivalent, because `make_cff` is only called, and hence `os.path.isfile(self.config_file_name)` only becomes true, if we actually do an injection. :+1:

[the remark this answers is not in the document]
Rodrigo-Tenorio: Done differently, everyone agrees.

PR #291, opened 2021-03-15
Add explicit link to run_all_examples.py

### LINKED ISSUE #287: document run_all_examples.py
suggestion from @khanx169 over on the JOSS review (https://github.com/openjournals/joss-reviews/issues/3000#issuecomment-792902860):

> I've gone through the review checklist and everything looks good to me. I've one minor suggestion: In the Github repo, there's a run_all_examples.py script. The docs point to GitHub, but it might be good to also explicitly mention this script there (especially since some examples need to be executed in order).

> I was thinking specifically of the Examples section of readthedocs. I understand most users would probably want to look at some specific example at a time, but my thought process when I reached that section was: "It'd be nice if I could run all the examples first, and then come back and look at the scripts and outputs one by one to familiarize myself with the package". I suspect occasionally some odd user might think like this as well. Currently, the cumulative runtime for all examples is not terrible, but if more examples are added in the future, perhaps interactive notebooks (with the expected output for each cell already displayed) would be more useful.

This should be quick to do. cc @Rodrigo-Tenorio

Rodrigo-Tenorio:
Closes #287


dbkeitel (changes requested):
Actually, if the user hasn't cloned the repository, just downloading that one file won't get them far...  

I don't think github provides download links per directory without weird external tools, so I think we have to recommend the three options

- git clone
- download full tree from Zenodo
- download full tree from github

Ideally, versioned docs should point to the matching Zenodo version and "latest" docs to github... but I suspect the effort would be disproportionate.

dbkeitel (approved):

dbkeitel:
@khanx169 is this close enough to what you had in mind? Output page is here: https://pyfstat--291.org.readthedocs.build/en/291/examples.html

khanx169:
> @khanx169 is this close enough to what you had in mind? Output page is here: https://pyfstat--291.org.readthedocs.build/en/291/examples.html

Yes, that looks good. Thanks!

Issue #294, opened 2021-03-19
ASCL


dbkeitel:
We got an "Astrophysics Source Code Library" entry generated for us: https://ascl.net/2102.027 (Thanks to their editor, Alice Allen.) Two follow-up housekeeping tasks:

We should quote that ID
- [x]  in the README
- [x]  and JOSS paper.

- [x] While we already have a standard license file and citation info in the Readme, they suggest using a separate file for citation info, which seems like a good idea:

> The ASCL recommends that code authors assign a license to their software and make it clear how use of the code should be cited, preferably by use of a CITATION.md, CITATION.cff, or Codemeta.json file added to the root of your code repo. Editable examples of the latter two can be generated by adding /CITATION.cff or /codemeta.json to the link provided above for your code’s entry. Should you choose to use the ASCL entry for citation, a suggested format is available online.

dbkeitel:
#295 adds README badge and codemeta.json file

PR #295, opened 2021-03-26
add codemeta.json and ASCL badge

dbkeitel:
- see #294

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -2,7 +2,8 @@
     "@context": "https://doi.org/10.5063/schema/codemeta-2.0",
     "@type": "SoftwareSourceCode",
     "name": "PyFstat: Continuous gravitational-wave data analysis",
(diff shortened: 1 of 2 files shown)


codemeta.json
```diff
@@ -0,0 +1,58 @@
+{
+    "@context": "https://doi.org/10.5063/schema/codemeta-2.0",
+    "@type": "SoftwareSourceCode",
+    "name": "PyFstat: Continuous gravitational-wave data analysis",
+    "description": "PyFstat performs F-statistic-based continuous gravitational wave (CW) searches and other CW data analysis tasks. It is built on top of the LALSuite library, its functionality more accessible through a Python interface. It also provides MCMC-based followup of promising candidates from wide-parameter-space searches.",
```
Rodrigo-Tenorio, line 5:
```suggestion
    "description": "PyFstat performs F-statistic-based continuous gravitational wave (CW) searches and other CW data analysis tasks. It is built on top of the LALSuite library, making its functionality more accessible through a Python interface. It also provides an MCMC-based followup of promising candidates from wide-parameter-space searches.",
   
```

dbkeitel:
There seem to be a lot of optional fields that one could add to this json file and all sorts of external services that one can use it for, but for now let's just have a minimal version and be done with it.

Issue #290, opened 2021-03-11
code coverage


dbkeitel:
from @RobertRosca over on the JOSS review (https://github.com/openjournals/joss-reviews/issues/3000#issuecomment-792999393):

>  Codecov could be included as part of the CI, ran it locally and the coverage (with just the tests.py file) is around 62%
    Also provides coverage details for PRs, letting you know if a contribution lowers the test coverage
    Going back to the examples point, if the examples were ipynb files with the output saved in them then nbval could check that the output has not changed, and executing the examples would probably bump the coverage up a decent amount

Indeed we've been planning to do this for a while, I just never got around to reading the full setup documentation. But I know it's a very popular solution, so should be straightforward.

dbkeitel:
@RobertRosca there's a basic working setup for this now in #296. Do you happen to know how we can exclude files from the calculation when using a github action setup? https://docs.codecov.io/docs/ignoring-paths applies to a setup with a top-level `codecov.yml` file but instead I went with a `.github/workflows/codecov.yml` action config, which while also yml seems to require a different structure.

* Should I just add that other file with the recommended syntax as well and the github action will know to get configuration options from it?
* Or would I do something like the following inside the action's own yml:
```
    - name: Generate coverage report
      with:
        ignore:
          - "setup.py"
          - "versioneer.py"
      run: |
        pip install pytest
        pip install pytest-cov
        pytest tests.py --cov=./ --cov-report=xml
```

RobertRosca:
Hi, sorry for the late reply, I've been very busy recently and didn't have much time to look through my open source stuff.

I noticed that the PR is already merged, did you end up solving this issue? Part of the problem is that you do `cov=./`, which will try to calculate the coverage for all the files in the repository. If you instead do `cov=pyfstat` then it will only compute coverage for the files that are in the module itself (e.g. anything that is imported when you do `import pyfstat...`, so not `setup.py` or `versioneer.py`).

dbkeitel:
Thanks @RobertRosca just changing the top level is a good idea! I'll make another PR for that. The reports also seem more useful if there is already a base version on master, so it's good to do this incrementally.

dbkeitel:
[...]

PR #296, opened 2021-03-26
codecov

dbkeitel:
- starting from https://github.com/codecov/codecov-action#example-workflowyml-with-codecov-action
- to address #290

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -34,7 +34,7 @@ jobs:
       run: |
         pip install pytest
         pip install pytest-cov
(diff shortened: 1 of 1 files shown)


dbkeitel:
Seems we're at about 60% out of the box. The bot's "Continue to review full report at Codecov" link above won't work, because it goes to a diff against master, where there is no report yet of course. Here's the commit-level full report: https://codecov.io/gh/PyFstat/PyFstat/commit/09370bc9e19ac085acf609176c2b8d74a3c24c0e

Rodrigo-Tenorio (approved):

PR #297, opened 2021-03-26
run_examples.yml: remove external gridcorner dependency

dbkeitel:
- should no longer need external gridcorner dependency now that we have an internal copy of it (introduced in #227 )


dbkeitel:
action running: https://github.com/PyFstat/PyFstat/actions/runs/689372475

PR #300, opened 2021-04-01
codecov.yml: only run on pyfstat package dir

dbkeitel:
- as suggest by @RobertRosca in #290


dbkeitel:
Hmm, it seems the report is not getting regenerated when adding new commits to a PR...

dbkeitel:
> Hmm, it seems the report is not getting regenerated when adding new commits to a PR...

..because the syntax was invalid. Reverting the "ignore" attempt and just doing teh basedir change for now.

PR #307, opened 2021-04-07
Simplify if-mkdir block to makedirs-exist_ok

### LINKED ISSUE #306: Replace `os.mkdir(*)` by `os.makedirs(*, exist_ok=True)`
Some search classes (`grep` finds three instances in `grid_based_searches.py` and 4 in `mcmc_based_searches.py`) use the following snippet to create output directories 
```
if os.path.isdir(outdir) is False:
    os.mkdir(outdir)
```
The main issue with this approach is that `os.mkdir` cannot create intermediate folders.

A simple solution would be to substitute the aforementioned `if`-block by the simpler 
```
os.makedirs(outdir, exist_ok=True)
```
This option 1) is able to properly handle multiple sub-folders 2) handles existing folders nicely and 3) is shorter.

Rodrigo-Tenorio:
Closes #306

### DIFF
--- a/pyfstat/grid_based_searches.py
+++ b/pyfstat/grid_based_searches.py
@@ -115,8 +115,7 @@ class GridSearch(BaseSearchClass):
         """
 
         self._set_init_params_dict(locals())
-        if os.path.isdir(outdir) is False:
-            os.mkdir(outdir)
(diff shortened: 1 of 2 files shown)


dbkeitel (approved):
Thanks, certainly an improvement!

PR #308, opened 2021-04-07
tweaks to corner plots

### LINKED ISSUE #184: axis label auto-scaling issues

Rodrigo-Tenorio:
I attempted to improve titles by using `g` format, but it's tricky.

This PR will close #184

### DIFF
--- a/pyfstat/mcmc_based_searches.py
+++ b/pyfstat/mcmc_based_searches.py
@@ -780,23 +780,24 @@ class MCMCSearch(BaseSearchClass):
 
         labels = []
         for key in self.theta_keys:
-            label = None
-            s = self.symbol_dictionary[key]
-            s.replace("_{glitch}", r"_\mathrm{glitch}")
-            u = self.unit_dictionary[key]
-            if key in self.transform_dictionary:
-                if "symbol" in self.transform_dictionary[key]:
-                    s = self.transform_dictionary[key]["symbol"]
-                if "label" in self.transform_dictionary[key]:
-                    label = self.transform_dictionary[key]["label"]
-                if "unit" in self.transform_dictionary[key]:
-                    u = self.transform_dictionary[key]["unit"]
+
+            if key in self.transform_dictionary.keys():
+                val = self.transform_dictionary[key]
+                s, label, u = [
+                    val.get(slu_key, None) for slu_key in ["symbol", "label", "unit"]
+                ]
+            else:
+                label = None
+
             if label is None:
-                if newline_units:
-                    label = "{} \n [{}]".format(s, u)
-                else:
-                    label = "{} [{}]".format(s, u)
+                s = self.symbol_dictionary[key].replace(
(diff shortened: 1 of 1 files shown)


dbkeitel:
- [ ] In the example, `tp` is a bit confusing because it uses exponential notation for the median and normal notation for the error bars. No, wait actually, the problem is in using exponential notation for _**GPS time variables**_ in general. We need a special logic for those.

- [ ] Also, for the quantities _with_ units, the `[s] = ...` notation is weird but intuitively clear, but for the `asini` without units the `[] = ...` makes no sense at all. While you're at it, could you move the units bracket to the first line right behind the variable name, ideally also suppressing it if empty?

Rodrigo-Tenorio:
Quite more work is needed to tweak around with the numerical values. 
Let's leave this as is, colors and labels are better.

dbkeitel (changes requested):
We're back to the error bars potentially being shown as "0.00" then, which is bad. But having a bit of a look at `corner.py` internals, I think there's really nothing we can do about it on our end. We should consider sending them a PR implementing either
1. allowing `title_fmt` to be a dictionary over parameters
2. a "smart" mode that first applies `title_fmt` to the error bars, then formats the median to match significant figures

Before merging this one though, could you just add parameterst to `plot_corner()` to change both the main and the truths color please? I like the new slateblue-orange scheme as a default, but being able to make it all-black for printing would be nice too.

dbkeitel:
Someone's been thinking about similar lines for the quantiles header already, but no results yet: https://github.com/dfm/corner.py/issues/108

dbkeitel:
Actuallt, we **can** do it on our end: pass `title_fmt=None` to `corner.corner` and then add the titles ourselves after, reusing the code from https://github.com/dfm/corner.py/blob/main/src/corner/core.py#L212 . The "smart" way is a bit tricky to get right, so I'd just go with a dict of "good standard" formats for our known set of parameters (i.e. ".2g" for most, something custom for Tgps, anything else...?). This will mean repeating quantile computations, but to quote the corner.py code comment: "This might redo unneeded computation but who cares."

Do you want to give it a try inside this PR @Rodrigo-Tenorio or should we merge first and do it separately?

Rodrigo-Tenorio:
Ok, I agree with this:
- [ ] Add density color and truths color as an argument at PyFstat's level

I'll do it now and we can merge this PR.

See #311 for discussion about the title.

Rodrigo-Tenorio:
Both arguments can be already accessed through `kwargs`. I'd prefer not to extend that plate of conditional spaghetti.

Color scheme is set to black-gray.

@dbkeitel PR is go.

dbkeitel (approved):
Ok, one kwarg to rule them all!

PR #310, opened 2021-04-08
Add return statement to get_semicoherent_twoF()

### LINKED ISSUE #309: `get_semicoherent_twoF` doesn't return anything
[SemiCoherentSearch's get_semicoherent_twoF](https://github.com/PyFstat/PyFstat/blob/a01668d56af784f7ce7d6f94c7d39864f4febafe/pyfstat/core.py#L1751) lacks a `return`.

This isn't problematic per se, as the result can be accessed through the `twoF`attribute, but should be fixed.

Rodrigo-Tenorio:
Closes #309

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1814,6 +1814,8 @@ class SemiCoherentSearch(ComputeFstat):
         if record_segments:
             self.twoF_per_segment = twoF_per_segment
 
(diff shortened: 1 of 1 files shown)


dbkeitel (approved):
Makes sense and will then be consistent with other detection statistic functions.

Issue #88, opened 2020-07-14
Improve installation/test of Docker images


Rodrigo-Tenorio:
In the same line as #85 , it would make sense to check/test the installed version of PyFstat inside Docker containers.

As of now, Docker images take PyFstat directly from repo; that means "tests.py" is available and one could install pytest there. However, this approach breaks the whole point of containers (nothing else than what you need) and ignores any  test done to a PyPi release.

Given that we accept the existence of two container "families" (latest containers and version-tagged containers), my suggestion would be to install PyFstat differently in each family:

1) Latest family: This family is more likely to be "broken" than the other one. One should *not* use them to run production jobs, since reproducibility may be compromised later in time. Hence, it's reasonable to install from source rather than PyPi and run pytest to check its proper functioning.

2) Version-tagged: This will be the actual, official, recommended way to run reproducible production jobs. It should install pyfstat from PyPi and check whether the python packages were properly installed.

This will probably imply re-formatting Dockerfile into a folder with several Dockerfiles. Nothing critical.

- [ ] Test *latest* contianers using pytest.
- [ ] Make version-tagged containers install from PyPi.
- [ ] Check version-tagged containers after installing packages.

dbkeitel:
:+1:  :+1:  :+1:  to all 3 suggestions

Rodrigo-Tenorio:
This can be also applied to the current [WIP] binder implementation: Keeping a container with the dependencies from `.binder/environment.yml`.

And, why not, another `dev` container to use in CI pipelines!

Rodrigo-Tenorio:
[...]

PR #312, opened 2021-04-09
Replace requirements for extras require

Rodrigo-Tenorio:
Following @RobertRosca 's suggestion in https://github.com/openjournals/joss-reviews/issues/3000#issuecomment-792999393
> Pretty subjective, but I think having all the extra requirements under extras_require would be nice, then you could have some for docs, style, test, etc... which makes development and contribution a bit easier, also makes the CI scripts cleaner since everything can be a pip install ".[style,test,...]" command instead of using separate requirements files

ditch `requirements-dev.txt` and `requirements.txt` in favor of `setup.py`'s `install_requires`and `extras_require`.

- [x] Include `extras_require` in `setup.py` and remove `requirements-dev.txt`.
- [x] Implement the usage of `extras_require` into the CI action.
- [x] Remove unnecessary usage of `requirements.txt` in Dockerfile.
- [x] Remove `requirements.txt` (after all, why should we keep it?).
- [x] Update bin/setup-dev-tools.sh
- [x] Document the usage of `extras_require` in the README.

Edit:

- [x] Include `readthedocs` dependencies into `extras_require` as well.


dbkeitel:
Makes sense as far as the extras go. (Do you have a docs link handy for how this is supposed to be done canonically?)

But I think `requirements.txt` is required for conda-forge...? Though I'll have to double-check the docs for that, too.

Rodrigo-Tenorio:
Here are the docs for extra dependencies: https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html

Essentially, if the user wants the extra dependencies labeled as `dev`: `pip install pyfstat[dev]`.

Rodrigo-Tenorio:
@dbkeitel I see no `requirements.txt` in  https://github.com/PyFstat/staged-recipes/blob/PyFstat/recipes/PyFstat/meta.yaml ; in fact, it looks like dependencies are hard-coded there (I seem to remember conda not updating dependencies by itself?).

dbkeitel:
> I see no requirements.txt in https://github.com/PyFstat/staged-recipes/blob/PyFstat/recipes/PyFstat/meta.yaml ; in fact, it looks like dependencies are hard-coded there (I seem to remember conda not updating dependencies by itself?).

You're probably right - I remembered they needed to be tracked separately, but probably not the right place. Let's try getting rid of it and seeing if the release workflow breaks anywhere.

.github/workflows/codecov.yml
```diff
@@ -19,21 +19,10 @@ jobs:
         python-version: 3.8
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
-        pip install -r requirements.txt
```
dbkeitel, line 23: I love that you got rid of the multiple copies of the ephemerides setup! Rest of the workflow cleanup also looks good. But why removing the `python -m pip install --upgrade pip` here and in the others, that seems a reasonable precaution given how often pip deprecates and randomly breaks?

README.md
```diff
@@ -170,25 +170,34 @@ pip install -r /PATH/TO/THIS/DIRECTORY/requirements.txt
...
+PyFstat manages optional dependencies through setuptool's `extras_require`.
+
+Available sets of optional dependencies are:
+
+* [chainconsumer](https://github.com/Samreay/ChainConsumer): Required to run some optional 
+plotting methods and some of the [example scripts](./examples).
+* [pycuda](https://pypi.org/project/pycuda/): Required for the `tCWFstatMapVersion=pycuda`
+  option of the `TransientGridSearch` class. (Note: Installing `pycuda` requires a working 
+  `nvcc` compiler in your path.)
+* style: Includes the [flake8](https://flake8.pycqa.org/en/latest/) linter
```
dbkeitel, line 184: use ``style`` and similar highlighting for all the tags to make clear they're the actual named tags and not just part of the README?

dbkeitel (changes requested):
Thanks @Rodrigo-Tenorio that looks like a great clean-up. We'll just have to try if it all works out in the release workflow. I just left two small comments.

.github/workflows/codecov.yml
Rodrigo-Tenorio: You are right, let's keep it.

README.md
Rodrigo-Tenorio: I though links couldn't be highlighted. I was [`wrong`](https://www.youtube.com/watch?v=ROGOCFpcj2Q) :wink:

Rodrigo-Tenorio:
@dbkeitel I've retracted the removal of `requirements.txt`since that's the way docker containers are currently built.

I propose to merge this and post-pone the removal of `requirements.txt`to whenever #88 is done.

dbkeitel:
Does requirements.txt need to go back into the manifest then too, or does docker not care about that?

Rodrigo-Tenorio:
> Does requirements.txt need to go back into the manifest then too, or does docker not care about that?

I was wrong: Docker installs the package by copying the repo and using pip (`pip install .`). 
I got confused by the `pip install -r requirements.txt` line, which is completely unnecessary if `setup.py` is properly set.
 
That line (which was kind of ubiquitous) must have bin a vestige from an ancient time.

Rodrigo-Tenorio:
@dbkeitel  Ready to merge

dbkeitel (approved):

Issue #142, opened 2020-08-26
Check plotting options...?


Rodrigo-Tenorio:
As seen in #135 and #141 , plotting options should have a check done at some point.

Since most of them are basically used for diagnosis (as one would tailor a proper figure using the results), I'm hesitant to spend too much time in this issue.

@dbkeitel Opinion? This will probably become irrelevant if we ever interface Bilby.

dbkeitel:
By "check" you mean adding calls to the test suite? Of course it's hard/impossible to automatically test if the plots are really sensible, but at least testing that they don't crash under various inputs would be good.

Rodrigo-Tenorio:
Sort of  yes: If we include them into the test suite we will at least be aware of any breaking changes.

dbkeitel:
Any PRs welcome, and we can also keep this open as a reminder issue for a good while.

I'll just adapt the labels so it doesn't sound like there's a specific known bug at the moment.

dbkeitel:
Basic coverage should be there with #315 and #320 now. Any major ones I've missed @Rodrigo-Tenorio ?

PR #315, opened 2021-04-12
tests: cover MCMC plotting methods

dbkeitel:
- we obviously don't test what's actually shown in the figures,
   but just that the calls don't fail
 - skipping MCMCTransientSearch for now
   due to calculate_twoF_cumulative() failure
 - refs #142
 - let's see how bad runtime becomes and reduce to a few
   test cases if needed

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -1379,6 +1379,12 @@ class BaseForMCMCSearchTests(BaseForTestsWithData):
             if assertTrue:
                 self.assertTrue(within)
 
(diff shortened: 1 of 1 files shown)


dbkeitel:
Tests only slowed down by a minute or so, and +3.32% coverage. :+1:

dbkeitel:
I'll merge this one first and put the fix to #316 on a separate PR.

PR #317, opened 2021-04-12
Fix cumulative plots for transients, extend examples and tests

### LINKED ISSUE #316: plot_cumulative_max() fails for MCMCTransientSearch
Adding a call to `plot_cumulative_max()` for transient searches in #315 leads to:
```
FAILED tests.py::TestMCMCTransientSearch::test_transient_MCMC_t0_tau - TypeError: calculate_twoF_cumulative() got an unexpected keyword argument 'transient_tstart'
```

dbkeitel:
Followup to #315, fixes #316 and extends support for transients in CFS/PFS cumulative functions in general.

Notes:
1. Main change for non-transient applications is that the MCMC cumulative plots will now be saved to a file by default, overriding the behaviour of the core function but better matching all other MCMC plotting methods.
2. For `calculate_twoF_cumulative()` it may be cleaner to refactor it with some kwarg logic, but this will do for now.
3. The addition of a `transientWindowType` override parameter to `get_predict_fstat_parameters_from_dict()` is also a hack, and could have just done it on the level of the examples calling that function, but since it's just a "helper", I think this is fine for now.

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -898,12 +898,13 @@ def get_dictionary_from_lines(lines, comments, raise_error):
     return d
 
 
-def get_predict_fstat_parameters_from_dict(signal_parameters):
+def get_predict_fstat_parameters_from_dict(signal_parameters, transientWindowType=None):
     """Extract a subset of parameters as needed for predicting F-stats.
 
     Given a dictionary with arbitrary signal parameters,
     this extracts only those ones required by `helper_functions.predict_fstat()`:
     Freq, Alpha, Delta, h0, cosi, psi.
+    Also preserves transient parameters, if included in the input dict.
 
     Parameters
     ----------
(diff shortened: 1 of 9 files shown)


pyfstat/helper_functions.py
```diff
@@ -912,16 +913,30 @@ def get_predict_fstat_parameters_from_dict(signal_parameters):
...
         The dictionary of selected parameters.
     """
-    predict_fstat_params = {
-        key: signal_parameters[key]
-        for key in ["F0", "Alpha", "Delta", "h0", "cosi", "psi"]
+    required_keys = ["F0", "Alpha", "Delta", "h0", "cosi", "psi"]
+    transient_keys = {
+        "transientWindowType": "transientWindowType",
+        "transient_tstart": "transientStartTime",
+        "transient_duration": "transientTau",
```
Rodrigo-Tenorio, line 932: Interestingly enough, these keys don't match the ones in `BaseSearchClass.translate_keys_to_lal`.

Rodrigo-Tenorio (approved):

pyfstat/helper_functions.py
dbkeitel: I noticed that too. Those are for CFSv2, this is for PFS, which don't agree with each other.

PR #320, opened 2021-04-12
TestGridSearch: basic plots coverage

dbkeitel:
More test coverage.

 - refs #142

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -1901,9 +1901,33 @@ class TestGridSearch(BaseForTestsWithData):
     Band = 0.5
     BSGL = False
 
-    def test_grid_search(self):
-        search = pyfstat.GridSearch(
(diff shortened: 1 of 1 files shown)


dbkeitel:
review request @Rodrigo-Tenorio mostly for a quick look over whether any of the `search` -> `self.search` replacements look fishy

Rodrigo-Tenorio (commented):
In principle I'm Ok with this.

Question: Is it really necessary to add `search` as an attribute?
Since this adds a method which already requires a particular argument, I don't see why we couldn't just pass `search` to said method as well.

If the required argument in `_test_plots` could be skipped, we could simply add a decorator to the methods checking for a plot...

Rodrigo-Tenorio (approved):

PR #325, opened 2021-04-29
setup.py : add project_urls for PyPI sidebar

dbkeitel:
Discovered this nice feature after seeing black's changelog link: https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls

@Rodrigo-Tenorio any other important things to link?


Rodrigo-Tenorio (approved):
That looks right for  now, nice

PR #326, opened 2021-04-29
Add black badge

Rodrigo-Tenorio:
Let's pay `black` the respect it deserves.


dbkeitel (approved):
Attempt at black humor: the badge could be more black.

But yes, let's use it.

PR #328, opened 2021-06-04
Sphinx v4

Rodrigo-Tenorio:
Sphinx was updated to v4, but we are still on 3.4 with some pinned versions.

I'll check whether we can unpin or what.


dbkeitel:
Readthedocs recommends to *always* pin to a known-to-be-working set of sphinx and dependency versions (first orange warning box on https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html ), so could you check which versions we're currently getting with the unpinned pipeline and then re-pin to those?

Rodrigo-Tenorio:
For some reason I don't understand, `style` and `test` items in the Optional dependencies of the README seem not to work as expected when converted into sphinx.

Apart from that, this is done.

dbkeitel:
> For some reason I don't understand, `style` and `test` items in the Optional dependencies of the README seem not to work as expected when converted into sphinx.

Maybe they're some type of reserved keywords...?

Rodrigo-Tenorio:
> > For some reason I don't understand, `style` and `test` items in the Optional dependencies of the README seem not to work as expected when converted into sphinx.
> 
> Maybe they're some type of reserved keywords...?

Not sure, it's kind of an unregular behaviour in the html. The README.md works fine, though.

20210604_15h18m00s_grim

dbkeitel (approved):
I don't see any syntax typos in the README, so let's chalk the minor formatting issues there up to random sphinx weirdness for now and get this merged.

PR #329, opened 2021-06-14
Pump Python3.8 to Python3.9 in Docker container

Rodrigo-Tenorio:
Python is dead, long live Python.


dbkeitel (approved):
3.8 certainly ain't dead yet, still on extended life support for many years I hope. ;)

PR #330, opened 2021-07-05
Add max mismatch

Rodrigo-Tenorio:
Ad interface to modify the maximum-allowed mismatch by CFSv2

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -403,6 +403,7 @@ class ComputeFstat(BaseSearchClass):
         self.init_computefstatistic()
         self.output_file_header = self.get_output_file_header()
         self.get_det_stat = self.get_fullycoherent_detstat
+        self.allowedMismatchFromSFTLength = allowedMismatchFromSFTLength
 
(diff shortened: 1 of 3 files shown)


Rodrigo-Tenorio:
I can't make it work in a semicoherent example, for some reason

dbkeitel:
Why is the default `None` in some place and `0` in others?

Rodrigo-Tenorio:
> Why is the default `None` in some place and `0` in others?

My mistake, I accidentally committed some debugging changes.

All should be None.

dbkeitel (changes requested):
Apart from the 0->None change this already looks good, test cases included. (Codecov complains that the line in `MCMCGlitchSearch` is not covered, but let's keep that for a putative "loop test cases over variant classes" general update.) Also I haven't checked carefully if any other classes should be updated, but that can be done separately if needed.

dbkeitel (approved):
Defaults look good now, and examples action passed: https://github.com/PyFstat/PyFstat/actions/runs/1008001453

PR #334, opened 2021-10-15
add example minimal yml to repo

dbkeitel:
- example environment recipe as described in
   https://github.com/PyFstat/PyFstat/wiki/conda-environments


dbkeitel:
> I agree with you though, how about pyfstat_conda.yml? Basically, pyfstat.yml could be anything unrelated to conda.

That's why it's in a `conda` dir, but let's do the official thing then.

PR #345, opened 2021-11-26
add timestampsFiles option to Writer

dbkeitel:
Wrapping MFDv5's corresponding feature, with an extra setup method to make sure we also extract and store `self.tstart` and `self.duration` for other methods which require them.

Not entirely sure if any other assumptions of the module will break when given gappy non-equal timestamps for multiple detectors, which the test suite doesn't yet cover.


pyfstat/make_sfts.py
```diff
@@ -319,6 +323,12 @@ def __init__(
         randSeed: int or None
             Optionally fix the random seed of Gaussian noise generation
             for reproducibility.
+        timestampsFiles: str or None
+            Comma-separated list of per-detector timestamps files
+            (simple text files, lines with <seconds> <nanoseconds>).
```
Rodrigo-Tenorio, line 328:
```suggestion
            (simple text files, lines with <seconds> <nanoseconds>,
            comments should use `%`).
```

pyfstat/make_sfts.py
```diff
@@ -469,8 +517,22 @@ def _basic_setup(self):
...
+        incompatible_with_TS = ["tstart", "duration", "noiseSFTs"]
+        TS_required_options = ["Tsft", "detectors"]
+        no_noiseSFTs_options = ["tstart", "duration", "Tsft", "detectors"]
+        if getattr(self, "timestampsFiles", None) is not None:
+            if np.any([getattr(self, k) is not None for k in incompatible_with_TS]):
+                raise ValueError(
+                    "timestampsFiles option is incompatible with"
+                    f" ({','.join(incompatible_with_TS)})."
+                )
+            if np.any([getattr(self, k) is None for k in TS_required_options]):
```
Rodrigo-Tenorio, line 529:
```suggestion
            if np.any([getattr(self, k, None) is None for k in TS_required_options]):
```

pyfstat/make_sfts.py
```diff
@@ -480,7 +542,7 @@ def _basic_setup(self):
             self._get_setup_from_noiseSFTs()
         elif np.any([getattr(self, k) is None for k in no_noiseSFTs_options]):
```
Rodrigo-Tenorio, line 543:
```suggestion
        elif np.any([getattr(self, k, None) is None for k in no_noiseSFTs_options]):
```

Rodrigo-Tenorio (dismissed):

PR #346, opened 2021-11-29
Solve numpy deprecation error

Rodrigo-Tenorio:
Should be simple enough

### DIFF
--- a/.github/workflows/test_lalsuite_dev.yml
+++ b/.github/workflows/test_lalsuite_dev.yml
@@ -1,0 +1,36 @@
+name: LALSuite-dev test
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  '42 9 * * 1'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
+    env:
+      LALPULSAR_DATADIR: ${{ github.workspace }}/ephem
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install package and development dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e ${GITHUB_WORKSPACE}[style,test,wheel]
+        pip install --upgrade --pre lalsuite
+    - name: Linting with flake8
+      run: |
+        flake8 --count --statistics .
+    - name: Style check with black
+      run: |
+        black --check --diff .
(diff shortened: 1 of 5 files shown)


Rodrigo-Tenorio:
As per the following:
```
In [1]: import numpy as np

In [2]: a = [np.array([i]) for i in np.random.random(5)]

In [3]: a
Out[3]: 
[array([0.86926741]),
 array([0.22876801]),
 array([0.09083037]),
 array([0.48390722]),
 array([0.74860603])]

In [4]: np.hstack(a)
Out[4]: array([0.86926741, 0.22876801, 0.09083037, 0.48390722, 0.74860603])

In [5]: %timeit np.hstack(a)
2.76 µs ± 12.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: np.concatenate(a)
Out[6]: array([0.86926741, 0.22876801, 0.09083037, 0.48390722, 0.74860603])

In [7]: %timeit np.concatenate(a)
1.23 µs ± 20.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Seems like I should be using `np.concatenate`.

dbkeitel:
Thanks! Some "real-life" testing with `get_fullycoherent_single_IFO_twoFs()` seems to imply the `np.concatenate()` solution actually is faster than the old plain array assignment, which is interesting... But let's get #347 merged first to make it easier to set up a proper test case, then compare a rebased version of this one against master cleanly, and then merge it.

dbkeitel:
Rebase on top of #347, now
```
import numpy as np
from timeit import timeit
import pyfstat
tstart = 700000000
duration = 4 * 1800
tref = tstart
detectors = "H1,L1"
sqrtSX = 1
F0 = 30
F1 = 0
F2 = 0
Alpha = 0
Delta = 0

search = pyfstat.ComputeFstat(
            tref=tref,
            minStartTime=tstart,
            maxStartTime=tstart + duration,
            detectors=detectors,
            injectSqrtSX=np.repeat(sqrtSX,len(detectors.split(","))),
            minCoverFreq=F0 - 0.1,
            maxCoverFreq=F0 + 0.1,
            singleFstats=True,
        )
FS = search.get_fullycoherent_twoF(F0, F1, F2, Alpha, Delta)
%timeit search.get_fullycoherent_single_IFO_twoFs()
```
is about 50% faster on my system with this change than on master 86260740142f6abb7a8babcb5aec9cf0ae82eb3f : something like `2.26 µs ± 187 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)` against `3.26 µs ± 137 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)` (with 2nd digit fluctuations between re-runs.) Nice bonus along the way!

dbkeitel (approved):

Issue #260, opened 2021-01-15
optional BSGL and transient feature enhancements


dbkeitel:
Extracted list of "things not done" from #245, none of these are crucial for current requirements but would all be "nice to have":

- [ ] enable BSGL for glitch search classes; put in an explicit error message instead
- [ ] adding explicit user options for the BSGL tuning parameters (note that MCMC classes already have a `rhohatmax` option which could be reused, as that happens to be an equivalent parameter to BSGL's `Fstar0`)
- [ ] adding stricter quantitative tests for BSGL numerical values
- [ ] changing the choices for how "transient BSGL" is done, which may not be the best currently
- [ ] finishing up the implementation and output of single-IFO F-stats on multi-IFO searches - currently these are only computed when `BSGL` option is set, and only exported from grid- not from MCMC searches. A followup should compute and export them always when Ndet>=2 (as they're essentially for free).
- [ ] further work on transient detection statistic choices - we should also wrap the PGM11 transient marginalized Bayes factor

PR #347, opened 2021-11-29
add separate singleFstats option to all ComputeFstat classes

dbkeitel:
In testing #346 I noticed that the handling of single-detector F-stats (`twoFX`) didn't really make sense. It was impossible to get them without setting `BSGL`, and the corresponding function didn't fail cleanly if one tried anyway.

 - add separate singleFstats option to all ComputeFstat classes
   (core classes and grid-based searches; no MCMC classes seem to currently make use of these attributes)
 - BSGL implies this
 - but can also use it on its own
 - always pre-initialise self.twoFX
   but only set FSTATQ_2F_PER_DET
   if self.singleFstats
 - get_fullycoherent_single_IFO_twoFs()
   now fails if not self.singleFstats

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1023,7 +1023,7 @@ class ComputeFstat(BaseSearchClass):
         if not self.transientWindowType:
             if self.singleFstats:
                 self.get_fullycoherent_single_IFO_twoFs()
-            if self.BSGL is False:
+            if not self.BSGL:
                 return self.twoF
             self.get_fullycoherent_log10BSGL()
             return self.log10BSGL
@@ -1785,7 +1785,7 @@ class SemiCoherentSearch(ComputeFstat):
 
         if self.singleFstats:
             self.get_semicoherent_single_IFO_twoFs(record_segments)
-        if self.BSGL is False:
+        if not self.BSGL:
             return self.twoF
         else:
             return self.get_semicoherent_log10BSGL()
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -369,6 +369,7 @@ class ComputeFstat(BaseSearchClass):
         randSeed : int or None
             random seed for on-the-fly noise generation using `injectSqrtSX`.
             Setting this to 0 or None is equivalent; both will randomise the seed,
+            following the behaviour of XLALAddGaussianNoise(),
             while any number not equal to 0 will produce a reproducible noise realisation.
         assumeSqrtSX : float or list or str
             Don't estimate noise-floors but assume this (stationary) single-sided PSD.
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -271,6 +271,7 @@ class ComputeFstat(BaseSearchClass):
         search_ranges=None,
         injectSources=None,
         injectSqrtSX=None,
+        randSeed=None,
         assumeSqrtSX=None,
         SSBprec=None,
         RngMedWindow=None,
@@ -365,6 +366,8 @@ class ComputeFstat(BaseSearchClass):
             List or comma-separated string: must match len(detectors)
             and/or the data in sftfilepattern.
             Detectors will be paired to list elements following alphabetical order.
+        randSeed : int or None
+            random seed for on-the-fly noise generation using `injectSqrtSX`.
         assumeSqrtSX : float or list or str
             Don't estimate noise-floors but assume this (stationary) single-sided PSD.
             Single float or str value: use same for all IFOs.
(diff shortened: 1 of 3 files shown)


pyfstat/core.py
```diff
@@ -1012,9 +1021,10 @@ def get_fullycoherent_detstat(
             F0, F1, F2, Alpha, Delta, asini, period, ecc, tp, argp
         )
         if not self.transientWindowType:
+            if self.singleFstats:
+                self.get_fullycoherent_single_IFO_twoFs()
             if self.BSGL is False:
```
Rodrigo-Tenorio, line 1026:
```suggestion
            if not self.BSGL:
```

pyfstat/core.py
```diff
@@ -1768,10 +1783,11 @@ def get_semicoherent_det_stat(
             F0, F1, F2, Alpha, Delta, asini, period, ecc, tp, argp, record_segments
         )
 
+        if self.singleFstats:
+            self.get_semicoherent_single_IFO_twoFs(record_segments)
         if self.BSGL is False:
```
Rodrigo-Tenorio, line 1788:
```suggestion
        if not self.BSGL:
```

pyfstat/grid_based_searches.py
```diff
@@ -131,7 +131,7 @@ def __init__(
     def _set_output_keys(self):
         self.output_keys = self.search_keys.copy()
         self.output_keys.append("twoF")
-        if hasattr(self.search, "twoFX"):
+        if self.search.singleFstats:
```
Rodrigo-Tenorio, line 134: Let me check if I got this right: `singleFstats` is the boolean label.

Rodrigo-Tenorio (changes requested):
See inline comments. Overall fine

pyfstat/grid_based_searches.py
dbkeitel: Yes. I named it for parallelism with CFSv2's `outputSingleFstats`, but dropping the verb since we're not necessarily outputting anything here...  
By contrast, the results are `self.twoFX`.

This whole mess could be reconsidered by making the attributes more hierarchical, e.g. having two dicts `self.results` and `self.options`. But I guess that falls under "PyFstat 2.0".

Rodrigo-Tenorio: Yes

dbkeitel:
Added test coverage by expanding the existing `TestComputeFstat::test_get_fully_coherent_BSGL` and on the way also
* added `randSeed` to `ComputeFstat` to make the `noiseSqrtSX` on-the-fly data generation reproducible
* store BSGL-related quantities to `self` (long-term these need to be come user variables anyway, see #260)

dbkeitel:
...now it might actually pass...

pyfstat/core.py
```diff
@@ -626,6 +632,10 @@ def init_computefstatistic(self):
             ] = self.injectSqrtSX
         else:
             FstatOAs.injectSqrtSX = lalpulsar.FstatOptionalArgsDefaults.injectSqrtSX
+        if hasattr(self, "randSeed") and self.randSeed is not None:
+            FstatOAs.randSeed = self.randSeed
+        else:
+            FstatOAs.randSeed = lalpulsar.FstatOptionalArgsDefaults.randSeed
```
Rodrigo-Tenorio, line 638:
```suggestion
        FstatOAs.randSeed = (
              getattr(self, "randSeed", None) or lalpulsar.FstatOptionalArgsDefaults.randSeed
       )
```


I agree with this. I tried to simplify it, but I *don't* want to prevent 0 from being used as a proper seed.
For the sake of it, I'll leave it as a suggestion.

Q: Is there any reason not to use 0 as a seed in a random number generator?

Rodrigo-Tenorio (dismissed):
Looks fine for now.

pyfstat/core.py
dbkeitel: Actually that's a good point: I just checked and the corresponding CFSv2 help is `pecify random-number seed for reproducible noise (0 means use /dev/urandom for seeding).`, and indeed that's what the low-level code does: https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_generate_pulsar_signal_8c_source.html#l01002

So there actually wouldn't be a problem if we interpret a user-given `0` as "give me the default" at the same level as a `None`, and our documentation should be explicit about that too.

dbkeitel: See latest commit.

Rodrigo-Tenorio: Added a bit more verbosity to the docstring. Cool!

Rodrigo-Tenorio (approved):
Just included an explicit reference into the docstring pointing to XLALAddGaussianNoise().

dbkeitel:
Thanks. Whitespace fixed, merging once linter is happy.

Issue #356, opened 2021-12-08
simplify the ephemerides mess


dbkeitel:
With https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1729 and https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1746 the next lalpulsar release will finally include a minimal set of ephemeris files also in the pip-installable wheels from PyPI :tada: and an improved logic for lal to resolve default file paths.

We can adapt to this in two steps:
- [x] update our `get_ephemeris_files()` to deprecate the hacky `$LALPULSAR_DATADIR` solution and use the new fallback -> this will be backwards-compatible to the current releases, #357
- [x] once the release is actually out, update the README, remove the `get-and-export-ephemeris.sh` script and simplify the actions setups: #358

In the second step, also the explicit options in `test_grid_search_against_CFSv2` can be removed; for now this would still fail as CFSv2 so far actually relies on the separate `$LAL_DATA_PATH` variable instead, which we never set.

- [ ] remove the deprecated `$LALPULSAR_DATADIR` support in a future release

dbkeitel:
@Rodrigo-Tenorio I'm ready to remove the deprecated bits for 1.14.0. Is the old style stuff in https://github.com/PyFstat/PyFstat/blob/master/Dockerfile still needed? Should I just remove it from there too, and afterwards we can test if it works with the default path resolution?

PR #357, opened 2021-12-08
Simplify ephems handling part 1

dbkeitel:
Addressing the first half of #356:

 - deprecate `$LALPULSAR_DATADIR` solution
 - fall back to standard `[earth/sun]00-40-DE405.dat.gz` file names
   and relying on lal's path resolution
   instead of "None"
 - this should now work reliably on conda and pip installs
   after https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1746

As long as not actually removing the `$LALPULSAR_DATADIR` hack from both the code and the actions setup, this step should be backwards-compatible for now.


Rodrigo-Tenorio (approved):
Discussed offline, sounds good.

PR #358, opened 2021-12-08
Simplify ephems handling part 2

dbkeitel:
Second part of #356, this is expected to fail until there's a new lal/lalpulsar release.


dbkeitel:
@Rodrigo-Tenorio lalsuite release is out, so once this passes all integration tests + [conda test](https://github.com/PyFstat/PyFstat/actions/runs/1578126330) + [examples](https://github.com/PyFstat/PyFstat/actions/runs/1578127010) we can merge.

dbkeitel:
- [ ] lal 7.1.5 and lalpulsar 3.1.1 don't seem to have made it to conda-forge yet (but lalapps 7.4.0 has)
- [ ] examples action fails to find ephemerides

dbkeitel:
>  examples action fails to find ephemerides

Had forgotten to remove the environment variable from `run_examples.yml`. Should now work: https://github.com/PyFstat/PyFstat/actions/runs/1578218865

dbkeitel:
To avoid conflicts, I've added a commit here on this branch to include 3.10 in the tests, now that lalsuite 7.2 on PyPI supports it. Leaving the conda action at a single 3.9 run for the moment, though in principle that could be matrixified too.

dbkeitel:
@Rodrigo-Tenorio looks like we never resolved this after conda-forged got eventually updated. I just reran a set of jobs end everything passed:
* regular integration tests with PyPI stable versions: https://github.com/PyFstat/PyFstat/actions/runs/1752644329
* with PyPI dev versions: https://github.com/PyFstat/PyFstat/actions/runs/1752644795
* with conda-forge versions: https://github.com/PyFstat/PyFstat/actions/runs/1752645503
 
Codecov is only unhappy about the now deprecated environment variable branch of get_ephemeris_files().

Any comments before merging?

README.md
```diff
@@ -207,59 +209,45 @@ A minimal configuration line to use would be e.g.:
...
-`source $MYLALPATH/etc/lalsuite-user-env.sh`,
-then the ephemerides path should be automatically picked up from
-the `$LALPULSAR_DATADIR` environment variable.
-Similarly, if you have installed lalsuite from conda-forge,
-it should come with ephemerides included and properly set up.
-
-Alternatively, you can place a file
-`~/.pyfstat.conf` into your home directory which looks like
-
+by either creating a `~/.pyfstat.conf` file in your home directory which looks like
```
Rodrigo-Tenorio, line 226:
```suggestion
by creating a `~/.pyfstat.conf` file in your home directory which looks like
```

Rodrigo-Tenorio (commented):
So the final status of ephemeris files is as follows:

- We still keep a function under `helper_functions` which, if nothing is given, looks for a `~/.pyfstat.conf`file (likely to be deprecated in the future) or returns the **hardcoded name** of the default ephemeris files shipped with lalsuite.
 
Q on the lalsuite side: Do we still need to provide the ephemeris file name ? Or is it enough to give an "empty string" (`None`) and any lal* program will be smart enough to look for the shipped-by-default ephemeris file?

dbkeitel:
> Q on the lalsuite side: Do we still need to provide the ephemeris file name ? Or is it enough to give an "empty string" (None) and any lal* program will be smart enough to look for the shipped-by-default ephemeris file?

Goods point. How about we merge this first and then check on a separate branch if None-ing works?

Rodrigo-Tenorio (approved):

PR #364, opened 2021-12-10
Fix expected SFT names for narrowband (freq-segmented) noiseSFTS

### LINKED ISSUE #327: Predicted SFT names fail if more SFTs than required are parsed
`pyfstat.Writer` includes a sanity check which compares the output SFT's name produced by MFD to what one should expect according to the way in which this name is built.

`Makefakedata_v5` constructs the name once the SFTs are completely created: If one gives "too many SFTs" (expecting MFD to take care of selecting the required ones), that's fine, since the filename will be composed after selecting the appropriate files.

On the other hand, PyFstat creates the "predicted" filename *before* querying the appropriate frequency band. As a result, the number of SFTs included in the name is equal to the number of SFTs fed into MFD (regardless of its frequency band).

dbkeitel:
fixes #327 but will need MFDv5 follow-up bug report for the actual signal-injection-into-narrowband-SFTs case

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -826,6 +826,8 @@ class Writer(BaseSearchClass):
                 raise IOError(
                     f"It seems we successfully ran {self.mfd},"
                     f" but did not get the expected SFT file path(s): {self.sftfilepath}."
+                    f" What we have in the output directory '{self.outdir}' is:"
+                    f" {os.listdir(self.outdir)}"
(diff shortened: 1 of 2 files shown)


dbkeitel:
Codecov is unhappy about added error messages that don't get hit during tests, but we can ignore that until deciding to globally test failure modes.

Rodrigo-Tenorio (approved):

PR #368, opened 2021-12-20
refactor generate_loudest() and add to GridSearch

dbkeitel:
- add generate_loudest() method to GridSearch via new helper function
    
     - add helper_functions.generate_loudest_file()
     - make GridSearch and MCMCSearch wrap that
     - add to examples and tests
 
- improve grid examples
    
     - report max point with injection offsets
     - use inj dictionaries
     - inject negative f1dot to be more typical
     - call GridSearch with named parameters

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -1176,7 +1176,7 @@ def generate_loudest_file(
         "ephemEarth": earth_ephem,
         "ephemSun": sun_ephem,
     }
-    CFSv2_params.update({par: key for par, key in opt_params.items() if key})
+    CFSv2_params.update({key: val for key, val in opt_params.items() if val})
     cmd += " ".join([f"--{key}={val}" for key, val in CFSv2_params.items()])
 
     run_commandline(cmd, return_output=False)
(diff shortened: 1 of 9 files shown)


pyfstat/helper_functions.py
```diff
@@ -1105,3 +1105,79 @@ def parse_list_of_numbers(val):
...
+    }
+    CFSv2_params.update(max_params)
+    opt_params = {
+        "minStartTime": minStartTime,
+        "maxStartTime": maxStartTime,
+        "transient-WindowType": transientWindowType,
+        "ephemEarth": earth_ephem,
+        "ephemSun": sun_ephem,
+    }
+    CFSv2_params.update({par: key for par, key in opt_params.items() if key})
```
Rodrigo-Tenorio, line 1179:
```suggestion
    CFSv2_params.update({par: val for par, val in opt_params.items() if val})
```

Just bugged by using key as a value

Rodrigo-Tenorio (changes requested):
Just a minor variable issue, looks good overall!

pyfstat/helper_functions.py
dbkeitel: Good catch, but let me make it even more canonical:
```suggestion
    CFSv2_params.update({key: val for key, val in opt_params.items() if val})
```

Rodrigo-Tenorio (approved):

PR #371, opened 2022-02-14
Improve SFT to Numpy array conversion

Rodrigo-Tenorio:
This new function supports mulit-IFO and returns complex amplitudes rather than absolute values.

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -474,16 +474,16 @@ def get_sft_as_arrays(sftfilepattern, fMin=None, fMax=None, constraints=None):
     multi_sfts = lalpulsar.LoadMultiSFTs(sft_catalog, fMin, fMax)
     logging.info("done!")
 
-    frequencies = []
     times = {}
     amplitudes = {}
 
+    old_frequencies = None
     for ind, ifo in enumerate(ifo_labels.data):
 
         sfts = multi_sfts.data[ind]
 
         times[ifo] = np.array([sft.epoch.gpsSeconds for sft in sfts.data])
-        amplitudes[ifo] = np.array([np.abs(sft.data.data) for sft in sfts.data]).T
+        amplitudes[ifo] = np.array([sft.data.data for sft in sfts.data]).T
 
         nbins, nsfts = amplitudes[ifo].shape
 
@@ -491,13 +491,15 @@ def get_sft_as_arrays(sftfilepattern, fMin=None, fMax=None, constraints=None):
 
         f0 = sfts.data[0].f0
         df = sfts.data[0].deltaF
-        frequencies.append(np.linspace(f0, f0 + (nbins - 1) * df, nbins))
+        frequencies = np.linspace(f0, f0 + (nbins - 1) * df, nbins)
 
-    if not all(
-        np.allclose(frequencies[0], other_freqs) for other_freqs in frequencies[1:]
-    ):
-        raise ValueError("Frequencies don't match across detectors, weird...")
-    frequencies = np.array(frequencies[0])
+        if (old_frequencies is not None) and not np.allclose(
+            frequencies, old_frequencies
+        ):
+            raise ValueError(
+                f"Frequencies don't match between {ifo_labels.data[ind-1]} and {ifo}"
+            )
+        old_frequencies = frequencies
 
     return frequencies, times, amplitudes
 
@@ -532,7 +534,7 @@ def get_sft_array(sftfilepattern, F0=None, dF0=None):
 
         warnings.warn(
             "`get_sft_array` is deprecated and will be removed in a future release. "
-            "Please, use `get_sft_as_numpy_array` to load SFT complex amplitudes."
+            "Please, use `get_sft_as_arrays` to load SFT complex amplitudes."
         )
 
     if F0 is None and dF0 is None:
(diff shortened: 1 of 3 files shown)


pyfstat/helper_functions.py
```diff
@@ -429,6 +429,74 @@ def convert_array_to_gsl_matrix(array):
     return gsl_matrix
 
 
+def get_sft_as_numpy_array(sftfilepattern, fMin=None, fMax=None, constraints=None):
```
dbkeitel, line 432: This should be `get_sfts_as_numpy_array` or `get_sft_data_as_numpy_array`, as it handles multiple SFTs. (Note: the old `get_sft_array` is grammatically ok in this regard, as it can be read as "an array of SFT-type things", though on the other hand, technically it's a list of arrays...)

pyfstat/helper_functions.py
```diff
@@ -429,6 +429,74 @@ def convert_array_to_gsl_matrix(array):
...
+    """
+
+    Parameters
+    ----------
+    sftfilepattern: str
+            Pattern to match SFTs using wildcards (`*?`) and ranges [0-9];
+            multiple patterns can be given separated by colons.
+    fMin, fMax: float or None
+        Restrict frequency range to `[fMin, fMax]`.
+        If None, retreive the full frequency range.
```
dbkeitel, line 442:
```suggestion
        If None, retrieve the full frequency range.
```

pyfstat/helper_functions.py
```diff
@@ -429,6 +429,74 @@ def convert_array_to_gsl_matrix(array):
...
+
+    Returns
+    ----------
+    times: np.ndarray
```
dbkeitel, line 446: Order does not match actual returns, and is the actual one different on purpose from the old func?

pyfstat/helper_functions.py
dbkeitel: PS: Actually this new one doesn't really do "as numpy array" either, it does a dict of arrays... `get_sfts_as_numpy_arrays`, or, since we don't really use non-nump arrays much, just `get_sfts_as_arrays`?

pyfstat/helper_functions.py
```diff
@@ -429,6 +429,74 @@ def convert_array_to_gsl_matrix(array):
     return gsl_matrix
 
 
+def get_sft_as_numpy_array(sftfilepattern, fMin=None, fMax=None, constraints=None):
+    """
+
+    Parameters
```
dbkeitel, line 435: doc of `constraints` is missing

pyfstat/helper_functions.py
```diff
@@ -429,6 +429,74 @@ def convert_array_to_gsl_matrix(array):
...
+        nbins, nsfts = amplitudes[ifo].shape
+
+        logging.info(f"{nsfts} retrieved from {ifo}.")
+
+        f0 = sfts.data[0].f0
+        df = sfts.data[0].deltaF
+        frequencies.append(np.linspace(f0, f0 + (nbins - 1) * df, nbins))
+
+    if not all(
+        np.allclose(frequencies[0], other_freqs) for other_freqs in frequencies[1:]
```
dbkeitel, line 492: Can this be simplified to just check [-1], which is the just-appended one, against [0]? Or even better, never append anything, just make a temporary in-loop variable here and compare that against the already-stored one (if already stored)?

pyfstat/helper_functions.py
```diff
@@ -429,6 +429,74 @@ def convert_array_to_gsl_matrix(array):
...
+    frequencies = []
+    times = {}
+    amplitudes = {}
+
+    for ind, ifo in enumerate(ifo_labels.data):
+
+        sfts = multi_sfts.data[ind]
+
+        times[ifo] = np.array([sft.epoch.gpsSeconds for sft in sfts.data])
+        amplitudes[ifo] = np.array([np.abs(sft.data.data) for sft in sfts.data]).T
```
dbkeitel, line 481: I thought you wanted to _not_ take the abs?

pyfstat/helper_functions.py
```diff
@@ -454,6 +522,13 @@ def get_sft_array(sftfilepattern, F0=None, dF0=None):
         A 2D array of the absolute values of the SFT data
         in each frequency bin at each timestamp.
     """
+    if True:  # pragma: no cover
+        import warnings
+
+        warnings.warn(
+            "`get_sft_array` is deprecated and will be removed in a future release. "
+            "Please, use `get_sft_as_numpy_array` to load SFT complex amplitudes."
```
dbkeitel, line 530: to be updated if func renamed

dbkeitel (changes requested):

pyfstat/helper_functions.py
Rodrigo-Tenorio: Agree

pyfstat/helper_functions.py
Rodrigo-Tenorio: Time and frequency are interchanged to return  with the following signature: `1D, {1D per ifo}, {2D per ifo}`.
I think it makes more sense to split, between container type, but can revert if you feel otherwise.

pyfstat/helper_functions.py
Rodrigo-Tenorio: Typo. Thanks!

pyfstat/helper_functions.py
Rodrigo-Tenorio: Yes.

Rodrigo-Tenorio (commented):
Done if test don't fail.

dbkeitel (approved):

dbkeitel:
Looks good now, and thanks for cleaning up the tests's structure along the way!

PR #378, opened 2022-03-04
move injection parameter generators to new module, simplify Writer class

Rodrigo-Tenorio:
Self explanatory, as a middle step before taclking a clean up of make_sfts.py.


dbkeitel:
- [x] moving the generatos to a new `injection_parameters` module -> no objections, but since it's the main change, let's adapt the title of this PR
- [ ] but why the rename from `*Generator` to `Generate*`?
- [ ] also (but depends on answer to previous point) can we put in deprecated/defunct warnings when trying to still import from the old module?
- [x] getting rid of `_gps_to_int` -> no objections
- [ ] `tend` changes -> I'm not getting something here, isn't e.g. `self.Writer.tend - 2 * self.Writer.Tsft` now trying to subtract a number from a function..?

dbkeitel (commented):
- [x] moving the generatos to a new `injection_parameters` module -> no objections, but since it's the main change, let's adapt the title of this PR
- [x] but why the rename from `*Generator` to `Generate*`?
- [x] also (but depends on answer to previous point) can we put in deprecated/defunct warnings when trying to still import from the old module?
- [x] getting rid of `_gps_to_int` -> no objections
- [x] `tend` changes -> I'm not getting something here, isn't e.g. `self.Writer.tend - 2 * self.Writer.Tsft` now trying to subtract a number from a function..?

Rodrigo-Tenorio:
> tend changes -> I'm not getting something here, isn't e.g. self.Writer.tend - 2 * self.Writer.Tsft now trying to subtract a number from a function..?
 
 Example:
 ```
 >>> class A:
...     a = 42
...     def successor(self):
...             return self.a + 1
... 
>>> class PropA:
...     a = 42
...     @property
...     def successor(self):
...             return self.a + 1

 >>> A().successor()
43
>>> A().successor
<bound method A.successor of <__main__.A object at 0x7f2f39c6f6d0>>

>>> PropA().successor()
TypeError: 'int' object is not callable
>>> PropA().successor
43
```
 
i.e. using the `@property` decorator, in a sense (and this is just a specific way in which this can be used), makes a method behave like an attribute from the user side.

dbkeitel:
I completely missed the ` @property` being added in the diff view. Makes total sense.

dbkeitel (approved):

PR #379, opened 2022-03-06
Add Python API to compute SNR

Rodrigo-Tenorio:
(Must be merged after #378)

Implement a new module to compute SNRs using PyFstat. 
Incidentally, this also....
- Adds class to parse detectors states, which may come in handy for other methods.
- Allows to directly access the M matrix from PyFstat. 

To do:
- [x] Compute SNR using `assumeSqrtSX *fixed and common to all detectors*.
- [x] Compute SNR using weights.
- [x] Add derived quantities such as E[2F] and Var[2F].

Testing will consist in comparing against LALSuite's PFS via helper functions.

### DIFF
--- a/docs/source/pyfstat.rst
+++ b/docs/source/pyfstat.rst
@@ -41,6 +41,14 @@ pyfstat.helper\_functions module
    :undoc-members:
    :show-inheritance:
 
+pyfstat.injection\_parameters module
+------------------------------------
+
+.. automodule:: pyfstat.injection_parameters
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
 pyfstat.make\_sfts module
 -------------------------
 
@@ -61,6 +69,14 @@ pyfstat.optimal\_setup\_functions module
 ----------------------------------------
 
 .. automodule:: pyfstat.optimal_setup_functions
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+pyfstat.snr module
+------------------
+
+.. automodule:: pyfstat.snr
    :members:
    :undoc-members:
    :show-inheritance:
--- a/pyfstat/snr.py
+++ b/pyfstat/snr.py
@@ -11,22 +11,22 @@ from pyfstat.helper_functions import get_ephemeris_files
 @define(kw_only=True, slots=False)
 class SignalToNoiseRatio:
     """
-    Compute the optimal signal-to-noise ratio (and derived quantities)
+    Compute the optimal signal-to-noise ratio (and derived quantities, such as the F-statistic)
     of a CW signal on Gaussian noise or an arbitrary dataset of SFTs.
 
     The definition of SNR (shortcut for "optimal signal-to-noise ratio")
-    is taken from Eq. (76) of https://dcc.ligo.org/LIGO-T0900149 and is
+    is taken from Eq. (76) of https://dcc.ligo.org/T0900149-v6/public and is
     such that `<2F> = 4 + SNR^2`, where `<2F>` represents the expected
-    value over noise realization of twice the F-statistic of a prefectly
+    value over noise realizations of twice the F-statistic of a perfectly
     matched template to an existing signal in the data.
 
     Computing SNR^2 requires two quantities:
-        - The M matrix, which depends on the sky position and polarization angle
+        - The antenna pattern matrix `M`, which depends on the sky position and polarization angle
         and encodes the effect of the detector's antenna pattern over the course
         of the observing run.
-        - The JKS amplitude paramaters {A^0, A^1, A^2, A^3}, which are functions
-        of the CW's amplitude parameters (h0, cosi, psi) or, alternatively,
-        (aPlus, aCross, psi).
+        - The JKS amplitude parameters `{A^0, A^1, A^2, A^3}`, which are functions
+        of the CW's amplitude parameters `(h0, cosi, psi)` or, alternatively,
+        `(aPlus, aCross, psi)`.
 
     Parameters
     ----------
@@ -35,7 +35,7 @@ class SignalToNoiseRatio:
         Provides the required information to compute the antenna pattern contribution.
     noise_weights: Union[lalpulsar.MultiNoiseWeights, None]
         Optional, incompatible with `(assumeSqrtSX, Tsft)`.
-        Can be compute from SFTs using `SignalToNoiseRatio.from_sfts`.
+        Can be computed from SFTs using `SignalToNoiseRatio.from_sfts`.
         Vector of noise weights to account for a varying noise floor or unequal noise
         floors in different detectors.
     assumeSqrtSX: float
@@ -124,10 +124,10 @@ class SignalToNoiseRatio:
         """
         Compute the SNR^2 of a CW signal using XLALComputeOptimalSNR2FromMmunu.
         Parameters correspond to the standard ones used to describe a CW
-        (see e.g. Eqs. (16), (26), (30) of https://dcc.ligo.org/LIGO-T0900149).
+        (see e.g. Eqs. (16), (26), (30) of https://dcc.ligo.org/T0900149-v6/public ).
 
         Mind that this function returns *squared* SNR
-        (Eq. (76) of https://dcc.ligo.org/LIGO-T0900149),
+        (Eq. (76) of https://dcc.ligo.org/T0900149-v6/public ),
         which can be directly related to the expected F-statistic as
         ```
         <2F> = 4 + SNR^2.
@@ -241,7 +241,7 @@ class SignalToNoiseRatio:
 
     def _convert_amplitude_parameters(self, h0, cosi, aPlus, aCross):
         """
-        Internal method to check and convert the given amplitude parametrs
+        Internal method to check and convert the given amplitude parameters
         into the required format.
         """
         h0_cosi = h0 is not None and cosi is not None
@@ -273,7 +273,7 @@ class DetectorStates:
         time_stamps: array-like or dict
             GPS timestamps at which detector states will be retrieved.
             If array, use the same set of timestamps for all detectors,
-            which must be explicitly given by the user via`detectors`.
+            which must be explicitly given by the user via `detectors`.
             If dictionary, each key should correspond to a valid detector name
             to be parsed by XLALParseMultiLALDetector and the associated value
             should be an array-like set of GPS timestamps for each individual detector.
@@ -328,7 +328,7 @@ class DetectorStates:
         sft_constraint: lalpulsar.SFTConstraint
             Optional argument to specify further constraints in XLALSFTdataFind.
         return_sfts: bool
-            If True, also return the loaded SFTs. This is useful to compute futher
+            If True, also return the loaded SFTs. This is useful to compute further
             quantities such as noise weights.
 
         Returns
@@ -373,7 +373,7 @@ class DetectorStates:
             if detectors is not None:
                 raise ValueError("`timestamps`' keys are redundant with `detectors`. ")
 
-            logging.debug("Retrieving detectors from tiemstamps dictionary.")
+            logging.debug("Retrieving detectors from timestamps dictionary.")
             detectors = list(timestamps.key())
             timestamps = timestamps.values()
 
--- a/pyfstat/snr.py
+++ b/pyfstat/snr.py
@@ -118,6 +118,10 @@ class SignalToNoiseRatio:
 
 
 class DetectorStates:
+    """
+    Python interface to XLALGetMultiDetectorStates and XLALGetMultiDetectorStatesFromMultiSFTs.
+    """
+
     def __init__(self):
         self.ephems = lalpulsar.InitBarycenter(*get_ephemeris_files())
 
@@ -158,11 +162,42 @@ class DetectorStates:
         self,
         sftfilepath,
         central_frequency,
-        frequency_wing_bins=1,
         time_offset=0,
+        frequency_wing_bins=1,
         sft_constraint=None,
         return_sfts=False,
     ):
+        """
+        Parameters
+        ----------
+        sftfilepath: str
+            Path to SFT files in a format compatible with XLALSFTdataFind.
+        central_frequency: float
+            Frequency [Hz] around which SFT data will be retrieved.
+            This option is only relevant if further information is to be
+            retrieved from the SFTs (i.e. `return_sfts=True`).
+        time_offset: float
+            Timestamp offset to retrieve detector states.
+        frequency_wing_bins: int
+            Frequency bins around the central frequency to retrieve from
+            SFT data. Bin size is determined using the SFT baseline time
+            as obtained from the catalog.
+            This option is only relevant if further information is to be
+            retrieved from the SFTs (i.e. `return_sfts=True`).
+        sft_constraint: lalpulsar.SFTConstraint
+            Optional argument to specify further constraints in XLALSFTdataFind.
+        return_sfts: bool
+            If True, also return the loaded SFTs. This is useful to compute futher
+            quantities such as noise weights.
+
+        Returns
+        -------
+        multi_detector_states: lalpulsar.MultiDetectorStateSeries
+            Resulting multi-detector states produced by XLALGetMultiDetectorStatesFromMultiSFTs
+        multi_sfts: lalpulsar.MultiSFTVector
+            Only if `return_sfts` is True.
+            MultiSFTVector produced by XLALLoadMultiSFTs along the specified frequency band.
+        """
         # FIXME: Use MultiCatalogView once lalsuite implements the proper
         # SWIG wrapper around XLALLoadMultiSFTsFromView.
         sft_catalog = lalpulsar.SFTdataFind(sftfilepath, sft_constraint)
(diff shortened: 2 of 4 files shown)


Rodrigo-Tenorio:
@dbkeitel , this also depends on `attrs` --> Should be merged after #378 .

Rodrigo-Tenorio:
>  Compute SNR using weights.

Current LALSuite does not provide a SWIG interface to MultiNoiseWeights or similar (more specifically, there is no constructor to create MultiNoiseWeights objects at python level), meaning the only way weights can be used is by retrieving them from SFTs.

I'll probably open an issue @ lalsuite's repo and take care of this at some point, but for know I'd say it's not a high priority.

pyfstat/snr.py
```diff
@@ -0,0 +1,230 @@
...
+        frequency_wing_bins=1,
+        time_offset=0,
+        sft_constraint=None,
+        return_sfts=False,
+    ):
+        multi_sft_catalog = lalpulsar.GetMultiSFTCatalogView(
+            lalpulsar.SFTdataFind(sftfilepath, sft_constraint)
+        )
+        df = multi_sft_catalog.data[0].data[0].header.deltaF
+        multi_sfts = lalpulsar.LoadMultiSFTsFromView(
```
Rodrigo-Tenorio, line 166: SegFault is in this line

Rodrigo-Tenorio: I already encountered this somewhere: Catalog view is a thin wrapper around SFTCatalog, meaning it does *not* own any memory. SWIG, however, seems not to know that (yet), hence it tries to free the same memory twice.

Probably the following SWIG directive should be added to `LoedMultiSFTsFromView`:
```
#ifdef SWIG // SWIG interface directives
SWIGLAL( RETURN_OWNED_BY_1ST_ARG( <first argument>*, XLALLoadMultiSFTsFromView) );
#endif
```
Not sure rn since it's been a while, but I'll leave it here for the record.

@dbkeitel

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+        frequency_wing_bins: int
+            Frequency bins around the central frequency to retrieve from
+            SFT data. Bin size is determined using the SFT baseline time
+            as obtained from the catalog.
+            This option is only relevant if further information is to be
+            retrieved from the SFTs (i.e. `return_sfts=True`).
+        sft_constraint: lalpulsar.SFTConstraint
+            Optional argument to specify further constraints in XLALSFTdataFind.
+        return_sfts: bool
+            If True, also return the loaded SFTs. This is useful to compute futher
```
dbkeitel, line 331:
```suggestion
            If True, also return the loaded SFTs. This is useful to compute further
```

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+        time_stamps: array-like or dict
+            GPS timestamps at which detector states will be retrieved.
+            If array, use the same set of timestamps for all detectors,
+            which must be explicitly given by the user via`detectors`.
+            If dictionary, each key should correspond to a valid detector name
+            to be parsed by XLALParseMultiLALDetector and the associated value
+            should be an array-like set of GPS timestamps for each individual detector.
+
+        detectors: list[str] or comma-separated string
+            List of detectors to be parsed using XLALParseMultiLALDetector.
```
dbkeitel, line 282: specific that it's optional and exclusive with ts dict?

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+        Otherwise, formats `detectors` into a list and makes sure `timestamps`
+        is a 1D array containing numbers.
+        """
+
+        if isinstance(timestamps, dict):
+
+            if detectors is not None:
+                raise ValueError("`timestamps`' keys are redundant with `detectors`. ")
+
+            logging.debug("Retrieving detectors from tiemstamps dictionary.")
```
dbkeitel, line 376:
```suggestion
            logging.debug("Retrieving detectors from timestamps dictionary.")
```

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+        self.multi_timestamps = lalpulsar.CreateMultiLIGOTimeGPSVector(
+            self.multi_detector.length
+        )
+        for ind, ts in enumerate(timestamps):
+            self.multi_timestamps.data[ind] = self._numpy_array_to_LIGOTimeGPSVector(ts)
+
+    @staticmethod
+    def _numpy_array_to_LIGOTimeGPSVector(numpy_array):
+        """
+        Maps a numpy array of into a LIGOTimeGPS array using `np.floor`.
```
dbkeitel, line 403: of what? ;)

dbkeitel (changes requested):
Looks great overall! Lots of detail comments, many just documentation typos though.

- [ ] haven't looked at tests.py yet

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+    noise_weights: Union[lalpulsar.MultiNoiseWeights, None]
+        Optional, incompatible with `(assumeSqrtSX, Tsft)`.
+        Can be compute from SFTs using `SignalToNoiseRatio.from_sfts`.
+        Vector of noise weights to account for a varying noise floor or unequal noise
+        floors in different detectors.
+    assumeSqrtSX: float
+        Optional, incompatible with `noise_weights`.
+        *Single-sided* *amplitude* spectral density (ASD) of the detector noise.
+        This value is used for all detectors, meaning it's not possible to manually
+        specify different noise floors without creating SFT files.
```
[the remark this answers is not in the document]
Rodrigo-Tenorio: Deeper problems: I cannot access MultiNoiseWeights from Python as there's no constructor.

PFS can do it because for different sqrtSX it sets up the weights manually, but that cannot be done without proper access to MultiNoiseWeights. It's simple once updated in LALSuite though.

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+    Tsft: float
+        Optional, incompatible with `noise_weights`.
+        SFT baseline in seconds.
```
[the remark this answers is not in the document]
Rodrigo-Tenorio: If I had two overlapping SFTs, would `deltaT` be Tsft or Tsft/2?

pyfstat/snr.py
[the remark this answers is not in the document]
dbkeitel: Ah, that's where that comment was referring to. Ok to keep for future upgrade, then.

pyfstat/snr.py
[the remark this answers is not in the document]
dbkeitel: Good question, I've never touched overlapping SFTs.

The documentation says `(e.g. typically Tsft)`, which isn't exactly helpful. But [here](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_detector_states_8c_source.html#l0376) it's just set to the deltaT of the input timestamps object, so I guess it should always be =Tsft.

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+    Parameters
+    ----------
+    detector_states: lalpulsar.MultiDetectorStateSeries
+        MultiDetectorStateSeries as produced by DetectorStates.
+        Provides the required information to compute the antenna pattern contribution.
+    noise_weights: Union[lalpulsar.MultiNoiseWeights, None]
+        Optional, incompatible with `(assumeSqrtSX, Tsft)`.
+        Can be compute from SFTs using `SignalToNoiseRatio.from_sfts`.
+        Vector of noise weights to account for a varying noise floor or unequal noise
+        floors in different detectors.
```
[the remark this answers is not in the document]
Rodrigo-Tenorio:
```suggestion
        Noise weights to account for a varying noise floor or unequal noise
        floors in different detectors.
```
The input structure should already be a `lalpulsar.MultiNoiseWeights`, so I'll remove right away the "vector" mention.

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+            raise ValueError(
+                "Need either (`assumeSqrtSX`, `Tsft`) or `noise_weights` to account for background noise"
+            )
+
+    @classmethod
+    def from_sfts(
+        cls,
+        F0,
+        sftfilepath,
+        time_offset,
```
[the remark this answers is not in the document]
Rodrigo-Tenorio: I've seen it all over LALSuite, but I'm not sure why.

The reason I did not include it was not to mess up with hidden defaults.
More specifically, if Tsft is not given (which sometimes happens), `time_offset` will be set a few lines
below the code and then a warning should be raised or something; without a default, the user must make a choice.

pyfstat/snr.py
```diff
@@ -0,0 +1,420 @@
...
+        - The M matrix, which depends on the sky position and polarization angle
+        and encodes the effect of the detector's antenna pattern over the course
+        of the observing run.
+        - The JKS amplitude paramaters {A^0, A^1, A^2, A^3}, which are functions
+        of the CW's amplitude parameters (h0, cosi, psi) or, alternatively,
+        (aPlus, aCross, psi).
+
+    Parameters
+    ----------
+    detector_states: lalpulsar.MultiDetectorStateSeries
```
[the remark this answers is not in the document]
Rodrigo-Tenorio: As per a previous discussion, this can be extended with further constructors.

pyfstat/snr.py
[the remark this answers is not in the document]
Rodrigo-Tenorio: > But [here](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_detector_states_8c_source.html#l0376) it's just set to the deltaT of the input timestamps object, so I guess it should always be =Tsft.

I'll support this:

- [XLALGetMultiDetectorStatesFromMultiSFTs](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___detector_states__h.html#ga002ce3ea6588ceaaefeed51c2aba1d13) uses  [XLALExtractMultiTimestampsFromSFTs](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tutils__h.html#ga8b1acca29f4dc2f56aebe6a02c955a5b).
- In there, the actual function doing stuff is [XLALExtractTimestampsFromSFTs](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tutils__h.html#ga64f4017cb73595d2d5d627cc86b5b9a2)
- In there, there's this line:
ret->[deltaT](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/struct_l_i_g_o_time_g_p_s_vector.html#a29871cd2e0d0e49e396ae06c0a94eda1) = [TSFTfromDFreq](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_s_f_tutils_8c.html#a4568e794c914a827dbf61bc9a40183c9) ( sfts->[data](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/struct_c_o_m_p_l_e_x8_frequency_series_vector.html#a9b3d7acec959184ee1c7b7422ee1f9af)[0].[deltaF](https://lscsoft.docs.ligo.org/lalsuite/lal/struct_c_o_m_p_l_e_x8_frequency_series.html#a1632891eb1ff7300defe83c34de7f3b8) );

pyfstat/snr.py
```diff
@@ -0,0 +1,423 @@
...
+    value over noise realizations of twice the F-statistic of a perfectly
+    matched template to an existing signal in the data.
+
+    Computing SNR^2 requires two quantities:
+        - The antenna pattern matrix `M`, which depends on the sky position and polarization angle
+        and encodes the effect of the detector's antenna pattern over the course
+        of the observing run.
+        - The JKS amplitude parameters `{A^0, A^1, A^2, A^3}`, which are functions
+        of the CW's amplitude parameters `(h0, cosi, psi)` or, alternatively,
+        `(aPlus, aCross, psi)`.
```
dbkeitel, line 29: Looks like these two comments from initial review fell through the cracks:

- [ ]  Also phi0, though it doesn't really make a difference?

- [ ] Also add JKS ref?

pyfstat/snr.py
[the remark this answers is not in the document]
dbkeitel: I'd argue "if the user didn't have to think about it in lalapps, they shouldn't have to think about it now", and use that as a default. By allowing to change it, we're still offering more flexibility.

pyfstat/snr.py
```diff
@@ -0,0 +1,423 @@
...
+
+class DetectorStates:
+    """
+    Python interface to XLALGetMultiDetectorStates and XLALGetMultiDetectorStatesFromMultiSFTs.
+    """
+
+    def __init__(self):
+        self.ephems = lalpulsar.InitBarycenter(*get_ephemeris_files())
+
+    def multi_detector_states(self, timestamps, Tsft, detectors=None, time_offset=0):
```
dbkeitel, line 266: Didn't catch this on first round, but this should probably better be called `get_multi_detector_states`.

pyfstat/snr.py
```diff
@@ -0,0 +1,423 @@
...
+
+        self._parse_timestamps_and_detectors(timestamps, Tsft, detectors)
+        return lalpulsar.GetMultiDetectorStates(
+            self.multi_timestamps,
+            self.multi_detector,
+            self.ephems,
+            time_offset,
+        )
+
+    def multi_detector_states_from_sfts(
```
dbkeitel, line 301: Didn't catch this on first round, but this should probably better be called `get_multi_detector_states_from_sfts`.

pyfstat/snr.py
```diff
@@ -277,10 +274,12 @@ def multi_detector_states(self, timestamps, detectors=None, time_offset=0):
             If dictionary, each key should correspond to a valid detector name
             to be parsed by XLALParseMultiLALDetector and the associated value
             should be an array-like set of GPS timestamps for each individual detector.
-
+        Tsft: float
+            Timespan covered by each timestamp. I does not need to coincide with the
```
dbkeitel, line 278:
```suggestion
            Timespan covered by each timestamp. It does not need to coincide with the
```

dbkeitel (changes requested):

pyfstat/snr.py
```diff
@@ -0,0 +1,423 @@
...
+    such that `<2F> = 4 + SNR^2`, where `<2F>` represents the expected
+    value over noise realizations of twice the F-statistic of a perfectly
+    matched template to an existing signal in the data.
+
+    Computing SNR^2 requires two quantities:
+        - The antenna pattern matrix `M`, which depends on the sky position and polarization angle
+        and encodes the effect of the detector's antenna pattern over the course
+        of the observing run.
+        - The JKS amplitude parameters `{A^0, A^1, A^2, A^3}`, which are functions
+        of the CW's amplitude parameters `(h0, cosi, psi)` or, alternatively,
```
Rodrigo-Tenorio, line 28:
```suggestion
        of the CW's amplitude parameters `(h0, cosi, psi, phi0)` or, alternatively,
```

pyfstat/snr.py
Rodrigo-Tenorio:
```suggestion
        - The JKS amplitude parameters `{A^0, A^1, A^2, A^3}` 
        [Jaranowski, Krolak, Schutz Phys. Rev. D 58 063001,(1998)], which are functions
        of the CW's amplitude parameters `(h0, cosi, psi, phi0)` or, alternatively,
        `(aPlus, aCross, psi, phi0)`.
```

docs/source/pyfstat.rst
```diff
@@ -1,11 +1,6 @@
 pyfstat package
 ===============
 
-These pages document the full API and set of classes provided by PyFstat.
-
-See :doc:`here <readme>` for installation instructions and
-other general information.
-
```
dbkeitel, line 8: please reinstate

pyfstat/snr.py
```diff
@@ -0,0 +1,436 @@
...
+import logging
+import numpy as np
+from typing import Union
+
+from pyfstat.helper_functions import get_ephemeris_files
+
+
+@define(kw_only=True, slots=False)
+class SignalToNoiseRatio:
+    """Compute the optimal SNR of a CW signal on Gaussian noise or an arbitrary dataset of SFTs.
```
dbkeitel, line 13: Actually this is a bit misleading, as it by definition always assumes Gaussian noise, and there is no non-Gaussianity information extracted from the SFTs. (At most, non-stationarity.) So let's simplify that part away from the 1-line doc.

```suggestion
    """Compute the optimal SNR of a CW signal as expected in Gaussian noise.
```

dbkeitel (changes requested):
Some bits in the documentation that don't render properly (itemization not processed, and equations with spurious grave accents):

image

image

image

docs/source/pyfstat.rst
dbkeitel: ...but those sentences still show up at https://pyfstat--379.org.readthedocs.build/en/379/modules.html so I'm confused...

dbkeitel: ...ah, the output page where they've disappeared is https://pyfstat--379.org.readthedocs.build/en/379/pyfstat.html

pyfstat/snr.py
Rodrigo-Tenorio: `(non-)stationary Gaussian noise`?

Rodrigo-Tenorio:
Flake complains about "escaping characters" in equations.

pyfstat/snr.py
dbkeitel: Let's keep it simple, that addition is only likely to confuse people.

dbkeitel (approved):

PR #382, opened 2022-03-21
Support cwmfmd

Rodrigo-Tenorio:
Implements direct access to CWMakeFakeData .


dbkeitel:
Philosophical question: should we maybe go with `MakeData` instead of `MakeFakeData` since, like its lal counterparts, it can actually also reprocess real data? On the other hand, "name things the same as lal" makes for easier recognition.

Rodrigo-Tenorio:
>  since, like its lal counterparts, it can actually also reprocess real data?

I'd say "No", as it doesn't "reprocess" real data: real data is read and literally added at the end.

pyfstat/simulate.py
```diff
@@ -0,0 +1,101 @@
...
+            timestamps,
+            Tsft,
+            detectors=None,
+        )
+
+        # FIXME This doesn't fail here, why does it fail at test time????
+        logger.error(f"{multi_timestamps.data[0].data[0]}")
+        self.data_params.multiTimestamps = multi_timestamps
+        logger.error(f"{self.data_params.multiTimestamps.data[0].data[0]}")
+        self.data_params.multiIFO = multi_detector
```
Rodrigo-Tenorio, line 47: This works.

tests/test_simulate.py
```diff
@@ -0,0 +1,118 @@
...
+        assert getattr(mfd.data_params, key) == data_params[key]
+
+    detectors = list(data_params["timestamps"].keys())
+    num_detectors = len(detectors)
+
+    assert mfd.data_params.multiTimestamps.length == num_detectors
+    
+    for ifo_ind, ifo in enumerate(detectors):
+        # This fails for some reason...?
+        print(mfd.data_params.multiTimestamps.data[0].data[0])
```
Rodrigo-Tenorio, line 76: But this, as opposed to [the previous case](https://github.com/PyFstat/PyFstat/pull/382/files#r963821345), doesn't.

@dbkeitel

pyfstat/simulate.py
Rodrigo-Tenorio: Well, this works because `self.data_params.multiTimestamps` points to `multi_timestamps` by reference.
Once this scope is left, `multi_timestamps` doesn't exist anymore, some reference to deeper levels of the structure exists and seg fault raises.

Rodrigo-Tenorio: I don't see any 1D_vector swig declarations in [the source code](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_c_w_make_fake_data_8h_source.html#l00069), so that could probably be it...?

dbkeitel:
Not going to be ready for merging anytime soon, closing.

PR #386, opened 2022-03-24
Split test suite

dbkeitel:
- split up test suite by module
 - not properly converting to fixtures yet,
       shared stuff for now imported from
       commons_for_tests.py
- update MANIFEST.in with new tests dir


Rodrigo-Tenorio (approved):
I agree with this.

PR #387, opened 2022-03-25
Use isort for sorting import statement, more pre-commit hooks

dbkeitel:
This adds (and pre-runs) [isort](https://github.com/PyCQA/isort) to have a fixed ordering of package imports.

It comes _before_ `black` in the pre-commit hook order, so that if black were to disagree with it on any formatting choices, it would overrule it. However, there is then also the [flake8-isort](https://pypi.org/project/flake8-isort/) integration, which should (i) do nothing if black agreed with isort, (ii) throw an error if there is disagreement.

I also looked at the alternative [flake8-import-order](https://pypi.org/project/flake8-import-order/) but (at least with the configuration variants I tried) it does not seem to agree entirely with `isort`, and it doesn't come with its own active mode to actually sort things instead of just complaining, so the double `isort` seems the cleanest solution to me.

While at it, I also added a bunch more of trivial hooks.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,6 +16,7 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
+        args: ["--profile", "black", "--filter-files"]
         name: isort (python)
   - repo: https://github.com/ambv/black
     rev: 22.1.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,19 @@
 repos:
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        name: isort (python)
   - repo: https://github.com/ambv/black
     rev: 22.1.0
     hooks:
       - id: black
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.9.2
+    rev: 4.0.1
     hooks:
     - id: flake8
       additional_dependencies: [
           'flake8-docstrings==1.6.0',
           'flake8-executable==2.1.1',
+          'flake8-import-order==0.18.1',
       ]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,17 @@
 repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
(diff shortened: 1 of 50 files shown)


setup.cfg
```diff
@@ -7,6 +7,9 @@ max-line-length = 80
 select = C,E,F,W,B,B950
 ignore = E501,W503,E203
 
+[isort]
+profile = black
```
Rodrigo-Tenorio, line 11: [Documentation](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html) says `"black"`?

.pre-commit-config.yaml
```diff
@@ -1,13 +1,32 @@
...
+        - id: check-toml
+        - id: check-yaml
+        - id: detect-private-key
+        - id: end-of-file-fixer
+        - id: trailing-whitespace
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        name: isort (python)
```
Rodrigo-Tenorio, line 19: [Documentatiuon] says
```
    hooks:
      - id: isort
        args: ["--profile", "black", "--filter-files"]
```

Rodrigo-Tenorio (changes requested):
I checked the documentation and there are two points where things disagree, did you do it for any specific reason?

setup.cfg
dbkeitel: Your link is for `.isort.cfg`, while [here](https://pycqa.github.io/isort/docs/configuration/config_files.html) they don't use quotation marks for `setup.cfg`, so maybe those are slightly different dialects. Anyway, the current version seems to work...

.pre-commit-config.yaml
dbkeitel: I think I got it from [here](https://pycqa.github.io/isort/docs/configuration/pre-commit.html) where no extra options are set, but [here](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html) it has your syntax for black compatibility specifically, so indeed we should use that.

dbkeitel:
```suggestion
        args: ["--profile", "black", "--filter-files"]
        name: isort (python)
```

Rodrigo-Tenorio (approved):

dbkeitel:
Codecov may be timing out again, just merging anyway.

PR #389, opened 2022-03-28
Ignore setup.py dependencies to test conda environment 

Rodrigo-Tenorio:
@dbkeitel 

Not sure if it's going to be a long term solution , but will do the trick for now if it works.

Edit: Falling back to the "install using pip" route, including a specific flag to avoid installing anything with pip.

### DIFF
--- a/.github/workflows/conda-test.yml
+++ b/.github/workflows/conda-test.yml
@@ -28,6 +28,7 @@ jobs:
           conda config --show-sources
           conda config --show
           printenv | sort
+          pip install --no-dependencies .
       - name: test with pytest
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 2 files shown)


Rodrigo-Tenorio:
Alternative: Simply `pip install .`?

Rodrigo-Tenorio:
@dbkeitel ready if [this pipeline](https://github.com/PyFstat/PyFstat/runs/5734110910?check_suite_focus=true) passes

dbkeitel (approved):
Ok! We should also check off-line what to do when a developer works without installing the package, but for now that should indeed fix this test build.

Also also will be interesting to see what happens with the actual conda forge update, I should definitely do a pre-release this time.

PR #390, opened 2022-03-29
remove deprecated $LALPULSAR_DATADIR use for ephemerides

### LINKED ISSUE #356: simplify the ephemerides mess

dbkeitel:
- closes #356

@Rodrigo-Tenorio please check the Docker file especially.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,22 +2,6 @@ FROM python:3.9-buster
 
 WORKDIR /pyfstat-repo
 
-# Ephemerides arguments
-## Put ephemerides in this folder
-ARG EPHEMERIDES_FOLDER=/ephemerides/
-## URL to wget ephemerides from
-ARG EPHEMERIDES_URL=https://git.ligo.org/lscsoft/lalsuite/raw/master/lalpulsar/lib/
-## Earth epehemerides file name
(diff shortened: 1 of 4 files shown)


Rodrigo-Tenorio (approved):
Looks good

dbkeitel:
Updated README, merging.

PR #394, opened 2022-04-27
Fix up a syntax error

Rodrigo-Tenorio:
This should have been called out by a test?

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -369,7 +369,7 @@ class Writer(BaseSearchClass):
             # Each key should correspond to `detectors` if given;
             # otherwise, construct detectors from the given keys.
             if self.detectors is not None:
-                ifos_in_ts = list(self.timetamps.keys())
+                ifos_in_ts = list(self.timestamps.keys())
                 ifos_in_detectors = self.detectors.split(",")
                 if np.setdiff1d(ifos_in_ts, ifos_in_detectors).size:
                     raise ValueError(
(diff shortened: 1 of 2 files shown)


Rodrigo-Tenorio:
Ok: Tests never checks for timestamps + input detector, so it wasn't covered at all.

dbkeitel:
> Ok: Tests never checks for timestamps + input detector, so it wasn't covered at all.

Can you add such a case?

Rodrigo-Tenorio:
I think I just did. This is one of those cases in which fixtures would make our life easier, but it will work for now (again).

pyfstat/make_sfts.py
```diff
@@ -368,17 +368,17 @@ def _parse_timestamps(self):
...
+
             if self.detectors is not None:
-                ifos_in_ts = list(self.timetamps.keys())
                 ifos_in_detectors = self.detectors.split(",")
-                if np.setdiff1d(ifos_in_ts, ifos_in_detectors).size:
+                if np.setdiff1d(ifos, ifos_in_detectors).size:
                     raise ValueError(
-                        f"Detector names in timestamps dictionary ({ifos_in_ts}) "
+                        f"Detector names in timestamps dictionary ({ifos}) "
                         "are inconsistent with detector names given via keyword ({ifos_in_detectors})."
```
dbkeitel, line 379:
```suggestion
                        f"are inconsistent with detector names given via keyword ({ifos_in_detectors})."
```

dbkeitel (approved):

PR #399, opened 2022-05-31
README: clarify lalsuite-from-source

dbkeitel:


README.md
```diff
@@ -198,14 +198,29 @@ pip install pyfstat[chainconsumer,pycuda,style]
...
+Then after the `make && make install` step,
+you can point your environment to this installation by doing
+```
+source ${HOME}/lalsuite-install/etc/lalsuite-user-env.sh
 ```
 
+You have to redo this in every new session,
+unless you tweak your venv / conda env activation script
+to do it automatically.
+
```
Rodrigo-Tenorio, line 223: Alternatively, Python's builtin [site](https://docs.python.org/3.9/library/site.html#module-site) can be used to direct the Python interpreter to the right LALSuite installation using `site-packages`. To do so, add the following minimal script to one of the available `site-packages` paths (which can be retrieved using `python -c 'import site; print( site.getsitepackages())'`):


```
usercustomize.py
----------------
import sys

custom_lalsuite = "${HOME}/lalsuite-install/lib/pythonX.Y/site-packages"
sys.path.insert(0, custom_lalsuite)
```

where `pythonX.Y` is dependent upon the Python version with which LALSuite was installed.

Rodrigo-Tenorio, line 223:
```suggestion
Alternatively, Python's builtin [site](https://docs.python.org/3.9/library/site.html#module-site) can be used to direct the Python interpreter to the right LALSuite installation using `site-packages`. To do so, add the following minimal script to one of the available `site-packages` paths (which can be retrieved using `python -c 'import site; print( site.getsitepackages())'`):


```
usercustomize.py
----------------
import sys

custom_lalsuite = "${HOME}/lalsuite-install/lib/pythonX.Y/site-packages"
sys.path.insert(0, custom_lalsuite)
```

where `pythonX.Y` is dependent upon the Python version with which LALSuite was installed.
```

Rodrigo-Tenorio (changes requested):
Hi @dbkeitel 

Two minor changes:

1. Simplified `swig` options since not all of them are required.
2. Added "the Python way" of dealing with `activate + source` so that it works "properly" according to what Python expects... allegedly?

I'm not sure if I want PyFstat to recommend people about mangling their `site-packages`, as that's always a ticking bomb , but nothing should happen if they are using a conda environment ("Doesn't work? Wipe and start again"), so... I'm inclined to say "We are all adults around here" and leave it for the sake of completeness.

README.md
```diff
@@ -198,14 +198,29 @@ pip install pyfstat[chainconsumer,pycuda,style]
...
+If you prefer to make your own LALSuite installation
 [from source](https://git.ligo.org/lscsoft/lalsuite/),
 make sure it is **swig-enabled** and contains at least the `lalpulsar` and `lalapps` packages.
-A minimal configuration line to use would be e.g.:
+
+Following the instructions from their README,
+at the step where it tells you to run `./configure`,
+an example minimal line would be e.g.:
+```
+./configure --prefix=${HOME}/lalsuite-install --disable-all-lal --enable-lalpulsar --enable-lalapps --enable-swig --enable-swig-python --disable-swig-octave
```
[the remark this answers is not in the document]
dbkeitel: Ah, you're right. I added `--disable-swig-octave` explicitly because @itsluanayall had problems without it, but indeed just `--enable-swig-python` without `--enable-swig` should be equivalent.

README.md
Rodrigo-Tenorio: Sure, #401

Rodrigo-Tenorio (approved):

Issue #403, opened 2022-06-24
index error from ptemcee


dbkeitel:
Integration runs started failing yesterday. At first I thought it was a temporary github runner issue, since there's a random-looking pool error at the end of the log. But scrolling up, the real reason is
```
 >       acf = np.fft.ifft(f * np.conjugate(f), axis=axis)[m].real
E       IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/ptemcee/util.py:43: IndexError
```
from any and all MCMC tests. Seems like some of the persistent deprecation warnings from the mostly-unmaintained `ptemcee` have finally caught up with us! We might be able to pin some dependency (numpy?) for now, but will need a long-term solution via fork or replacement too.

dbkeitel:
Indeed there was a numpy 1.23.0 release on June 23: https://pypi.org/project/numpy/1.23.0/

PR #404, opened 2022-06-24
pin numpy<1.23.0 to avoid ptemcee index errors

dbkeitel:
- temporary workaround
- refs #403


dbkeitel:
That seems to have fixed it for now. Expedited merge.

PR #405, opened 2022-06-24
versioneer: set `tag_prefix = v`

### LINKED ISSUE #402: Declare 'v' as the tag prefix for versioneer
As we just learned over at https://github.com/Rodrigo-Tenorio/distromax/issues/14 that should be a more standard-compliant way. (Matching lalsuite, numpy, scipy amongst other examples.)

dbkeitel:
- closes #402

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,4 +15,4 @@ VCS = git
 style = pep440
 versionfile_source = pyfstat/_version.py
 versionfile_build = pyfstat/_version.py
-tag_prefix =
+tag_prefix = v


dbkeitel:
Not sure if this is actually doing what it should, as locally from this branch I still get
```
>>> import pyfstat
>>> pyfstat.__version__
'v1.14.1+13.5773ffea.dirty'
```

Rodrigo-Tenorio:
For reference, on  `distromax`....

- Installing as `pip install -e .`:
```
>>> import distromax
>>> distromax.__version__
'v1.0.4+21.5056896e.clean
```

```
$ pip list
...
distromax       1.0.4+21.5056896e.clean /home/rodrigo/gravity/git/distromax
...
```

- Installing from PyPI:
```
```
>>> import distromax
>>> distromax.__version__
'v1.0.4+21.5056896e.clean
```
```


```
$ pip list
...
distromax       1.0.4
...
```

dbkeitel:
1. It already works correctly (using the `tag_prefix` loaded from `setup.cfg` via the `parser` inside `versioneer.py`) when installing via a distribution made from `python setup.py sdist bdist_wheel` or simply by `pip install .` into a clean venv and then importing from _outside the source dir_
1. It does _not_ yet work when importing from within the source dir, because of `pyfstat/_version.py` _not_ parsing `setup.cfg` and instead just doing `cfg.tag_prefix = ""`.

 
The easiest fix to 2., short of putting in proper parsing, is to just hardcode the prefix there.

Rodrigo-Tenorio (approved):
Tested locally, it works :tada:

PR #413, opened 2022-06-30
enable context manager mode for ComputeFstat and TransientGridSearch

### LINKED ISSUE #411: pycuda context errors at garbage collection
To clean up GPU contexts after the pycuda transient functions (to avoid `The context stack was not empty upon module cleanup` errors), I originally implemented a custom `__del__()` on both `ComputeFstat` and the child transient class~~es~~ (note: only `TransientGridSearch`, never implemented for MCMC because it's not useful when sampling (t0,tau) points one-by-one). This seemed to work fine back in 2018 with whatever python and cuda versions I was using at the time.

But currently, on both CIT and Artemisa, this is no longer reliable: when using a `TransientGridSearch` directly from an executable script, it still works cleanly. But as soon as there's another function call around it, there's a `pycuda._driver.LogicError: cuCtxGetDevice failed: invalid device context` - essentially saying that python tried to garbage-collect the function twice and that isn't safe.

General python lesson learned: don't use `__del__`. If you think you want to use `__del__`, you don't, you actually want to use a custom context manager.

I think I have a working solution, just need to clean it up and test it properly.

dbkeitel:
- more robust for GPU usage in different contexts (pun intended)
   than `__del__()`
 - should not affect standard style
 - add test cases
 - include in example
 - fixes #411

Still in testing.

Edit: to be merged AFTER #414.

### DIFF
--- a/pyfstat/tcw_fstat_map_funcs.py
+++ b/pyfstat/tcw_fstat_map_funcs.py
@@ -132,31 +132,24 @@ class pyTransientFstatMap:
         B_tSG is marginalized over start-time and timescale of transient CW signal,
         using given type and parameters of transient window range.
 
-        NOTE: naive python port of the `lalpulsar.ComputeTransientBstat` implementation.
-        Should be optimised by vectorizing the for loops!
+        This is a python port of the `lalpulsar.ComputeTransientBstat` implementation,
+        replacing for loops by numpy operations.
         """
         # step through F_mn array subtract maxF and sum e^{F_mn - maxF}
         # The maximum-likelihood Fmax is globally subtracted from F_mn, and stored separatedly in the struct, because in most
         # expressions it is numerically more robust to compute e^(F_mn - Fmax), which at worst can underflow, while
         # e^F_mn can overflow (for F>~700). The constant offset e^Fmax is irrelevant for posteriors (normalization constant), or
         # can be handled separately, eg by computing log(B) = Fmax + log(sum e^(Fmn-Fmax)) for the Bayes-factor.
-        N_t0Range = self.F_mn.shape[0]
-        N_tauRange = self.F_mn.shape[1]
-        sum_eB = 0
-        for m in range(N_t0Range):
-            for n in range(N_tauRange):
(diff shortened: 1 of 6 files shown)


dbkeitel:
Seems to work well on Artemisa.

pyfstat/tcw_fstat_map_funcs.py
```diff
@@ -122,6 +124,33 @@ def get_maxF_idx(self):
...
+        replacing for loops by numpy operations.
+        """
+        # step through F_mn array subtract maxF and sum e^{F_mn - maxF}
+        # The maximum-likelihood Fmax is globally subtracted from F_mn, and stored separatedly in the struct, because in most
+        # expressions it is numerically more robust to compute e^(F_mn - Fmax), which at worst can underflow, while
+        # e^F_mn can overflow (for F>~700). The constant offset e^Fmax is irrelevant for posteriors (normalization constant), or
+        # can be handled separately, eg by computing log(B) = Fmax + log(sum e^(Fmn-Fmax)) for the Bayes-factor.
+        sum_eB = np.sum(np.exp(-self.maxF - self.F_mn))
+        # combine this to final log(Bstat) result with proper normalization (assuming rhohMax=1):
+        logBhat = self.maxF + np.log(sum_eB)  # unnormalized Bhat
```
Rodrigo-Tenorio, line 145:
```suggestion
        logBhat = self.maxF + logsumexp(-self.maxF - self.F_mn)
```
from `scipy.special.logsumexp`. Should be more stable.

Haven't timed it though.

dbkeitel: That could be because logsumexp has to find the maximum internally, while in your comparison call you didn't include it, and in the Bstat use case it is already precomputed.

I can still time the real thing to make sure.

But anyway this should be part of #414 which I wanted to get merged before this one; I had just rebased this one to be able to test more easily.

dbkeitel:
Closing in favour of something closer to #415 (which will need rebasing and cleanup too).

PR #414, opened 2022-07-01
implement transient BtSG statistic

dbkeitel:
- implemented in `ComputeFstat` and `TransientGridSearch` classes
- based on [XLALComputeTransientBstat](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___transient_c_w__utils__h.html#gaf66cf2ffa017a6b1dffd266139b52fad) and Prix, Giampanis & Messenger 2011
- introduces a `ComputeFstat.get_transient_detstats()` method and makes `get_transient_detstats.get_transient_maxTwoFstat()` a legacy wrapper around it
- refs #260

Currently with `tCWFstatMapVersion="lal"`, the original lalpulsar function is called for BtSG as well. With `tCWFstatMapVersion="pycuda"`, the python version of BtSG is called after the pycuda computation of the FstatMap. A naive python port of that function (with double for loop) was very slow (BtSG took up to factor 10 longer than FstatMap), but a simple change to `np.sum()` was enough to make the extra cost negligible, and so it seems porting BtSG to cuda is not necessary.

TODO:

- [x] add to `TransientMCMCSearch`
- [x] add meaningful quantitative tests
- [x] fix pipeline
- [x] squash

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -240,6 +240,12 @@ class ComputeFstat(BaseSearchClass):
     Normally this will read in existing data through the `sftfilepattern` argument,
     but if that option is `None` and the necessary alternative arguments are used,
     it can also generate simulated data (including noise and/or signals) on the fly.
+
+    NOTE that the detection statistics that can be computed from an instance of this class
+    depend on the `BSGL`, `BtSG` and `transientWindowType` arguments given at initialisation.
+    See `get_fullycoherent_detstat()` and `get_transient_detstats()` for details.
+    To change what you want to compute,
+    you may need to initialise a new instance with different options.
     """
 
     @helper_functions.initializer
@@ -253,6 +259,7 @@ class ComputeFstat(BaseSearchClass):
         binary=False,
         singleFstats=False,
         BSGL=False,
+        BtSG=False,
         transientWindowType=None,
         t0Band=None,
         tauBand=None,
@@ -309,6 +316,13 @@ class ComputeFstat(BaseSearchClass):
               for semicoherent searches.
 
             * Uniform per-detector prior line-vs-Gaussian odds.
+        BtSG: bool
+            If true and `transientWindowType` is not `None`,
+            compute the transient
+            :math:`\\ln\\mathcal{B}_{\\mathrm{tS}/\\mathrm{G}}`
+            statistic from Prix, Giampanis & Messenger (PRD 84, 023007, 2011)
+            (tCWFstatMap marginalised over uniform t0, tau priors).
+            rather than the maxTwoF value.
         transientWindowType: str
             If `rect` or `exp`,
             allow for the Fstat to be computed over a transient range.
@@ -540,6 +554,13 @@ class ComputeFstat(BaseSearchClass):
 
         self._get_SFTCatalog()
 
+        # some sanity checks on user options
+        if self.BSGL:  # pragma: no cover
+            if len(self.detector_names) < 2:
+                raise ValueError("Can't use BSGL with single detector data")
+            if getattr(self, "BtSG", False):
+                raise ValueError("Please choose only one of [BSGL,BtSG].")
+
         logging.info("Initialising ephems")
         ephems = lalpulsar.InitBarycenter(self.earth_ephem, self.sun_ephem)
 
@@ -661,17 +682,15 @@ class ComputeFstat(BaseSearchClass):
 
         # always initialise the twoFX array,
         # but only actually compute it if requested
+        self.twoF = 0
         self.twoFX = np.zeros(lalpulsar.PULSAR_MAX_DETECTORS)
         self.singleFstats = self.singleFstats or self.BSGL  # BSGL implies twoFX
         if self.singleFstats:
             self.whatToCompute += lalpulsar.FSTATQ_2F_PER_DET
 
         if self.BSGL:
-            if len(self.detector_names) < 2:
-                raise ValueError("Can't use BSGL with single detectors data")
-            else:
-                logging.info("Initialising BSGL")
-
+            logging.info("Initialising BSGL")
+            self.log10BSGL = np.nan
             # Tuning parameters - to be reviewed
             # We use a fixed Fstar0 for coherent searches,
             # and recompute it from a fixed p-value for the semicoherent case.
@@ -699,6 +718,9 @@ class ComputeFstat(BaseSearchClass):
 
         if self.transientWindowType:
             logging.info("Initialising transient parameters")
+            self.maxTwoF = 0
+            if getattr(self, "BtSG", False):
+                self.lnBtSG = np.nan
             self.windowRange = lalpulsar.transientWindowRange_t()
             transientWindowTypes = {
                 "none": lalpulsar.TRANSIENT_NONE,
@@ -993,15 +1015,39 @@ class ComputeFstat(BaseSearchClass):
         tstart=None,
         tend=None,
     ):
-        """Computes the detection statistic (twoF or log10BSGL) fully-coherently at a single point.
+        """Computes the detection statistic(s) fully-coherently at a single point.
 
-        These are also stored to `self.twoF` and `self.log10BSGL` respectively.
-        As the basic statistic of this class, `self.twoF` is always computed.
+        Currently supported statistics:
+
+        * twoF (CW)
+
+        * log10BSGL (CW or transient)
+
+        * maxTwoF (transient)
+
+        * lnBtSG (transient)
+
+        All computed statistics are stored as attributes,
+        but only one statistic is returned.
+
+        As the basic statistic of this class, `twoF` is always computed
+        and stored as `self.twoF` as well,
+        and it is the default return value.
+
         If `self.singleFstats`, additionally the single-detector
-        2F-stat values are saved in `self.twoFX`.
+        2F-stat values are stored in `self.twoFX`.
+
+        If `self.BSGL`, the `log10BSGL` statistic for CWs is additionally stored,
+        and it is returned instead of `twoF`.
 
         If transient parameters are enabled (`self.transientWindowType` is set),
-        the full transient-F-stat map will also be computed here,
+        `maxTwoF` will always be computed and stored,
+        and returned by default.
+        Depending on the `self.BSGL` and `self.BtSG` options,
+        either `log10BSGL` (a transient version of it, superseding the CW version)
+        or `lnBtSG` will also be computed, stored,
+        and returned instead of `maxTwoF`.
+        The full transient-F-stat map is also computed here,
         but stored in `self.FstatMap`, not returned.
 
         Parameters
@@ -1013,15 +1059,14 @@ class ComputeFstat(BaseSearchClass):
         tstart, tend: int or None
             GPS times to restrict the range of data used.
             If None: falls back to self.minStartTime and self.maxStartTime.
-            This is only passed on to `self.get_transient_detstat()`,
+            This is only passed on to `self.get_transient_detstats()`,
             i.e. only used if `self.transientWindowType` is set.
 
         Returns
         -------
         stat: float
-            A single value of the detection statistic (twoF or log10BSGL)
+            A single value of the main detection statistic
             at the input parameter values.
-            Also stored as `self.twoF` or `self.log10BSGL`.
         """
         self.get_fullycoherent_twoF(
             F0, F1, F2, Alpha, Delta, asini, period, ecc, tp, argp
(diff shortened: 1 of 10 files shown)


dbkeitel:
cc @itsluanayall @ReinhardPrix

dbkeitel:
Did some necessary cleanup and extended tests. As per the resolved in-line thread above, using `scipy.special.logsumexp` seems counterproductive, and moving the marginalisation into cuda does not seem to be required either. So I think this is ready for review - let me assign @Rodrigo-Tenorio but also add @itsluanayall because there are enough chances for me having forgotten a factor 2 or something somewhere that 3 pairs of eyes will be ideal.

pyfstat/tcw_fstat_map_funcs.py
```diff
@@ -122,6 +124,33 @@ def get_maxF_idx(self):
...
+        replacing for loops by numpy operations.
+        """
+        # step through F_mn array subtract maxF and sum e^{F_mn - maxF}
+        # The maximum-likelihood Fmax is globally subtracted from F_mn, and stored separatedly in the struct, because in most
+        # expressions it is numerically more robust to compute e^(F_mn - Fmax), which at worst can underflow, while
+        # e^F_mn can overflow (for F>~700). The constant offset e^Fmax is irrelevant for posteriors (normalization constant), or
+        # can be handled separately, eg by computing log(B) = Fmax + log(sum e^(Fmn-Fmax)) for the Bayes-factor.
+        sum_eB = np.sum(np.exp(-self.maxF - self.F_mn))
+        # combine this to final log(Bstat) result with proper normalization (assuming rhohMax=1):
+        logBhat = self.maxF + np.log(sum_eB)  # unnormalized Bhat
```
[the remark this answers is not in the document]
Rodrigo-Tenorio: So... No logsumexp on implementation but yes to test against?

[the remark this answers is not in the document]
dbkeitel: I've deliberately added a commit that introduces an extra function for it, and an immediate revert, so that we have this version in the history and can easily go back to testing against it if needed.

dbkeitel:
The codecov ` 85.48% of diff hit (target 70.78%)` doesn't look too great, but the missing lines are all exceptions, the `get_transient_maxTwoFstat` method I only kept around as a backwards compatibility helper, and the pycuda code that we currently can't test anyway.

I guess I could learn the syntax to tell codecov to ignore something, but for now it's above target already and hence happy.

pyfstat/core.py
```diff
@@ -661,16 +669,20 @@ def init_computefstatistic(self):
...
         if self.singleFstats:
             self.whatToCompute += lalpulsar.FSTATQ_2F_PER_DET
 
         if self.BSGL:
+            self.log10BSGL = np.nan
             if len(self.detector_names) < 2:
                 raise ValueError("Can't use BSGL with single detectors data")
             else:
                 logging.info("Initialising BSGL")
+            if getattr(self, "BtSG", False):
```
Rodrigo-Tenorio, line 684:
```suggestion
            if getattr(self, "BtSG", False): # pragma: no cover
```

pyfstat/core.py
```diff
@@ -1138,37 +1165,48 @@ def get_fullycoherent_log10BSGL(self):
...
         -------
-        maxTwoF: float
-            A single value of the detection statistic (twoF or log10BSGL)
+        detstat: float
+            A single value of the main chosen detection statistic
+            (twoF, log10BSGL or logBtSG)
             at the input parameter values.
-            Also stored as `self.maxTwoF`.
         """
+        if BtSG and BSGL:
```
Rodrigo-Tenorio, line 1207:
```suggestion
        if BtSG and BSGL: # pragma: no cover
```

pyfstat/grid_based_searches.py
```diff
@@ -943,8 +946,12 @@ def __init__(
         self.search_keys = ["F0", "F1", "F2", "Alpha", "Delta"]
         for k in self.search_keys:
             setattr(self, k, np.atleast_1d(getattr(self, k + "s")))
-        if self.BSGL:
+        if self.BSGL and self.BtSG:
```
Rodrigo-Tenorio, line 949:
```suggestion
        if self.BSGL and self.BtSG: # pragma: no cover
```

pyfstat/tcw_fstat_map_funcs.py
```diff
@@ -87,10 +97,22 @@ def __init__(self, N_t0Range=None, N_tauRange=None, transientFstatMap_t=None):
...
+            pre-allocated matrix from lalpulsar to initialise from.
+        from_file: str or None
+            Text file,
+            compatible with `lalpulsar.write_transientFstatMap_to_fp()` format,
+            to load and initialise from.
+        BtSG: boolean
+            If true, also compute the logBtSG transient Bayes factor statistic.
         """
-        if transientFstatMap_t:
+        if transientFstatMap_t and from_file:
```
Rodrigo-Tenorio, line 108:
```suggestion
        if transientFstatMap_t and from_file: # pragma: no cover
```

pyfstat/tcw_fstat_map_funcs.py
```diff
@@ -612,6 +716,11 @@ def pycuda_compute_transient_fstat_map(multiFstatAtoms, windowRange):
         )
     )
 
+    if BtSG:
```
Rodrigo-Tenorio, line 719:
```suggestion
    if BtSG: # pragma: no cover
```

Rodrigo-Tenorio (changes requested):
I'll let you decide whether you want to add these or not :wink:

Rodrigo-Tenorio (approved):
Examples and __init__ look fine to me.

pyfstat/core.py
Rodrigo-Tenorio, line 684: Can this be moved earlier during initialization?

pyfstat/core.py
```diff
@@ -993,10 +1008,22 @@ def get_fullycoherent_detstat(
...
+
+        Currently supported statistics:
+
+        * twoF (CW)
+
+        * log10BSGL (CW or transient)
+
+        * maxTwoF (transient)
+
+        * logBtSG (transient)
```
Rodrigo-Tenorio, line 1021:
```suggestion
        * log10BtSG (transient)
```
If not, why? Is `log10BSGL` historical and `logBtSG` ye new trend? In such case, `log` ---> `ln`.

pyfstat/core.py
```diff
@@ -1033,11 +1059,12 @@ def get_fullycoherent_detstat(
...
-        self.get_transient_maxTwoFstat(tstart, tend)
-        if self.BSGL is False:
-            return self.maxTwoF
-        else:
-            return self.get_transient_log10BSGL()
+        return self.get_transient_detstats(
+            tstart=tstart,
+            tend=tend,
+            BSGL=getattr(self, "BSGL", False),
+            BtSG=getattr(self, "BtSG", False),
```
Rodrigo-Tenorio, line 1066: Can these be `self.BSGL=self.BtSG=False` by default at initialization time?

Rodrigo-Tenorio: Actually, why give it as inputs if already present as attributes?

pyfstat/grid_based_searches.py
```diff
@@ -50,7 +50,9 @@ class GridSearch(BaseSearchClass):
         "Alpha": r"$\alpha$",
         "Delta": r"$\delta$",
         "twoF": r"$\widetilde{2\mathcal{F}}$",
-        "log10BSGL": r"$\log_{10}\mathcal{B}_{\mathrm{SGL}}$",
+        "maxTwoF": r"$\max\widetilde{2\mathcal{F}}$",
+        "log10BSGL": r"$\log_{10}\mathcal{B}_{\mathrm{S/GL}}$",
+        "logBtSG": r"$\log\mathcal{B}_{\mathrm{tS/G}}$",
```
Rodrigo-Tenorio, line 55: Same question re `log`.

pyfstat/tcw_fstat_map_funcs.py
```diff
@@ -122,11 +184,36 @@ def get_maxF_idx(self):
         """
         return np.unravel_index(np.argmax(self.F_mn), self.F_mn.shape)
 
+    def get_logBtSG(self):
+        """Compute (natural log of the) transient-CW Bayes-factor B_tSG = P(x|HyptS)/P(x|HypG).
```
Rodrigo-Tenorio, line 188: Ahà! So its natural log @dbkeitel : Why not `ln`?

Rodrigo-Tenorio: Also... `HyptS`-->`Hyp[tS]`? `H_{tS}`?

pyfstat/tcw_fstat_map_funcs.py
```diff
@@ -122,11 +184,36 @@ def get_maxF_idx(self):
...
+
+        Here HypG = Gaussian noise hypothesis, HyptS = transient-CW signal hypothesis.
+
+        B_tSG is marginalized over start-time and timescale of transient CW signal,
+        using given type and parameters of transient window range.
+
+        This is a python port of the `lalpulsar.ComputeTransientBstat` implementation,
+        replacing for loops by numpy operations.
```
Rodrigo-Tenorio, line 196:
```suggestion
        replacing `for loops by numpy operations.
```

pyfstat/tcw_fstat_map_funcs.py
dbkeitel: Rejecting this one, this whole function _should_ be covered, including this case.

pyfstat/core.py
dbkeitel: probably, will include in a cleanup commit

pyfstat/core.py
dbkeitel: Historical divergence in lalsuite conventions for the two statistics. I thought about normalising to log10, but then one always has to convert when wanting to compare to lalsuite outputs. Using the less ambiguous `ln` is a very good suggestion, though.

pyfstat/core.py
dbkeitel: We're in a bit of a messy situation of sometimes allowing such methods to be called independently of what the setup was (see e.g. the test for this one), sometimes not. So giving the lower-level function these input options, despite its caller in the standard graph using the values from `self`, allows for such more flexible out-of-graph calls. It's probably not great practice, but kinda convenient...

pyfstat/tcw_fstat_map_funcs.py
dbkeitel: Copied like this from lalpulsar comments. Version with underscore (and mathrm) would be the proper texy one.

pyfstat/core.py
dbkeitel: Got rid of these as discussed.

dbkeitel:
Cleaned up and simplified the API after discussions with @Rodrigo-Tenorio . Now if people want to switch between computing lnBtSG and log10BSGL, they need to initialise a new ComputeFstat class. See `test_transient_detstats()` for a worked example.

pyfstat/core.py
```diff
@@ -699,6 +718,9 @@ def init_computefstatistic(self):
 
         if self.transientWindowType:
             logging.info("Initialising transient parameters")
+            self.maxTwoF = 0
+            if getattr(self, "BtSG", False):
```
Rodrigo-Tenorio, line 722:
```suggestion
            if self.BtSG:
```
I don't recall if there was a good reason for this specific one or we simply forgot to change it.
Attribute `BtSG` is always initialized, so no need for getattr.

pyfstat/core.py
```diff
@@ -1180,23 +1222,63 @@ def get_transient_maxTwoFstat(
...
         )
 
-        # Now instead of the overall twoF,
-        # we get the maximum twoF over the transient window range.
+        # get the maximum twoF over the transient window range
         self.maxTwoF = 2 * self.FstatMap.maxF
         if np.isnan(self.maxTwoF):
             self.maxTwoF = 0
+
+        if getattr(self, "BtSG", False):
```
Rodrigo-Tenorio, line 1237:
```suggestion
        if self.BtSG:
```

Rodrigo-Tenorio (changes requested):
Overall consistent with what we discussed

pyfstat/core.py
dbkeitel: Put it back in specifically because when called from semicoherent classes, it was not initialised.

pyfstat/core.py
dbkeitel: See above, needed to make semicoherent classes to run without explicitly setting it to False in those too.

Issue #417, opened 2022-07-11
migration from lalapps to lalpulsar


dbkeitel:
We still rely on a bunch of lalapps executables. There is now finally a move to break up that package, with some tools (relevant for us: `lalapps_tconvert`) moved to `lal` in https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1901 and everything else of relevance to us moving to `lalpulsar` in https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1904 .

Of course, ideally we'd replace most calls of such executables with pure python/SWIG solutions, but I don't think we have the time for that currently.

- [x] Meanwhile, the `lalapps` executable names will continue to function as wrappers, but throw warnings, so once the new release with the completed moves is out, we should actually rename them. -> more urgent due to #425 -> fix and first part of migration in #426

- [ ] We can then also drop the `lalapps` dependency in the [conda recipe](https://github.com/conda-forge/pyfstat-feedstock/blob/main/recipe/meta.yaml) but probably have to add new binary ~~`lal` and~~ `lalpulsar` dependencies (in addition to the `python-*` we already have). -> pending on next lalsuite release

- [x] For now, the `lalapps_tconvert` wrapper is actually broken in the nightly wheels, see [this test failure](https://github.com/PyFstat/PyFstat/runs/7279899502?check_suite_focus=true#step:7:6126) -> reported upstream as https://git.ligo.org/lscsoft/lalsuite/-/issues/555 -> actually it was easy to just get rid of this one, see #418

dbkeitel:
Nightly issue also fixed from lalsuite side, see successful run https://github.com/PyFstat/PyFstat/actions/runs/2663238609

dbkeitel:
Warning: Current conda-forge releases are messed up with lalapps 9.0.0 being wrappers-only already but lalpulsar 4.0.0 not having the executables yet, and lalapps not enforcing a dependency. -> https://git.ligo.org/lscsoft/lalsuite/-/issues/588

But they're working on new releases already, so I'll wait for that to update our recipes.

dbkeitel:
After #438 we jut need a release and to update the feedstock.

dbkeitel:
https://anaconda.org/conda-forge/pyfstat  1.17.0 should have this done

PR #426, opened 2022-08-01
lalapps->lalpulsar migration with compatibility helper function

### LINKED ISSUE #425: lalapps->lalpulsar change breaks Writer.check_cached_data_okay_to_use()
lalsuite-dev tests started failing at `test_makefakedata_usecached": https://github.com/PyFstat/PyFstat/runs/7617202347?check_suite_focus=true

The problem is that after https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/1904/ we still call `lalapps_Makefakedata_v5` but that is now just a wrapper to `lalpulsar_Makefakedata_v5`, which is then the executable name stored in the SFT header and read for comparison, triggering an unneeded rerun.

dbkeitel:
fixes #425 and makes progress on #417 by adding a transitional helper function `get_lal_exec()` for dynamically selecting `lalpulsar_` if available and falling back to `lalapps_` else.

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -666,7 +666,7 @@ def get_doppler_params_output_format(keys):
 
     This uses the same format (`%.16g`) as
     the `write_FstatCandidate_to_fp()` function of
-    `lalapps_ComputeFstatistic_v2`.
+    the `ComputeFstatistic_v2` executable.
 
     This assigns that format to each parameter name in `keys`
     which matches a hardcoded list of known standard 'Doppler' parameters,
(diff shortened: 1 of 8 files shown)


dbkeitel:
dev test with pre-release LALSuite passed: https://github.com/PyFstat/PyFstat/actions/runs/2777648827

pyfstat/helper_functions.py
```diff
@@ -1277,3 +1279,28 @@ def gps_to_datestr_utc(gps):
...
+    -------
+    cmd: str
+        Base executable name without lalapps/lalpulsar prefix.
+
+    Returns
+    -------
+    full_cmd: str
+        Full executable name with the right prefix.
+    """
+    full_cmd = shutil.which("lalpulsar_" + cmd) or shutil.which("lalapps_" + cmd)
```
Rodrigo-Tenorio, line 1301: :+1:

Rodrigo-Tenorio (approved):

PR #428, opened 2022-08-22
Hotfix timestamps snr

### LINKED ISSUE #427: Bug in `_parse_timestamps_and_detectors` prevents use of dict timestamps
When trying to create detector states from a dictionary of timestamps the following happens:
```
  File "pyfstat/snr.py", line 436, in _parse_timestamps_and_detectors
    detectors = list(timestamps.key())
AttributeError: 'dict' object has no attribute 'key'
```

- [x] Check why that wasn't covered by the test suite.
- [x] Change `key` to `keys`.

Rodrigo-Tenorio:
Closes #427 

Add minimal tests for `DetectorStates` to make sure that
1. Different inputs actually work.
2. Different inputs are consistent amongst each other.

### DIFF
--- a/pyfstat/snr.py
+++ b/pyfstat/snr.py
@@ -56,20 +56,19 @@ class SignalToNoiseRatio:
 
     def __attrs_post_init__(self):
 
-        self.Tsft = self.detector_states.data[0].deltaT
+        have_noise_weights = self.noise_weights is not None
+        have_sqrtSX = self.assumeSqrtSX is not None
 
-        # FIXME: There has to be a simpler logic for this
-        if self.noise_weights is None:
-            if self.assumeSqrtSX is not None:
-                self.Sinv_Tsft = self.Tsft / self.assumeSqrtSX**2
-            else:
-                raise ValueError(
-                    "Need either `assumeSqrtSX` or `noise_weights` to account for background noise"
-                )
-        elif self.assumeSqrtSX is not None:
+        if have_noise_weights == have_sqrtSX:
             raise ValueError(
-                "Need either `assumeSqrtSX` or `noise_weights` to account for background noise"
+                "Need either `assumeSqrtSX` or `noise_weights` to account for background noise."
             )
+
+        self.Tsft = self.detector_states.data[0].deltaT
+        if not have_noise_weights:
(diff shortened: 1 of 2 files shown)


pyfstat/snr.py
```diff
@@ -56,21 +56,20 @@ class SignalToNoiseRatio:
...
-                )
-        elif self.assumeSqrtSX is not None:
+        if have_noise_weights == have_sqrtSX:
             raise ValueError(
-                "Need either `assumeSqrtSX` or `noise_weights` to account for background noise"
+                "Need either `assumeSqrtSX` or `noise_weights` to account for background noise."
             )
 
+        self.Tsft = self.detector_states.data[0].deltaT
+        if not have_noise_weights:
```
dbkeitel, line 68:
```suggestion
        if have_sqrtSX:
```

Just to be in principle ready for a non-binary future.

pyfstat/snr.py
```diff
@@ -430,10 +429,19 @@ def _parse_timestamps_and_detectors(self, timestamps, Tsft, detectors):
...
 
             if detectors is not None:
-                raise ValueError("`timestamps`' keys are redundant with `detectors`. ")
+                raise ValueError("`timestamps`' keys are redundant with `detectors`.")
+            for ifo in timestamps:
+                try:
+                    lalpulsar.FindCWDetector(name=ifo, exactMatch=True)
+                except Exception:
+                    raise ValueError(
+                        "Invalid detector name in timestamps. "
```
dbkeitel, line 438:
```suggestion
                        f"Invalid detector name {ifo} in timestamps. "
```

dbkeitel (changes requested):
Two small amendments. Otherwise looks good.

Also for the future I'd suggest to first commit the new test so that one can see the breakage in the logs, then the fix after. (On lalsuite I was told the opposite is preferred; but as we're a smaller project, the break-first approach seems more informative than a hassle.)

pyfstat/snr.py
Rodrigo-Tenorio: I agree. There is a XOR right before this, so `not have_noise_weights` <-> `have_sqrtSX`.

dbkeitel (approved):

PR #429, opened 2022-08-23
Miscelaneous overhaul: LaTeX, notebooks, bashplotlib, tqdm, argparse

### LINKED ISSUE #78: make LaTeX use optional in plotting
As suggested by @GregoryAshton in #26 it would be nice to make the latex use in any plotting optional, to make it more robust on some clusters and possibly reduce the risk of #26. I think this only applies to MCMC classes at the moment.

### LINKED ISSUE #421: black vs tutorial notebooks
@Rodrigo-Tenorio just noticed that a top-level `black .` will modify the notebooks in `examples/tutorials`. Questions:

- [x] can you confirm whether they'd be screwed up by that or if it's safe?
- [x] do you remember where they are explicitly, or implicitly, excluded from the standard black hooks?

Rodrigo-Tenorio:
- [x] LaTeX 
It's disable by default in matplotlib and gives several problems on systems without LaTeX, so we'll just let the user activate it if they really want it. Closes #78 .

- [x] Add `%matplotlib inline` to tutorials. 
Otherwise they may not correctly display in some systems.

- [x] Add proper linting to tutorials. Closes #421.

- [x] Bashplotlib
No apparent use, let's remove dependencies.

- [x] tqdm is already required
Remove optional tqdm functions.

- [ ] Remove argparse.
Needs @dbkeitel 's approval.

- [ ] Deprecate follow-up module. 
@dbkeitel 's approval.


dbkeitel:
Let's keep the last two items for a different PR/discussion, I'll look at what you've implemented so far and we can merge that on its own.

Bashplotlib is/was used for those "SFT timesamps" graphics I think, but honestly I never found them very informative.

Issue #333, opened 2021-10-04
Integrate multi-stage follow-up Bayes factor functionalities into PyFstat


Rodrigo-Tenorio:
[Tenorio, Keitel, Sintes (2021)](https://arxiv.org/abs/2105.13860) introduces a new Bayes factor to evaluate the multi-stage follow-up of a CW candidate. The following items should be addressed in order to fully implement this new functionality into PyFstat:

- [ ] Overhaul `MCMCFollowUpSearch` and create a new class `BstarSN`:
    - [ ] Re-purpose not do duplicate Coherent and `Semicoherent MCMC` searches.
    - [ ] Uncouple and clean up coherence-time ladder computation.
    - [ ] Add prior class (or similar) to parse posterior samples from previous stages to setup priors. 
    
-  [ ] Implement noise-hypothesis functions into `BstarSN`:
      - [ ] Implement off-sourcing (maybe it's simpler to use CFSv2's input grid functionality ? )
      - [ ] Implement Gumbel fit.

- [ ] Implement signal-hypothesis functions into `BstarSN`:
    - [ ] Re-compute 2F at different stages using the loudest candidate from the last one. 
    - [ ] Implement 2F -> rho -> 2F integral with numerical checks.

- [ ] Add reference to README #332

PR #432, opened 2022-08-23
Deprecate old MCMCFollowUpSearch and optimal setup functions

### LINKED ISSUE #431: Deprecate follow-up
As a first step towards #333 , current follow-up should probably be preemptively deprecated just to avoid any confusion.

Rodrigo-Tenorio:
First step towards #333 : Make sure there's no confusion and things are not wrongly used by mistake.

- closes #431

### DIFF
--- a/pyfstat/mcmc_based_searches.py
+++ b/pyfstat/mcmc_based_searches.py
@@ -2938,7 +2938,7 @@ class MCMCSemiCoherentSearch(MCMCSearch):
         return np.sum(H)
 
 
-class MCMCFollowUpSearch(MCMCSemiCoherentSearch):
+class MCMCFollowUpSearch(MCMCSemiCoherentSearch, core.DeprecatedClass):
     """Hierarchical follow-up procedure
 
     Executes MCMC runs with increasing coherence times in order to follow up a parameter space


dbkeitel (commented):
Not so sure about deprecating the optimal module. I guess those aren't coded up anywhere else, so there's some value in keeping them around, and users could then still emulate the behaviour of  `MCMCFollowUpSearch` by calling those functions and the standard class in a loop. Thoughts?

Rodrigo-Tenorio:
@dbkeitel ideally the useful methods from that part will be re-implemented in a similar way to the `snr` module so they are easily access in general.

dbkeitel:
> @dbkeitel ideally the useful methods from that part will be re-implemented in a similar way to the `snr` module so they are easily access in general.

Sounds good, but then I'd leave the deprecation until those are available. The class itself can be cleanly deprecated already in the sense that it can be reproduced by a loop over the standard class, but there's no replacement for these functions yet so seems premature to deprecate them.

Rodrigo-Tenorio:
Sure, makes sense  :+1:

dbkeitel (approved):

PR #434, opened 2022-08-23
test examples with --clean

dbkeitel:
While we still have the global `--clean` option, let me use the opportunity to run the all-examples action and see where it fails: https://github.com/PyFstat/PyFstat/actions/runs/2912512784

(I already know it's broken for at least some grid classes)


dbkeitel:
To be followed up as #437.

PR #433, opened 2022-08-23
Refurbish argparse options

Rodrigo-Tenorio:
Remove unused, add as method/class attributes the ones on use.

PR #435, opened 2022-08-23
Refurbish logging

### LINKED ISSUE #163: Fix logging behaviour

Rodrigo-Tenorio:
Make logging work once and for all.

Must be merged after #433 .

Closes #163 and #435

For the record: Logging module was initially inspired by that of [nessai](https://github.com/mj-will/nessai),
but it looks like we are starting to diverge from it into another direction.


Rodrigo-Tenorio:
@dbkeitel , I finally understood *the* `logging` shenanigan.

Loggers have a name and allow for "subloggers" to be created using the standard dot notation (i.e. `A.B` is a child logger of `A` and reports with the same format and so on unless overwritten).

The trick here is that `set_up_logger` creates a logger called `pyfstat` at "aplication level" (i.e. everytime you want to run something), and every singe module of PyFstat creates a new logger via
```
logger = logging.getLogger(__name__)
```
Since `__name__` is essentially `pyfstat.<module_name>`, this implies each module creates a *sublogger* from *pyfstat* reporting to the same outputs, meaning most of Greg's old code on logging is actually unnecessary nowadays if I'm not mistaken.

Rodrigo-Tenorio:
For the record: Ran all examples [here](https://github.com/PyFstat/PyFstat/actions/runs/2913556483)

dbkeitel:
Thanks @Rodrigo-Tenorio that looks promising. However I don't like requiring the user to do an extra setup call every time. Can't that be avoided somehow? I see your "based on nessai" is in turn "based on bilby", and I don't remember bilby requiring such a line in every caller script. Can't the default setup still happen at import time somehow?

This is also still showing changes from #433 in the diff, so seems to need a manual rebase.

Rodrigo-Tenorio:
All I find so far is that logging shouldn't be configured by default, because is up to the user to configure it, if I'm not mistaken (e.g. https://stackoverflow.com/questions/65303202/setup-logging-before-importing-other-modules).

I *could* add a `set_up_logger` call under `pyfstat/__init__.py` such that upon importing PyFstat a logger logging to stdout is configured by default @dbkeitel ?

dbkeitel:
Tested the branch locally, a few extra observations:
- I like that it now identifies (via the sublogger name) where a message comes from, e.g. `pyfstat.make_sfts`
-[ ] would it be easy enough to amend the datetime format `08-26 15:57` to match lal's `2022-08-26 15:57:22.2583`?
- An actual nice side effect of having to set up the logger at caller script level is that we could then also replace all `print()` in examples by proper logging. But still not sure if that's worth making it required. I'll also check how bilby and pesummary handle this.

Rodrigo-Tenorio:
- [ ] Ammend datetime format to match LAL's (as simple as writing it down [here] :+1: )
- [ ] Make all prints go through log (that includes `print` and `tqdm`).

dbkeitel:
After playing around a bit with the current version:
1. if caller script does not even import `logging`, or imports it and does nothing else, then no output from PyFstat is shown on terminal at all (unfortunate for new users)
2. if caller script does manual setup like e.g. a minimal `logging.basicConfig(level=logging.INFO)` then one gets full output, with format as of that setup (as it should be)
3. if caller script does `logger = pyfstat.set_up_logger(label=label, outdir=outdir)`, as currently implemented, everything is formatted as we want it

So I'd say 2+3 are ok as is, but we could fix 1 without violating `is up to the user to configure it` if we only do it _as a fallback_. If somehow within the package, we can check that no logger configuration has been done yet. Then we could activate ours and even add a friendly "you didn't set logging, I'll do it for you, but feel free to override me".

pyfstat/helper_functions.py
```diff
@@ -41,48 +42,80 @@ def set_up_matplotlib_defaults():
...
-    stream_handler.setFormatter(
-        logging.Formatter("%(asctime)s %(levelname)-8s: %(message)s", datefmt="%H:%M")
-    )
+    Based on the implementation in Nessai:
+    https://github.com/mj-will/nessai/blob/main/nessai/utils/logging.py
+
+    Parameters
+    ----------
+    outdir : str, optional
+        Path of to outdir directory.
```
dbkeitel, line 54: Is this actually doing something currently? We don't seem to actively support stream redirection to files through a separate parameter here. But then I see below there's some logic using this param if something is found in `logging.FileHandler`. So is the idea that if, and only if, the user has already preconfigured some file output before calling this, we will intercept it and redirect to our outdir...?

Rodrigo-Tenorio: Yes. We actually were using that functionality before in `BaseSearchClass`.

`set_up_logger` takes a logger (the thing that logs) and attaches two handlers (things that "record" the logs): A `StreamHandler`, which I just fixed to log into STDOUT, and a `FileHandler`(as before), which goes into a file.

Loggers set up by `set_up_logger` *always* print logs to STDOUT *and* a specified file.

dbkeitel: > Loggers set up by set_up_logger always print logs to STDOUT and a specified file.

Ah, I didn't know that!

But do I understand right that when a file handler was preconfigured, we'll "steal" it silently?

Rodrigo-Tenorio: As in we will be writing to the same file?
I... think so?

Handlers are additive and independent, so I'd assume that we'd just be wrottong to the same file with different loggin label, but I haven't tried that situation.

Rodrigo-Tenorio: Yes: Some transient examples set the filename manually.

examples/transient_examples/PyFstat_example_long_transient_MCMC_search.py
```diff
@@ -25,6 +25,8 @@
         "Please first run PyFstat_example_make_data_for_long_transient_search.py !"
     )
 
+logger = pyfstat.set_up_logger(label="long_transient_mcmc_search", outdir=data.outdir)
```
Rodrigo-Tenorio, line 28: @dbkeitel

pyfstat/helper_functions.py
dbkeitel: Ok figured out an easy way to test: adding `logging.basicConfig(level=logging.INFO, filename="log.log")` in any example before a `pyfstat.set_up_logger` call -> all of the following get the info:
*    log.log
*    stdout
*    outdir/label.log

I think that's the expected behaviour. Thread resolved.

examples/transient_examples/PyFstat_example_long_transient_MCMC_search.py
dbkeitel: see https://github.com/PyFstat/PyFstat/pull/435#discussion_r956237275

pyfstat/helper_functions.py
```diff
@@ -41,48 +43,80 @@ def set_up_matplotlib_defaults():
...
-    )
+    Based on the implementation in Nessai:
+    https://github.com/mj-will/nessai/blob/main/nessai/utils/logging.py
+
+    Parameters
+    ----------
+    outdir : str, optional
+        Path to outdir directory.
+    label : str, optional
+        Label for this instance of the logger.
```
dbkeitel, line 57: I don't think that's true. It seems to only get used for the file name. While the logger instance is always called "pyfstat". And actually I'd say that's fine, as the logger instance name gets written on every log line, which would get spammy with run labels.

pyfstat/loggers.py
```diff
@@ -0,0 +1,89 @@
...
+
+    """
+    level = _parse_log_level(log_level)
+
+    logger = logging.getLogger("pyfstat")
+    logger.setLevel(level)
+
+    common_formatter = logging.Formatter(
+        "%(asctime)s.%(msecs)03d %(name)s %(levelname)-8s: %(message)s",
+        datefmt="%y-%m-%d %H:%M:%S",
```
dbkeitel, line 63:
```suggestion
        datefmt="%y-%m-%d %H:%M:%S",  # intended to match LALSuite's format
```

pyfstat/loggers.py
```diff
@@ -0,0 +1,89 @@
...
+"""
+
+import logging
+import os
+import sys
+
+from pyfstat.helper_functions import get_version_string
+
+
+def _parse_log_level(log_level):
```
dbkeitel, line 18: Why do you need this one? Isn't it best to always set levels as `logging.INFO` and similar? And if you need to do so from a string, I guess you could just do `getattr(logging,string)`?

pyfstat/loggers.py
```diff
@@ -0,0 +1,126 @@
...
+
+The default behaviour is to attach a `logging.StreamHandler` to
+the 'pyfstat' printing out to `sys.stdout` upon importing the package.
+If, for any reason, `logging` cannot access `sys.stdout` at import time,
+the exception is reported via `print` and no handlers are attached
+(i.e. the logger won't print to `sys.stoud`).
+
+The user can modify the logger's behaviour at run-time using `set_up_logger`.
+This function attaches extra `logging.StreamHandler` and `logging.FileHandler`
+handlers to the logger, allowing to redirect loggin messages to either a different
```
dbkeitel, line 25:
```suggestion
handlers to the logger, allowing to redirect logging messages to a different
```

pyfstat/loggers.py
```diff
@@ -0,0 +1,126 @@
...
+stream or a specific output file specified using the `outdir, label` variables
+(with the same format as in the rest of the package).
+
+Finally, logging can be disable at run-time by manually configuring the 'pyfstat'
```
dbkeitel, line 29:
```suggestion
Finally, logging can be disabled, or the level changed, at run-time by manually configuring the 'pyfstat'
```

pyfstat/loggers.py
```diff
@@ -0,0 +1,126 @@
...
+    ----------
+    outdir: str, optional
+        Path to outdir directory. If `None`, no file handler will be added.
+    label: str, optional
+        Label for this instance of the logger.
+        This is consistent with the rest of the package: ``label'' referes
+        to the string prepended at every file produced by a script.
+        Required, in conjunction with `outdir`, to add a file handler.
+    log_level: {'ERROR', 'WARNING', 'INFO', 'DEBUG'}, optional
+        Level of logging. This level is imposed the logger itself and
```
dbkeitel, line 62:
```suggestion
        Level of logging. This level is imposed on the logger itself and
```

pyfstat/loggers.py
```diff
@@ -0,0 +1,126 @@
...
+        *every single handler* attached to it.
+    streams: io.TextIOWrapper, optional
+        Stream to which logging messages will be passed using a
+        StreamHandler object. If `None`, a handler to `sys.stdout`
+        will be attached unless it already exists.
+        Other common streams include e.g. `sys.stderr`.
+    append: bool, optional
+        If True, removes all handlers from the `pyfstat` logger.
```
dbkeitel, line 70: That's the opposite of appending...?

pyfstat/loggers.py
```diff
@@ -0,0 +1,126 @@
...
+    obj:`logging.Logger`
+        Instance of the Logger class.
+
+    """
+    logger = logging.getLogger("pyfstat")
+    logger.setLevel(log_level)
+
+    if not append:
+        while logger.hasHandlers():
+            logger.removeHandler(logger.hadlers[0])
```
dbkeitel, line 83:
```suggestion
            logger.removeHandler(logger.handlers[0])
```

pyfstat/loggers.py
```diff
@@ -0,0 +1,126 @@
...
+        stream_handler = logging.StreamHandler(stream)
+        stream_handler.setFormatter(common_formatter)
+        stream_handler.setLevel(log_level)
+        logger.addHandler(stream_handler)
+
+    if label and outdir:
+        os.makedirs(outdir, exist_ok=True)
+        log_file = os.path.join(outdir, f"{label}.log")
+
+        if log_file not in file_names:
```
dbkeitel, line 119: So this will now add our logfile even if the user preconfigured another?

pyfstat/__init__.py
```diff
@@ -40,3 +41,12 @@
 
 __version__ = get_versions()["version"]
 del get_versions
+
+try:
+    logger = set_up_logger()
+    logger.info(f"Running PyFstat version {__version__}")
```
dbkeitel, line 47: So this differs from #441 in that we always spin up our own in addition, even if the user preconfigured something else already...?

dbkeitel (changes requested):
- [ ] open threads to understand differences of current version from #441 and `append` option
- [x] all examples use the explicit setup call, which is not strictly necessary but good practice to demonstrate
- [x] You've tested yourself that output in tutorial notebooks is as expected?
- [ ] haven't looked at tests output myself yet

pyfstat/loggers.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Parameters
+    ----------
+    outdir: str, optional
+        Path to outdir directory. If `None`, no file handler will be added.
+    label: str, optional
+        Label for this instance of the logger.
+        This is consistent with the rest of the package: ``label'' referes
+        to the string prepended at every file produced by a script.
+        Required, in conjunction with `outdir`, to add a file handler.
```
dbkeitel, line 60:
```suggestion
        Label for the file output handler,
        i.e. the log file will be called `label.log`
        Required, in conjunction with `outdir`, to add a file handler.
```

pyfstat/loggers.py
```diff
@@ -0,0 +1,126 @@
...
+import logging
+import os
+import sys
+
+
+def set_up_logger(
+    outdir=None,
+    label="pyfstat",
+    log_level="INFO",
+    streams=None,
```
dbkeitel, line 47:
```suggestion
    streams=sys.stdout,
```

pyfstat/logging.py
```diff
@@ -0,0 +1,136 @@
...
+    obj:
+        Configured instance of the ``logging.Logger`` class.
+
+    """
+    logger = logging.getLogger("pyfstat")
+    logger.setLevel(log_level)
+
+    if not append:
+        while logger.hasHandlers():
+            logger.removeHandler(logger.handlers[0])
```
Rodrigo-Tenorio, line 96: This one doesn't work, as [hasHandlers](https://github.com/python/cpython/blob/3.10/Lib/logging/__init__.py#L1658) goes up into parent loggers as well, as rightly pointed out in [this post](https://stackoverflow.com/questions/7484454/removing-handlers-from-pythons-logging-loggers).

Fixing by explicitly looping over `logger.handlers`.

tests/test_logging.py
```diff
@@ -0,0 +1,51 @@
...
+    # assert handler.fileBasename == log_file
+
+    # Test append's behaviour
+    logger = set_up_logger()
+    # logger = set_up_logger(label="test_log", outdir=pytest.tmp_path)
+    # assert len(logger.handlers) == 2
+
+    # Check file handler
+
+    # Check stream handler
```
Rodrigo-Tenorio, line 51: Pending as I forgot how to create a temporary directory and I don't trust myself using fixtures on singletons.

docs/source/modules.rst
```diff
@@ -1,14 +1,5 @@
...
-
-PyFstat module documentation
-============================
-
-These pages document the full API and set of classes provided by PyFstat.
-
-See :doc:`here <readme>` for installation instructions and
-other general information.
+PyFstat
+=======
```
dbkeitel, line 2: please reinstate (that's why I suggested to use `git add -p` after `sphinx-apidoc`)

pyfstat/__init__.py
dbkeitel: The `_get_default_logger` logic sounds good, I'll test it a bit but just from reading it sounds like a good replacement for my stuff.

pyfstat/logging.py
```diff
@@ -0,0 +1,136 @@
...
+and *handler* objects. Loggers are the ones in charge of logging,
+hence you call them to emit a logging message with a specific logging
+level (e.g. ``logger.info``); handlers are in charge of redirecting that
+message to a specific place (e.g. a file or your terminal, which is
+usually referred to as a *stream*).
+
+The default behaviour is to attach a ``logging.StreamHandler`` to
+the *pyfstat* logger, printing out to ``sys.stdout`` upon importing the package.
+If, for any reason, ``logging`` cannot access ``sys.stdout`` at import time,
+the exception is reported via ``print`` and no handlers are attached
```
dbkeitel, line 19: Would that actually work - `print` without `stdout`...? :thinking: :slot_machine: 
```
>>> import sys
>>> print("test")
test
>>> sys.stdout=None
>>> print("test")
>>>
```
Well, at least it throws no error, but it also doesn't actually print. ;)

pyfstat/logging.py
```diff
@@ -0,0 +1,136 @@
...
+        root_logger
+        if (root_logger := logging.getLogger()).handlers
+        else set_up_logger()
+    )
+
+
+def set_up_logger(
+    outdir: Optional[str] = None,
+    label: Optional[str] = "pyfstat",
+    log_level: Literal["ERROR", "WARNING", "INFO", "DEBUG"] = "INFO",
```
dbkeitel, line 56: think you forgot one:
```suggestion
    log_level: Literal["CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"] = "INFO",
```

pyfstat/logging.py
```diff
@@ -0,0 +1,136 @@
...
+    this function.
+
+    Parameters
+    ----------
+    outdir:
+        Path to outdir directory. If ``None``, no file handler will be added.
+    label:
+        Label for the file output handler, i.e.
+        the log file will be called `label.log`.
+        Required, in conjunction with ``outdir``, to add a file handler.
```
dbkeitel, line 73: Putting this here as a random example: why the mix of single and double quotes?

pyfstat/logging.py
```diff
@@ -0,0 +1,136 @@
...
+        Ignored otherwise.
+    log_level:
+        Level of logging. This level is imposed on the logger itself and
+        *every single handler* attached to it.
+    streams:
+        Stream to which logging messages will be passed using a
+        StreamHandler object. By default, log to ``sys.stdout``.
+        Other common streams include e.g. ``sys.stderr``.
+    append:
+        If ``False``, removes all handlers from the `pyfstat` logger.
```
dbkeitel, line 83:
```suggestion
        If ``False``, removes all handlers from the `pyfstat` logger
        before adding new ones.
```

...as there seems to be no early return after L96

pyfstat/logging.py
```diff
@@ -0,0 +1,136 @@
...
+        if (root_logger := logging.getLogger()).handlers
+        else set_up_logger()
+    )
+
+
+def set_up_logger(
+    outdir: Optional[str] = None,
+    label: Optional[str] = "pyfstat",
+    log_level: Literal["ERROR", "WARNING", "INFO", "DEBUG"] = "INFO",
+    streams: Optional[Iterable["io.TextIOWrapper"]] = (sys.stdout,),
```
dbkeitel, line 57: Why does the default value need to be a tuple?

setup.py
```diff
@@ -29,6 +29,7 @@
     ],
     "docs": [
         "sphinx==4.3.1",
+        "sphinx_autodoctypehints==1.19.2",
```
dbkeitel, line 32: No napoleon here? It'll get pulled in automatically as a depdency, I guess?

docs/source/conf.py
```diff
@@ -37,7 +37,8 @@
 extensions = [
     "sphinx_rtd_theme",
     "sphinx.ext.autodoc",
-    "sphinx.ext.napoleon",
+    "sphinx.ext.napoleon",  # https://github.com/tox-dev/sphinx-autodoc-typehints/issues/15
+    "sphinx_autodoc_typehints",  # Must import napoleon before typehints
```
dbkeitel, line 41: Is napoleon just needed as a dependency of typehints, or also because you use some different style of docstrings than the existing codebase?

pyfstat/__init__.py
```diff
@@ -40,3 +41,12 @@
 
 __version__ = get_versions()["version"]
 del get_versions
+
+try:
+    logger = _get_default_logger()
+    logger.info(f"Running PyFstat version {__version__}")
+except Exception as e:
```
dbkeitel, line 48:
```suggestion
except Exception as e:  # pragma: no cover
```

dbkeitel:
...a problem, and probably the reason for that documentation admonishment ` you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements`: this seems to stop `pytest -s` or `pytest -rP` from reporting any output from non-failing runs.

...and worse, even for failing tests only the actual error messages are output, no more logging up to that point.

But it might be as easy as actually setting up the logger from each test module...?

dbkeitel:
> But it might be as easy as actually setting up the logger from each test module...?

Yes. A `logger = pyfstat.set_up_logger()` at the top of each `tests/test_x.py` should do it.

docs/source/modules.rst
Rodrigo-Tenorio: :facepalm: I did it a few times correctly. Thanks.

pyfstat/logging.py
Rodrigo-Tenorio: I could remove this and let the logger fail...?

pyfstat/logging.py
Rodrigo-Tenorio: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

Single seems to be for italic emphasis, double is for code. Or so I understood.

pyfstat/logging.py
Rodrigo-Tenorio: This question may have two branches:
- Why tuple rather than list? Because list are mutable and those are problematic as default values and I don't want to fiddle down in the function to force an stdout in certain situations in which stream was set as `None`.
- Why tuple rather than a value? Any iterable would do.

docs/source/conf.py
Rodrigo-Tenorio: [You yourself](https://github.com/PyFstat/PyFstat/blame/master/docs/source/conf.py#L40) introduced Napoleon, and I belive it's because of the way we are formatting the docstrings.

setup.py
Rodrigo-Tenorio: Apparently is not installed, but just [pulled in after installing sphinx](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html)

pyfstat/logging.py
dbkeitel: Guess we did it wrong in a bunch of other places then, but fine here. Yet another needless divergence betwen sphinx/rst and markdown...

pyfstat/logging.py
dbkeitel: ok

setup.py
dbkeitel: I misread the diff, thought you were adding it.

docs/source/conf.py
dbkeitel: I misread the diff, thought you were adding it.

pyfstat/logging.py
dbkeitel: Not sure if that would be better, as the user then wouldn't have a clear way around that error, because they can't access our setup function without importing first. If they really wanted to get rid of their stdout on purpose, and don't have any logging handlers on their root logger either, let them suffer in silence...?

Also this documentation needs to be updated to clarify that now this default is only done if there are no handlers on the root logger yet, or else those are maintained without extras.

setup.py
dbkeitel, line 32: docs build fails with
```
ERROR: Could not find a version that satisfies the requirement sphinx_autodoctypehints==1.19.2 (from pyfstat[docs]) (from versions: none)
ERROR: No matching distribution found for sphinx_autodoctypehints==1.19.2
```

dbkeitel:
> EDIT: Aha, [look at this](https://docs.pytest.org/en/7.1.x/how-to/logging.html).
> 
> I already noticed something weird when setting up `test_logging.py`: Pytest sets up its own handler, hence the pyfstat logger is _not_ imported, as intended by our configuration! @dbkeitel

Ok, so "even for failing tests only the actual error messages are output, no more logging up to that point" is correct default behaviour, and we'd have to configure to be more informative. (Which I'd like.)

But `pytest -s`, `pytest -rP` or `pytest -rX` should still work at info level, or am I misunderstanding that too...?

setup.py
dbkeitel: now
```
ERROR: Cannot install PyFstat and pyfstat[docs]==1.17.0+26.3843e571.clean because these package versions have conflicting dependencies.

The conflict is caused by:
    pyfstat[docs] 1.17.0+26.3843e571.clean depends on sphinx==4.3.1
    sphinx-autodoc-typehints 1.19.2 depends on Sphinx>=5.1.1
```

I don't think there was a breaking change in Sphinx that we needed to pin to `4.3.1` specifically for, the pin is just there because of their suggestion to always pin to a known working configuration. So you can try to update the whole stack.

Rodrigo-Tenorio:
> But pytest -s, pytest -rP or pytest -rX should still work at info level, or am I misunderstanding that too...?

By default they are configured at warning.

pyfstat/logging.py
```diff
@@ -0,0 +1,138 @@
...
+
+Basics of logging: For all our purposes, there are *logger* objects
+and *handler* objects. Loggers are the ones in charge of logging,
+hence you call them to emit a logging message with a specific logging
+level (e.g. ``logger.info``); handlers are in charge of redirecting that
+message to a specific place (e.g. a file or your terminal, which is
+usually referred to as a *stream*).
+
+The default behaviour is to attach a ``logging.StreamHandler`` to
+the *pyfstat* logger, printing out to ``sys.stdout`` upon importing the package.
```
dbkeitel, line 17:
```suggestion
The default behaviour upon importing ``pyfstat`` is to attach a ``logging.StreamHandler`` to
the *pyfstat* logger, printing out to ``sys.stdout``.
This is only done if the root logger has no handlers attached yet;
if it does have at least one handler already,
then we inherit those and do not add any extra handlers by default.
```

setup.py
```diff
@@ -28,12 +28,13 @@
...
     "docs": [
-        "sphinx==4.3.1",
-        "sphinx_rtd_theme==1.0.0",
-        "sphinx_gallery==0.10.1",
-        "m2r2==0.3.1",
-        "mistune<2.0.0",
-        "Pillow==8.4.0",
+        "sphinx==5.1.1",
+        "sphinx_autodoc_typehints==1.19.2",
+        "sphinx_rtd_theme==0.5.1",
```
dbkeitel, line 33: do you know why this needed to be downgraded?

setup.py
```diff
@@ -28,12 +28,13 @@
...
+        "sphinx_gallery==0.11.1",
+        "m2r2==0.3.3",
+        "mistune==0.8.4",
```
dbkeitel, line 36: this one has marched on quite a bit: https://pypi.org/project/mistune/#history

but the "<2.0.0" is equivalent to "==0.8.4.", so ok for this PR

docs/source/modules.rst
dbkeitel, line 2: still/again messed up

docs/source/pyfstat.rst
```diff
@@ -1,5 +1,9 @@
-pyfstat package
-===============
+.. initially generated by
+   sphinx-apidoc -f -o source ../ ../versioneer.py ../setup.py
+   (everything after the first "../" are exclude patterns)
+
+PyFstat module documentation
```
dbkeitel, line 5: Extra copy of the "initially generated" here is fine, but please go back to:
```suggestion
pyfstat package
```

why? compare master
image

against this one
image

dbkeitel (changes requested):

setup.py
dbkeitel: -> #446

docs/source/modules.rst
Rodrigo-Tenorio: Done in 14eaf8a6905c67b3f9797298ea2cc691a57243be

docs/source/pyfstat.rst
Rodrigo-Tenorio: 7ec24ac925a45cd5fd6983ab04675d9397de4a6c

docs/source/modules.rst
```diff
@@ -1,7 +1,6 @@
 .. initially generated by
    sphinx-apidoc -f -o source ../ ../versioneer.py ../setup.py
    (everything after the first "../" are exclude patterns)
-
```
dbkeitel, line 4: this causes
`/home/docs/checkouts/readthedocs.org/user_builds/pyfstat/checkouts/435/docs/source/modules.rst:4: WARNING: Explicit markup ends without a blank line; unexpected unindent`
so please reinstate the blank (can't do via suggestion, apparently)

Rodrigo-Tenorio: Done :+1:

setup.py
dbkeitel: for my own curiosity, removing that one pin:
```
    m2r2 0.3.3 depends on docutils>=0.19
    sphinx 5.1.1 depends on docutils<0.20 and >=0.14
    sphinx-rtd-theme 1.0.0 depends on docutils<0.18
```
so whatever, let's stick with your working set for now

dbkeitel (approved):

PR #441, opened 2022-08-26
logging fallback config

dbkeitel:
This seems rather hacky but implements the following behaviour, which seems most user-friendly to me. But this is very much up for dicussion.

1. user does no logging config at all -> we do everything at info level on stdout with our preferred formatting, but tell them they can override it
2. user calls `pyfstat.set_up_logger()` -> behaviour as on the parent branch
3. user does their own config before importing pyfstat -> we respect it

The 3 cases can be tested with commenting in/out the relevant lines in `PyFstat_example_grid_search_F0.py`.

The main caveat (besides hackiness) is that 3 cares about order, the setup really needs to happen before `import pyfstat`.

### DIFF
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -11,6 +11,7 @@ import logging
 import os
 import shutil
 import subprocess
+import sys
 from datetime import datetime, timezone
 from functools import wraps
 
@@ -83,7 +84,7 @@ def set_up_logger(outdir=None, label="pyfstat", log_level="INFO"):
     )
 
     if any([type(h) == logging.StreamHandler for h in logger.handlers]) is False:
-        stream_handler = logging.StreamHandler()
+        stream_handler = logging.StreamHandler(sys.stdout)
         stream_handler.setFormatter(common_formatter)
         stream_handler.setLevel(level)
         logger.addHandler(stream_handler)
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -51,7 +51,7 @@ def set_up_logger(outdir=None, label="pyfstat", log_level="INFO"):
     Parameters
     ----------
     outdir : str, optional
-        Path of to outdir directory.
+        Path to outdir directory.
     label : str, optional
         Label for this instance of the logger.
         Defaults to `pyfstat`, which is the "root" logger of this package.
@@ -78,8 +78,8 @@ def set_up_logger(outdir=None, label="pyfstat", log_level="INFO"):
     logger.setLevel(level)
 
     common_formatter = logging.Formatter(
-        "%(asctime)s %(name)s %(levelname)-8s: %(message)s",
-        datefmt="%m-%d %H:%M",
+        "%(asctime)s.%(msecs)03d %(name)s %(levelname)-8s: %(message)s",
+        datefmt="%y-%m-%d %H:%M:%S",
     )
 
     if any([type(h) == logging.StreamHandler for h in logger.handlers]) is False:
--- a/pyfstat/helper_functions.py
+++ b/pyfstat/helper_functions.py
@@ -99,9 +99,12 @@ def set_up_logger(outdir=None, label="pyfstat", log_level="INFO"):
             file_handler.setLevel(level)
             logger.addHandler(file_handler)
             # need the following because the file didn't catch any previous version logs yet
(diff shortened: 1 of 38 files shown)


pyfstat/helper_functions.py
```diff
@@ -90,24 +89,23 @@ def set_up_logger(outdir=None, label="pyfstat", log_level="INFO"):
         logger.addHandler(stream_handler)
 
     if any([type(h) == logging.FileHandler for h in logger.handlers]) is False:
-        if label:
-            if outdir:
-                if not os.path.exists(outdir):
-                    os.makedirs(outdir, exist_ok=True)
-            else:
-                outdir = "."
```
dbkeitel, line 98: I've also dropped this `else` case since I'd say we shouldn't spam files into any root directory without the user explicitly asking for it.

pyfstat/helper_functions.py
```diff
@@ -90,24 +89,23 @@ def set_up_logger(outdir=None, label="pyfstat", log_level="INFO"):
...
+            logger.info(f"Additionally logging to file: {log_file}")
             file_handler = logging.FileHandler(log_file)
             file_handler.setFormatter(common_formatter)
-
             file_handler.setLevel(level)
             logger.addHandler(file_handler)
+            # need the following because the file didn't catch any previous version logs yet
+            logger_file_only = logging.Logger("pyfstat")
+            logger_file_only.addHandler(file_handler)
+            logger_file_only.info(f"Running PyFstat version {get_version_string()}")
```
dbkeitel, line 104: This is ugly and I hope it doesn't lead to race conditions, but I couldn't find any documented way to issue a single line to a single handler of an existing logger. Also not sure if I should somehow close this one-off logger after use. (Unfortunately `Logger` can't be used as a context manager.)

pyfstat/helper_functions.py
Rodrigo-Tenorio: I agree :+1:

pyfstat/helper_functions.py
Rodrigo-Tenorio: Why would you want this? The logger itself logs the version to all handlers.

Rodrigo-Tenorio:
> The main caveat (besides hackiness) is that 3 cares about order, the setup really needs to happen before import pyfstat

I wold argue that import-order-dependent behaviour is a hard no-no.

How about adding a root logging line in `pyfstat/__init__.py` saying
```
Importing PyFstat vXYZ. Please, make sure to enable logging using `set_up_logger`.
```
or something in that direction?

pyfstat/helper_functions.py
dbkeitel: Because the logger can't travel back in time. If we're in this situation where there is no file handler assigned yet (`if any([type(h) == logging.FileHandler for h in logger.handlers]) is False`) but we do have `label and outdir`, then the following should have happened:
1. the user didn't pre-configure logging before `import pyfstat`
2. the fallback stdout config from `__init__.py` triggered, printing version info
3. after import, the user called `pyfstat.set_up_logger(label=label, outdir=outdir)` (as we'll want to recommend in the docs)

So at this point we'll be adding the file handler to the existing logger (which already prints to stdout), and stdout already had one round of version info. But the file didn't get it. So if we'd just raise the version info again via `logger.info` at the end of this function (deleted L109 below), it would go to both stdout and file, being duplicate in the stdout. If we did nothing, the file would have no version info. This hack makes sure it's printed once to each of the two handlers.

Alternatively we could just reinstate L109 instead, and don't worry about a duplicate version info on stdout. Or wait... I could just temporarily change levels per handler!

dbkeitel:
> I wold argue that import-order-dependent behaviour is a hard no-no.

It's not really import-order, it's command-vs-import order... but aside from splitting hairs, in my understanding of the `logging` world, some degree of that seems unavoidable. And it's already present in your version: whether or not the caller script does something like a `logging.basicConfig` before `import pyfstat` or not influences the file handler logic.

> How about adding a root logging line

Doesn't that violate the best practices too?

pyfstat/helper_functions.py
dbkeitel: Avoiding the second logger now, with same behaviour. Still hacky but less so.

dbkeitel:
Ok I see this has now diverged from the parent branch, but let's keep discussing before I try rebasing.

PR #442, opened 2022-08-27
update conda recipes

dbkeitel:
pyfstat-dev.yml
- sort dependencies
- add missing attrs

 pyfstat.yml
- bump to python >= 3.8

### DIFF
--- a/etc/pyfstat-dev.yml
+++ b/etc/pyfstat-dev.yml
@@ -4,19 +4,20 @@ channels:
   - conda-forge
 
 dependencies:
-  - python >=3.8
-  - numpy
-  - matplotlib-base >=2.1
-  - scipy
-  - ptemcee
+  - attrs
   - corner
   - dill
-  - tqdm
+  - lalpulsar >=5.0.0
+  - matplotlib-base >=2.1
+  - numpy
   - peakutils
   - pathos
+  - pip
+  - ptemcee
+  - python >=3.8
   - python-lal >=7.1.5
   - python-lalpulsar >=5.0.0
-  - lalpulsar >=5.0.0
-  - pip
+  - scipy
+  - tqdm
   - pip:
     - -e ../[dev]
--- a/etc/pyfstat-dev.yml
+++ b/etc/pyfstat-dev.yml
@@ -21,21 +21,21 @@ dependencies:
   - tqdm
   - pip:
     - -e ../ --install-option="--no-deps"
-    - pre-commit
-    - sphinx==4.3.1
-    - sphinx_rtd_theme==1.0.0
-    - sphinx_gallery==0.10.1
-    - m2r2==0.3.1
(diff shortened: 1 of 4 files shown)


etc/pyfstat-dev.yml
```diff
@@ -4,19 +4,20 @@ channels:
...
+  - ptemcee
+  - python >=3.8
   - python-lal >=7.1.5
   - python-lalpulsar >=5.0.0
-  - lalpulsar >=5.0.0
-  - pip
+  - scipy
+  - tqdm
   - pip:
     - -e ../[dev]
```
dbkeitel, line 23: There's a problem here: this pulls in all requirements from `setup.py`, including e.g. the monolithic PyPI version of `lalsuite`, so we end up with both conda- and pip-installed versions of lalsuite, which is a mess.

Adding `--no-dependencies` would avoid that, but then we won't pull in the `[dev]` dependencies either. I think the safest is to do that, but add the dev stuff manually here, even if that means double the maintenance in case of updates.

dbkeitel: Done that for now, but should really also go and check how many of them can be upgraded to conda dependencies instead of pip.

etc/pyfstat-dev.yml
```diff
@@ -4,19 +4,38 @@ channels:
...
+    - m2r2==0.3.1
+    - mistune<2.0.0
+    - Pillow==8.4.0
+    - pre-commit
+    - pytest
+    - pytest-cov
+    - sphinx==4.3.1
+    - sphinx_gallery==0.10.1
+    - sphinx_rtd_theme==1.0.0
+    - wheel
```
dbkeitel, line 41: migrate from pip -> conda?

setup.py
```diff
@@ -48,7 +48,7 @@
     "test": ["pytest", "pytest-cov", "flaky"],
     "wheel": ["wheel", "check-wheel-contents"],
 }
-for key in ["style", "test", "wheel"]:
+for key in ["docs", "style", "test", "wheel"]:
     extras_require["dev"] += extras_require[key]
```
dbkeitel, line 52: I think we just forgot to add those, but makes sense to me to enable devs to also build docs locally. At least I often do that when their build fails and the online output is less than readable.

etc/pyfstat-dev.yml
dbkeitel: will try @Rodrigo-Tenorio 's suggestion of adding a [dev-nolalsuite] instead

etc/pyfstat-dev.yml
dbkeitel: will no longer apply after new idea

setup.py
dbkeitel: agreed verbally

dbkeitel:
closing and will make a new one

PR #448, opened 2022-08-31
Update recipes v2

dbkeitel:
- supersedes #442 
- pyfstat-dev.yml
  - sort dependencies
  - add missing attrs
- pyfstat.yml
  - bump to python >= 3.8
- setup: skip lalsuite if NO_LALSUITE_FROM_PYPI set
     - use this in pyfstat-dev.yml
       to avoid duplicate packages from conda and pip
- plus some misc README updates

Wiki will be updated to match.

### DIFF
--- a/etc/pyfstat-dev.yml
+++ b/etc/pyfstat-dev.yml
@@ -28,4 +28,4 @@ dependencies:
   - scipy
   - tqdm
   - pip:
-    - -e ../[dev]
+    - -e ../[dev]  # the ".." is needed because pip seems to get called from within "etc"
--- a/etc/pyfstat-dev.yml
+++ b/etc/pyfstat-dev.yml
@@ -1,3 +1,11 @@
+# important notes:
+# 1. to ensure correct relative path for the editable mode installation,
+# always create your environment while at the root directory of your pyfstat git clone.
+#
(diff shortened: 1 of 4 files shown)


etc/pyfstat-dev.yml
```diff
@@ -1,22 +1,31 @@
...
+  - ptemcee
+  - python >=3.8
   - python-lal >=7.1.5
   - python-lalpulsar >=5.0.0
-  - lalpulsar >=5.0.0
-  - pip
+  - scipy
+  - tqdm
   - pip:
     - -e ../[dev]
```
Rodrigo-Tenorio, line 31:
```suggestion
    - -e .[dev]
```
If you run from the root of PyFstat, should be like this.

Rodrigo-Tenorio (changes requested):
Just a minor typo on the environment after re-locating installation to root

etc/pyfstat-dev.yml
dbkeitel: no it's actually needed like this

```suggestion
    - -e ../[dev]  # the ".." is needed because pip seems to get called from within "etc"
```

Rodrigo-Tenorio (approved):

PR #451, opened 2022-09-01
DetectorStates ts nparray

dbkeitel:
@Rodrigo-Tenorio exposes and fixes a missing list->np.array conversion in DetectorStates

For reference, error from running with the first commit only is
```
FAILED tests/test_snr.py::test_DetectorStates - AttributeError: 'list' object has no attribute 'ndim'
```

### DIFF
--- a/tests/test_snr.py
+++ b/tests/test_snr.py
@@ -121,7 +121,7 @@ def test_DetectorStates(data_parameters, writer):
     mds2 = pyfstat.DetectorStates().get_multi_detector_states(
         Tsft=writer.Tsft, timestamps={key: list(common_ts) for key in ("H1", "L1")}
     )
-    assert compare_multi_detector_states_series(*[mds[0], mds2])
+    assert compare_multi_detector_states_series(mds[0], mds2)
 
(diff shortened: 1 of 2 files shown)


tests/test_snr.py
```diff
@@ -117,6 +117,13 @@ def test_DetectorStates(data_parameters, writer):
     ]
     assert compare_multi_detector_states_series(*mds)
 
+    # test again with plain list instead of np.array
+    mds2 = pyfstat.DetectorStates().get_multi_detector_states(
+        Tsft=writer.Tsft, timestamps={key: list(common_ts) for key in ("H1", "L1")}
+    )
+    assert compare_multi_detector_states_series(*[mds[0], mds2])
```
Rodrigo-Tenorio, line 124:
```suggestion
    assert compare_multi_detector_states_series(mds[0], mds2)
```
No unpacking is needed here

Rodrigo-Tenorio (changes requested):
Just that suggestion, otherwise good, thanks.

Rodrigo-Tenorio (approved):

PR #452, opened 2022-09-01
Remove tqdm redirection functions

### LINKED ISSUE #449: tqdm vs logging
Another followup from #435 : it seems the tqdm capture doesn't really work. E.g. if I do
```
logger = pyfstat.set_up_logger(label=label, outdir=outdir)
for handler in logger.handlers:
    if type(handler) == logging.StreamHandler:
        logger.removeHandler(handler)
```
at the start of any grid/mcmc search example, which should then send no logging info to stdout but only to the file, I still get the progress bar on the terminal, and none of it in the file.

On the one hand, a progress bar obviously makes more sense on stdout than in a file, so maybe this is just what tqdm wants to keep doing and can't be persuaded otherwise, which I could live with. On the other hand, it seems like the `logging_redirect_tqdm` is supposed to change this, so it'd be good to check why it doesn't/

Rodrigo-Tenorio:
Closes #449 , as `tqdm` seems to always go on its own towards stdout.

### DIFF
--- a/pyfstat/grid_based_searches.py
+++ b/pyfstat/grid_based_searches.py
@@ -11,7 +11,6 @@ import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
 from tqdm import tqdm
-from tqdm.contrib.logging import logging_redirect_tqdm
 
 import pyfstat.helper_functions as helper_functions
 from pyfstat.core import (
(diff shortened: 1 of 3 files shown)


dbkeitel:
There _probably is_ some way to get this to play nicer, but can't be bothered to try and fix for now. Leave for future.

PR #453, opened 2022-09-01
Split helper_functions into several modules under utils

Rodrigo-Tenorio:
May be too much for a single PR, but let's see how this evolves.


Rodrigo-Tenorio:
- [ ] Restore documentation
- [ ] Use the new function to import matplotlib
- [ ] Test SFT reading
- [ ] Test matplotlib function

dbkeitel:
Had a quick look over the set of submodules so far, and I like it. My only extra suggestion so far would be to put everything PFS-related into a single submodule too, which will then eventually be deprecated and killed wholesale in favour of `snr`.

dbkeitel (changes requested):
Minor fixes, plus can you please post a list of **removed** functions.

dbkeitel (approved):

dbkeitel:
> Minor fixes, plus can you please post a list of **removed** functions.

--> https://github.com/PyFstat/PyFstat/pull/454#pullrequestreview-1096714513

will merge this now, thanks again!

PR #457, opened 2022-09-05
Add basic integrity check to get_sft_as_arrays

Rodrigo-Tenorio:
Just checking that the output has the proper shape and deterministic data.

### DIFF
--- a/pyfstat/utils/sft.py
+++ b/pyfstat/utils/sft.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Optional, Tuple
 
 import lalpulsar
 import numpy as np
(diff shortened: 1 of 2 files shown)


pyfstat/utils/sft.py
```diff
@@ -45,11 +54,11 @@ def get_sft_as_arrays(sftfilepattern, fMin=None, fMax=None, constraints=None):
     sft_catalog = lalpulsar.SFTdataFind(sftfilepattern, constraints)
     ifo_labels = lalpulsar.ListIFOsInCatalog(sft_catalog)
 
-    logger.info(
+    logger.debug(
         f"Loading {sft_catalog.length} SFTs from {', '.join(ifo_labels.data)}..."
     )
```
dbkeitel, line 59: I'd say that's a useful bit at info at info level, and consistent with the typical verbosity level of most lalpulsar tools. Strong reasons to downgrade it, to avoid excessive spamming in some particular workflow scenario?

(Next two downgrades are fine.)

tests/test_utils/test_sft.py
```diff
@@ -0,0 +1,49 @@
...
+    num_frequencies = np.floor(writer_kwargs["Band"] * writer_kwargs["Tsft"] + 0.5)
+    expected_frequencies = (
+        writer_kwargs["F0"]
+        - writer_kwargs["Band"] / 2
+        + frequency_step * np.arange(num_frequencies)
+    )
+
+    # Single detector
+    single_detector_path = writer.sftfilepath.split(";")[0]
+    ifo = "H1" if "H1" in single_detector_path else "L1"
```
dbkeitel, line 38: Why not just `detectors[0]`? Is there something that can break the ordering in the returned path?

tests/test_utils/test_sft.py
```diff
@@ -0,0 +1,49 @@
...
+
+def test_get_sft_as_arrays(tmp_path):
+
+    writer_kwargs = {
+        "sqrtSX": 1,
+        "Tsft": 1800,
+        "timestamps": {
+            "H1": np.array([1, 2, 3, 6, 7, 10, 12]),
+            "L1": np.array([14, 25, 36, 47]),
+        },
```
dbkeitel, line 15: Those are the most overlapping SFTs I've ever seen. 1980 sure was wild times. :wink: A bit surprised that this doesn't trigger any funky errors, but ok. Still, in the sense of "tests are also a bit of an example documentation of how to use the code", maybe put it something more "normal"?

pyfstat/utils/sft.py
Rodrigo-Tenorio: Since we are loading things into dictionaries and all that, I guess such log wouldn't be too much of a burden.

tests/test_utils/test_sft.py
Rodrigo-Tenorio: I hope not, so let's do that

dbkeitel (approved):

PR #470, opened 2022-09-27
Revert walrus operator

Rodrigo-Tenorio:
Python 3.7 is still partially alive (and more so in scientific contexts), so let's try to keep PyFstat 3.7 compliant even though Numpy itself is not anymore.

### DIFF
--- a/pyfstat/logging.py
+++ b/pyfstat/logging.py
@@ -46,11 +46,8 @@ if TYPE_CHECKING:
 
 
 def _get_default_logger() -> logging.Logger:
-    return (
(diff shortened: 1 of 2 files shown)


dbkeitel (commented):
We don't want to and can't go back to supporting 3.7 officially on master, because the lalsuite dependencies aren't there any more. But since this isn't actually making any feature difference, I'm fine with getting rid of it again for now. Then at least people can hack local support for 3.7 easily if they really want to.

dbkeitel (approved):

PR #471, opened 2022-09-27
Revert Python 3.7 into the package

Rodrigo-Tenorio:
To be merged after #470.

@dbkeitel @mj-will there are basically two degrees of freedom to be tweaked:
- Force Python 3.7 or allow for >= 3.7.
- Pin a specific lalsuite and numpy or let pip / conda resolve it.

Thoughts welcome


setup.py
```diff
@@ -35,7 +35,7 @@
     "tqdm",
     "versioneer",
 ]
-lalsuite = "lalsuite>=7.7"
+lalsuite = "lalsuite"
```
Rodrigo-Tenorio, line 38:
```suggestion
lalsuite = "lalsuite>=7.5"
```

PR #473, opened 2022-09-28
Remove Literal typing for now

Rodrigo-Tenorio:
It's not Python 3.7 compliant


dbkeitel (approved):
As for !470: 3.7 _is_ officially dead for our purposes, but as long as there's no real feature difference, I don't care much about the newer fancy bits either.

PR #480, opened 2022-10-20
add (h0,cosi)<->(aplus,across) functions to utils.converting

dbkeitel:
Extracts the existing one from an `snr` internal method to `utils.converting`, and adds the inverse, so they can be reused by other modules too.

Inverse should match https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_compute_fstatistic__v2_8c_source.html#l02045

### DIFF
--- a/pyfstat/utils/converting.py
+++ b/pyfstat/utils/converting.py
@@ -109,7 +109,7 @@ def gps_to_datestr_utc(gps):
     return dt.strftime("%c %Z")
 
 
-def convert_h0_cosi_to_aCross_aPlus(h0, cosi):
+def convert_h0_cosi_to_aPlus_aCross(h0, cosi):
     """
     Converts amplitude parameters from a pair of `(h0,cosi)` to a pair of `(aPlus,aCross)`.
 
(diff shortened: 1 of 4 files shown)


pyfstat/utils/converting.py
```diff
@@ -107,3 +107,62 @@ def gps_to_datestr_utc(gps):
...
+    """
+
+    if np.abs(aCross) > np.abs(aPlus):
+        raise ValueError("not valid for abs(aCross)>abs(aPlus)")
+
+    h0 = aPlus + np.sqrt(aPlus**2 - aCross**2)
+    if h0 > 0:
+        cosi = aCross / h0
+    else:
+        cosi = 0
```
Rodrigo-Tenorio, line 167: @dbkeitel what am I missing here? `h0 > 0` by definition?

```suggestion
    cosi = aCross / h0
```

Rodrigo-Tenorio (changes requested):
Nice @dbkeitel .

I've put some suggestions so this functions can be used with numpy arrays as well.

pyfstat/utils/converting.py
```diff
@@ -107,3 +107,62 @@ def gps_to_datestr_utc(gps):
...
+    Returns
+    -------
+    h0: float
+        Nominal GW amplitude.
+    cosi: float
+        Cosine of the source inclination w.r.t. line of sight.
+    """
+
+    if np.abs(aCross) > np.abs(aPlus):
+        raise ValueError("not valid for abs(aCross)>abs(aPlus)")
```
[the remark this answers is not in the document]
Rodrigo-Tenorio: Right now it fails, we could also decide to return `NaN` on those invalid values, if you prefer.

Thoughs?

pyfstat/utils/converting.py
dbkeitel: No, only `h0 >= 0` is guaranteed. ;)

Turning around the if-else, so the special case is specifically labelled `==0`, would make it more intuitive, but this order is inherited from CFSv2.

pyfstat/utils/converting.py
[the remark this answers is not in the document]
Rodrigo-Tenorio: > ? Seems more intuitive to read to me than the fraction version, and doesn't have a singularity..?

Didn't think to much about it, I actually like yours better :+1:.

pyfstat/utils/converting.py
Rodrigo-Tenorio:
```suggestion
    cosi = np.divide(aCross, h0, out=np.zeros_like(aCross), where=(h0 > 0))
```

Rodrigo-Tenorio (approved):

Issue #481, opened 2022-10-24
SFT-spec v3 compatibility


dbkeitel:
The update of lalpulsar to v3 of the SFT specification will require some updates on our side:
- [x] We will need to put a lalsuite pin in place for the _last release before the SFTspec change_, do our own release with that pin; then do the following adaptations
- [x] At a minimum, `XLALOfficialSFTFilename` will be replaced, and hence its use has to be changed  in `make_sfts.py`.
- [x] While we're at it, since it's used 3 times in the code with the `ifo[0],ifo[1]` idiom, a local wrapper would be nice for cleanup.
- [ ] Will then need to actually start passing window parameters; see also #474
- [ ] Since the new `lalpulsar.SFTFilenameSpec` struct stores narrowband info, we'll either need to populate it properly or (I tend to just do that) slightly update the wildcard postprocessing hack
- [ ] See what else...
- [ ] remove the lalsuite pin and re-release

PR #482, opened 2022-10-24
add wrapper for lalpulsar.OfficialSFTFilename

dbkeitel:
- to make SFTspec v3 migration easier later on
 - refs #481

### DIFF
--- a/pyfstat/utils/sft.py
+++ b/pyfstat/utils/sft.py
@@ -120,3 +120,53 @@ def get_commandline_from_SFTDescriptor(descriptor):
     return next(
         (line for line in comment_lines if "lalpulsar" in line or "lalapps" in line), ""
     )
+
+
+def get_official_sft_filename(
+    IFO, numSFTs, Tsft, tstart, duration, label=None, window_type=None, window_beta=None
+):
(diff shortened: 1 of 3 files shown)


pyfstat/utils/sft.py
```diff
@@ -120,3 +120,53 @@ def get_commandline_from_SFTDescriptor(descriptor):
...
+    Returns
+    -------
+    filename: str
+        The canonical SFT file name for the input parameters.
+    """
+    if window_type or window_beta:
+        raise NotImplementedError(
+            "The parameters 'window_type' and 'window_beta'"
+            " are only included for SFT-spec v3 forwards compatibility"
+            " and not yet implemented."
```
Rodrigo-Tenorio, line 162: Since we'll ignore them, should we put a `warning` here and let it be?

dbkeitel: basically just put it there as a reminder to self to implement them once lalpulsar is updated

Rodrigo-Tenorio: Discussed offline: `NotImplemented` is the right thing, as such a functionality won't be implemented until SFTv3 is out.

Rodrigo-Tenorio (approved):

PR #486, opened 2022-10-25
485 refurbish injection parameters

### LINKED ISSUE #485: Improve `injection_parameters.py` to better handle seeds and configuration files
There are two issues with the current implementation of [injection_parameters](https://github.com/PyFstat/PyFstat/blob/master/pyfstat)/injection_parameters.py

1) Seed handling is kinda problematic whenever a custom function is used. This is *almost always the case*, as things like spindown usually need a custom function to add the extra minus sign or to allow for mixed positive and negative values.

2) Not too much support for configuration files. This is a thing I'm starting to consider as a *must*, but haven't really got into it since there are too many choices. In the end, however, Python seems to incline towards TOML, which will be natively supported in [Python 3.11](https://docs.python.org/3.11/library/tomllib.html). `tomllib` itself is, to my understanding, essentially [tomli](https://pypi.org/project/tomli/) (do *NOT* confuse `tomli` with [toml](https://pypi.org/project/toml/), so we can use that until Python 3.11 is fully supported.

Things to do:
- [ ] Rethink how to specify seeds and distributions. Probably the solution is [this one](https://stackoverflow.com/questions/16016959/scipy-stats-seed) (i.e. move to Scipy stats and let the class attach the random number generator).
- [ ] Find a transparent way of specifying functions, that includes:
    - [ ] A deterministic function being applied to the output of a random number generator.
    - [ ] A random function which combines several sampling procedures into a black box (probably easy as giving `Generator` as an input). 
- [  ] Test that all these can be done from a TOML file with a proper syntax.

Rodrigo-Tenorio:
Closes #485 

Things to do:

- [x] Rethink how to specify seeds and distributions. Probably the solution is [this one](https://stackoverflow.com/questions/16016959/scipy-stats-seed) (i.e. move to Scipy stats and let the class attach the random number generator).
- [x] Keep supporting fixed values.
- [x] Find a transparent way of specifying functions, that includes:
    - [x] A deterministic function being applied to the output of a random number generator.
    - [x] A random function which combines several sampling procedures into a black box (probably easy as giving `Generator` as an input). 
- ~Test that all these can be done from a TOML file with a proper syntax.~ Input is consistent with what the user can specify in a TOML file, so this is all god.
- ~Use structured arrays rather than dicts?~ Left for future implementation. Typing of outputs may be tricky.

### DIFF
--- a/pyfstat/__init__.py
+++ b/pyfstat/__init__.py
@@ -33,6 +33,8 @@ from .gridcorner import gridcorner
 from .injection_parameters import (
     AllSkyInjectionParametersGenerator,
     InjectionParametersGenerator,
+    custom_prior,
+    isotropic_amplitude_priors,
 )
 from .make_sfts import (
     BinaryModulatedWriter,
--- a/pyfstat/__init__.py
+++ b/pyfstat/__init__.py
@@ -33,8 +33,6 @@ from .gridcorner import gridcorner
 from .injection_parameters import (
     AllSkyInjectionParametersGenerator,
     InjectionParametersGenerator,
-    custom_prior,
-    isotropic_amplitude_priors,
 )
 from .make_sfts import (
     BinaryModulatedWriter,
--- a/pyfstat/injection_parameters.py
+++ b/pyfstat/injection_parameters.py
@@ -16,12 +16,12 @@ def custom_prior(prior_function: Callable) -> Callable:
     Intended to be used as a decorator to add custom functions to
     the list of available priors for `InjectionParametersGenerator`.
 
-    For example:
-    ```
-    @pyfstat.custom_prior
-    def negative_log_uniform(generator):
-        return -10**(generator.uniform())
-    ```
+    For example,::
+
+        @pyfstat.custom_prior
+        def negative_log_uniform(generator):
+            return -10**(generator.uniform())
+
     will add the key `negative_log_uniform` to `_pyfstat_custom_priors`
     with said function as the corresponding value.
 
@@ -35,7 +35,6 @@ def custom_prior(prior_function: Callable) -> Callable:
     -------
     prior_function:
         Same function as the input function.
-
     """
 
     function_name = "{prior_function.__name__}"
@@ -90,21 +89,22 @@ class InjectionParametersGenerator:
         Each parameter's prior should be given as a dictionary entry as follows:
         `{"parameter": {"<function>": {**kwargs}}}`
         where <function> may be (exclusively) either a user-defined function decorated
-        with `@custom_prior` or the name of a `scipy.stats` `rv_continuous`.
+        with `@custom_prior` or the name of a `scipy.stats` random variable.
 
-        - | If a user-defined function is used, such a function *must* take
+        * | If a user-defined function is used, such a function *must* take
           | a `generator` kwarg as one of its arguments and use such a generator
           | to generate any required random number within the function.
           | The `generator` kwarg is *required* regardless of whether this is a
           | deterministic or random function.
           | For example, a negative log-distributed random number could be constructed as
-        ```
-        @pyfstat.custom_prior
-        def negative_log_uniform(generator):
-            return -10**(generator.uniform())
-        ```
 
-        - | If a `scipy.stats` function is used, it *must* be given as `stats.*`
+        ::
+
+            @pyfstat.custom_prior
+            def negative_log_uniform(generator):
+                return -10**(generator.uniform())
+
+        * | If a `scipy.stats` function is used, it *must* be given as `stats.*`
           | (i.e. the `stats` namespace should be explicitly included).
           |  For example, a uniform prior between 3 and 5 would be written as
           | `{"parameter": {"stats.uniform": {"loc": 3, "scale": 5}}}`.
--- a/pyfstat/injection_parameters.py
+++ b/pyfstat/injection_parameters.py
@@ -26,6 +26,9 @@ def custom_prior(prior_function: Callable) -> Callable:
     will add the key `negative_log_uniform` to `_pyfstat_custom_priors`
     with said function as the corresponding value.
 
+    See docstring of :func:`~pyfstat.injection_parameters.InjectionParametersGenerator`
+    for an example on how to draw samples from a custom prior.
+
     Parameters
     ----------
     prior_function:
@@ -102,14 +105,28 @@ class InjectionParametersGenerator:
 
         ::
 
-            @pyfstat.custom_prior
+            import pyfstat
+
+            @pyfstat.injection_parameters.custom_prior
             def negative_log_uniform(generator):
                 return -10**(generator.uniform())
+
+            priors = {"my_parameter": "negative_log_uniform": {}}
+            ipg = pyfstat.InjectionParametersGenerator(priors=priors, seed=42)
+            ipg.draw()
 
         * | If a `scipy.stats` function is used, it *must* be given as `stats.*`
           | (i.e. the `stats` namespace should be explicitly included).
           |  For example, a uniform prior between 3 and 5 would be written as
           | `{"parameter": {"stats.uniform": {"loc": 3, "scale": 5}}}`.
+
+        ::
+
+            import pyfstat
+
+            priors = {"my_parameter": "stats.uniform": {"loc": 3, "scale": 5}}
+            ipg = pyfstat.InjectionParametersGenerator(priors=priors, seed=42)
+            ipg.draw()
 
         Delta-priors (i.e. priors for a determinisitic output) can also be specified by
         giving the fixed value to be returned as-is. For example, specifying a fixed
--- a/pyfstat/injection_parameters.py
+++ b/pyfstat/injection_parameters.py
@@ -111,6 +111,10 @@ class InjectionParametersGenerator:
           |  For example, a uniform prior between 3 and 5 would be written as
           | `{"parameter": {"stats.uniform": {"loc": 3, "scale": 5}}}`.
 
+        Delta-priors (i.e. priors for a determinisitic output) can also be specified by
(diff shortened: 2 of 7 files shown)


Rodrigo-Tenorio:
Current state: Looks like `eval`, `globals` and `locals` aren't as friendly as I though.

pyfstat/__init__.py
```diff
@@ -33,6 +33,8 @@
 from .injection_parameters import (
     AllSkyInjectionParametersGenerator,
     InjectionParametersGenerator,
+    custom_prior,
+    isotropic_amplitude_priors,
```
dbkeitel, line 37: Shouldn't be at top namespace level, as they're (currently) _not_ the same as used for MCMC sampling.

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
+          | deterministic or random function.
+          | For example, a negative log-distributed random number could be constructed as
+
+        ::
 
-        2. Dict containing numpy.random distribution as key and kwargs in a dict as value:
-        `{"ParameterA": {"uniform": {"low": 0, "high":1}}}`.
+            @pyfstat.custom_prior
+            def negative_log_uniform(generator):
+                return -10**(generator.uniform())
```
dbkeitel, line 107: should have an explicit example of how to instantiate an object of this class with that same custom prior (best in that same example code box, going all the way to draw()

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
 
-        3. Constant value to be returned as is:
-        `{"ParameterA": 1.0}`.
+        * | If a `scipy.stats` function is used, it *must* be given as `stats.*`
+          | (i.e. the `stats` namespace should be explicitly included).
+          |  For example, a uniform prior between 3 and 5 would be written as
+          | `{"parameter": {"stats.uniform": {"loc": 3, "scale": 5}}}`.
```
dbkeitel, line 112: put as nice box and including the explicit instantiation

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
+
+
 class InjectionParametersGenerator:
     """
     Draw injection parameter samples from priors and return in dictionary format.
 
     Parameters
     ----------
-    priors: dict
+    priors:
```
dbkeitel, line 87: Can be made a bit more beginner-friendly  (more comments to follow).

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
+        on a previous release, are still a valid input. They will be used as a fall-back
+        if none of the two previous options are matched,
+        but their use is highly discouraged for newly produced code.
+
+            1. Callable without required arguments:
+            `{"ParameterA": np.random.uniform}`.
+
+            2. Dict containing numpy.random distribution as key and kwargs
+            in a dict as value:
+            `{"ParameterA": {"uniform": {"low": 0, "high":1}}}`.
```
dbkeitel, line 128: Confusing: sounds like the _syntax_ of that dict of dicts has changed, but actually only the expected contents are now different.

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
+
+    generator:
+        Numpy random number generator (e.g. `np.random.default_rng`)
+        which will be used to draw random numbers from. Conflicts with `seed`.
+
+    seed:
+        Random seed to create instantiate `np.random.default_rng` from scratch.
+        Conflicts with `generator`. If neither `seed` nor `generator` are given,
+        a random number generator will be instantiated using `seed=None`
+        and a warning will be thrown.
```
dbkeitel, line 143: Maybe info suffices, to be reconsidered.

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
+                )
+                self.priors[parameter_name] = parameter_prior
+                continue
+
+            if not isinstance(parameter_prior, Mapping):
+                # If not dictionary, then return as is
+                self.priors[parameter_name] = lambda val=parameter_prior: val
+                continue
+
+            distribution = next(iter(parameter_prior))
```
dbkeitel, line 203: put a check that there are no spurious keys

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
+                    f"with kwargs `{parameter_prior[distribution]}` "
+                    f"to parameter {parameter_name}"
+                )
+                custom_function = _pyfstat_custom_priors[distribution]
+                self.priors[parameter_name] = functools.partial(
+                    custom_function,
+                    generator=self._rng,
+                    **parameter_prior[distribution],
+                )
+            elif "stats." in distribution:
```
dbkeitel, line 217: check

- that it's a string at this point
- that it's not a custom prior of `x.stats.y` format...?

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
+                _, stats_function = distribution.split(".")
```
dbkeitel, line 218: check if there may be n>2 deeper `.` hierarchies in scipy?

pyfstat/injection_parameters.py
```diff
@@ -1,83 +1,256 @@
...
+                logger.warning(
+                    f"Parsing parameter `{parameter_name}` using a deprecated API. "
+                    "This will raise an error in the future."
+                )
+                self.priors[parameter_name] = self._deprecated_prior_parsing(
+                    parameter_prior
+                )
+                continue
+
+    def draw(self) -> dict:
```
dbkeitel, line 248: Would be nice to support `size` argument ->
* need to allow passing it through in parser above
* what should the output be - dict of arrays, or structured array?

dbkeitel (changes requested):

pyfstat/injection_parameters.py
dbkeitel: Structured array should be easier to re-use in many cases (e.g. write to text file), and are easy to convert back to a dict of arrays if needed, so sounds like a good option.
Only downside: need to figure out right syntax for setting the right dtype. (By default numpy would cast everything to floats I think.)

pyfstat/injection_parameters.py
Rodrigo-Tenorio: Point 1) done, point 2) is illegal in python and is not allowed by `custom_priors` either, as discussed offline.

pyfstat/injection_parameters.py
Rodrigo-Tenorio: This code fails if any number of substrings but two is returned by the split.

pyfstat/injection_parameters.py
```diff
@@ -1,121 +1,372 @@
...
+            distribution = next(iter(parameter_prior))
+            if not isinstance(distribution, str):
+                raise ValueError(
+                    "Prior distribution's key should be a string."
+                    f" Got {type(distribution)}`"
+                )
+
+            if distribution in _pyfstat_custom_priors:
+                logger.debug(
+                    f"Setting {distribution} distribution form `_pyfstat_custom_priors` "
```
dbkeitel, line 262:
```suggestion
                    f"Setting {distribution} distribution from `_pyfstat_custom_priors` "
```

pyfstat/injection_parameters.py
```diff
@@ -1,121 +1,372 @@
...
+            Dictionary of arrays. Each key corresponds to that found in ``self.priors``.
+        """
+
+        return {
+            parameter_name: parameter_prior(size=1)[0]
+            for parameter_name, parameter_prior in self.priors.items()
+        }
+
+    def draw_many(self, size) -> dict:
+        """Draw a ``size`` multi-dimensional parameter space point from the given priors.
```
dbkeitel, line 318:
```suggestion
        """Draw ``size`` multi-dimensional parameter space points from the given priors.
```

pyfstat/injection_parameters.py
```diff
@@ -1,121 +1,372 @@
...
+            "Delta": {"uniform_sky_declination": {}},
         }
 
         for key in sky_priors:
-            if key in priors_input_format:
+            if key in priors:
                 logger.warning(
-                    f"Found {key} key in input priors with value {priors_input_format[key]}.\n"
+                    f"Found input key `{key}` with value {priors[key]}.\n"
                     "Overwritting to produce uniform samples across the sky."
```
dbkeitel, line 367:
```suggestion
                    "Overwriting to produce uniform samples across the sky."
```

pyfstat/injection_parameters.py
```diff
@@ -1,121 +1,372 @@
...
 
+_pyfstat_custom_priors = {}
+
+
+def custom_prior(prior_function: Callable) -> Callable:
+    """
+    Intended to be used as a decorator to add custom functions to
+    the list of available priors for ``InjectionParametersGenerator``.
+
+    For example,::
```
dbkeitel, line 21:
```suggestion
    For example::
```

dbkeitel (changes requested):

pyfstat/injection_parameters.py
```diff
@@ -1,121 +1,380 @@
...
+        if (seed is not None) and (generator is not None):
+            raise ValueError(
+                "Incompatible inputs: please, "
+                "use either a `seed` or an already initialized `np.random.Generator`"
+            )
+        if (not seed) and (not generator):
+            logger.warning(
+                f"No `generator` was provided and `seed` was set to {seed}, "
+                "which looks uninitialized. "
+                "Please, make sure you are aware of your seed choice"
```
dbkeitel, line 212:
```suggestion
            logger.info(
                "No `generator` or `seed` was provided."
                f" Will use default `np.random.default_rng({seed})`"
```

dbkeitel (approved):

PR #492, opened 2022-11-02
Writer timestamps handling improvements

dbkeitel:
- check for missing `detectors` if `timestamps` are a string
- allow single-column timestamps files (and warn for Ncol>1)

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -346,6 +346,13 @@ class Writer(BaseSearchClass):
                     " we will ignore the rest."
                 )
                 tsX = tsX[:, 0]
+            if not np.issubdtype(tsX.dtype, np.integer):
+                logger.warning(
+                    "Detected non-integer timestamps in timestamp file."
+                    " Will floor into ints."
+                )
+                tsX = tsX.astype(int)
+
             this_start_time = int(tsX[0])
             this_end_time = int(tsX[-1]) + self.Tsft
             tstart.append(this_start_time)
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -346,12 +346,12 @@ class Writer(BaseSearchClass):
                     " we will ignore the rest."
                 )
                 tsX = tsX[:, 0]
-            if not np.issubdtype(tsX.dtype, np.integer):
+            if not tsX[0].is_integer() or not tsX[-1].is_integer():
                 logger.warning(
                     "Detected non-integer timestamps in timestamp file."
-                    " Will floor into ints."
+                    " We will floor to the nearest integer for the SFT name"
+                    " and let MFDv5 to properly deal with it."
                 )
-                tsX = tsX.astype(int)
 
             this_start_time = int(tsX[0])
             this_end_time = int(tsX[-1]) + self.Tsft
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -349,8 +349,9 @@ class Writer(BaseSearchClass):
             if not tsX[0].is_integer() or not tsX[-1].is_integer():
                 logger.warning(
                     "Detected non-integer timestamps in timestamp file."
-                    " We will floor to the nearest integer for the SFT name"
-                    " and let MFDv5 to properly deal with it."
+                    " We will floor start and end times to the nearest integer"
+                    " for the SFT name,"
+                    " and let lalpulsar_Makefakedata_v5 handle the rest."
                 )
 
             this_start_time = int(tsX[0])
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -188,11 +188,12 @@ class Writer(BaseSearchClass):
             Dictionary of timestamps (each key must refer to a detector),
             list of timestamps (`detectors` should be set),
             or comma-separated list of per-detector timestamps files
-            (simple text files, lines with <seconds> <nanoseconds>,
-            comments should use `%`, each time stamp gives the
-            start time of one SFT).
+            (simple text files,
+            comments should use `%`,
+            the first column is interpreted as SFT start times
+            and additional columns are ignored)
             WARNING: In that last case, order must match that of `detectors`!
-            NOTE: mutually exclusive with [`tstart`,`duration`]
+            NOTE: mutually exclusive with [`tstart`, `duration`]
             and with `noiseSFTs`.
         """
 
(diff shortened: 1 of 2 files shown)


pyfstat/make_sfts.py
```diff
@@ -187,11 +188,12 @@ def __init__(
             Dictionary of timestamps (each key must refer to a detector),
             list of timestamps (`detectors` should be set),
             or comma-separated list of per-detector timestamps files
-            (simple text files, lines with <seconds> <nanoseconds>,
-            comments should use `%`, each time stamp gives the
-            start time of one SFT).
+            (simple text files,
+            comments should use `%`,
```
Rodrigo-Tenorio, line 192:
```suggestion
            comments must use `%`,
```

We explicitly set that in `genfromtxt`. We may want to support `genfromtxt_kwargs` if you think it would be usable.
Probably it should, as it looks like we are moving away from octave?

pyfstat/make_sfts.py
```diff
@@ -334,8 +340,14 @@ def _get_setup_from_timestamps(self):
         self.sftfilenames = []  # This refers to the MFD output!
         for X, IFO in enumerate(IFOs):
             tsX = np.genfromtxt(tsfiles[X], comments="%")
-            this_start_time = int(tsX[0, 0])
-            this_end_time = int(tsX[-1, 0]) + self.Tsft
+            if len(np.shape(tsX)) > 1:
```
Rodrigo-Tenorio, line 343:
```suggestion
            if len(tsX.shape) > 1:
```

pyfstat/make_sfts.py
```diff
@@ -334,8 +340,14 @@ def _get_setup_from_timestamps(self):
...
+                logger.warning(
+                    f"Timestamps file {tsfiles[X]} has more than 1 column,"
+                    " we will ignore the rest."
```
Rodrigo-Tenorio, line 346: Opening discussion:

Most LALSuite timestamp files have two columns ((please correct me if this is just my case). Should this really be a bare warning?

As per what we discussed:
1. Two column files with wanoseconds, ignoring nanoseconds --> Not really a surprise, since CW software does that anyways.
2. Two column files with second column being zeros --> Not really a surprise either.
3. Two column files which are actually *segment* files --> Should crash, that's not how it works.

Q: Is there any simple way of distinguishing 1. and 3.?

Rodrigo-Tenorio (changes requested):
I agree with the approach so far. Final changes dependent upon opened discussion.

pyfstat/make_sfts.py
dbkeitel: > Most LALSuite timestamp files have two columns ((please correct me if this is just my case)

https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___s_f_tfile_i_o__h.html#gaaf70ab5d23c42df41f76e2702cfe1098

-> the seconds,nanoseconds format is actually considered **obsolete** as far as lalpulsar is concerned

That doesn't technically contradict your point though - in more than just this one case, _most_ of what people do is legacy nowadays... :wink:

Regarding the segments format, the only way to identify (most of) them would be by saying that (recent) GPS seconds are one digit longer than what nanosecond remnants should be. Anyway, there are common files of this type with one SFT per line/segment (`ts[n,1]-ts[n,0]=Tsft`), so I wouldn't hard-fail on those. As long as the docs are clear what we do with the files, users who give us weird file formats will have the responsibility to check if they got back the right results.

dbkeitel:
> We may want to support genfromtxt_kwargs if you think it would be usable.

Could, but I'd leave that as a future enhancement request? Maybe together with having a central timestamps parsing util that `DetectorStates` can share?

dbkeitel:
> Could, but I'd leave that as a future enhancement request?

-> #493

dbkeitel:
Left to do: sanity check that the first column is actually `int`

Rodrigo-Tenorio (approved):
Added check for integer timestamps and test case for different warnings.

pyfstat/make_sfts.py
```diff
@@ -334,8 +340,21 @@ def _get_setup_from_timestamps(self):
         self.sftfilenames = []  # This refers to the MFD output!
         for X, IFO in enumerate(IFOs):
             tsX = np.genfromtxt(tsfiles[X], comments="%")
```
Rodrigo-Tenorio, line 342: @dbkeitel this is going to read as a float unless we tell it otherwise.

If we force integer, we will not know what happened.

So... we cannot meaningfully check for float / integer.

Opinion?

tests/test_make_sfts.py
```diff
@@ -19,6 +19,66 @@
...
+    np.savetxt(tmp_path / "dual_column.txt", np.hstack(2 * [single_column[:, None]]))
+
+    writer = pyfstat.Writer(
+        outdir=str(tmp_path / "timestamp_file_testing"),
+        label="timestamp_file_testing",
+        F0=10,
+        Band=0.1,
+        detectors="H1",
+        sqrtSX=1e-23,
+        Tsft=1800,
```
dbkeitel, line 34: should assign TSFT variable before

tests/test_make_sfts.py
```diff
@@ -19,6 +19,66 @@
...
+        timestamps=str(tmp_path / "single_column.txt"),
+    )
+
+    assert single_column[0] == writer.tstart
+    assert single_column[-1] - single_column[0] == writer.duration - 1800
```
dbkeitel, line 39: Tsft

tests/test_make_sfts.py
```diff
@@ -19,6 +19,66 @@
...
+            sqrtSX=1e-23,
+            Tsft=1800,
+            timestamps=str(tmp_path / "dual_column.txt"),
+        )
+
+        _, _, log_message = caplog.record_tuples[-1]
+        assert "has more than 1 column" in log_message
+        assert "will ignore" in log_message
+
+    with pytest.raises(ValueError):
```
dbkeitel, line 75: comment: wrong detectors

tests/test_make_sfts.py
```diff
@@ -19,6 +19,66 @@
...
+            outdir=str(tmp_path / "timestamp_file_testing"),
+            label="timestamp_file_testing",
+            F0=10,
+            Band=0.1,
+            detectors="H1,L1",
+            sqrtSX=1e-23,
+            timestamps=str(tmp_path / "single_column.txt"),
+        )
+
+    with pytest.raises(ValueError):
```
dbkeitel, line 87: comment: no detectors

pyfstat/make_sfts.py
```diff
@@ -334,8 +340,21 @@ def _get_setup_from_timestamps(self):
...
+                logger.warning(
+                    f"Timestamps file {tsfiles[X]} has more than 1 column,"
+                    " we will ignore the rest."
+                )
+                tsX = tsX[:, 0]
+            if not tsX[0].is_integer() or not tsX[-1].is_integer():
+                logger.warning(
+                    "Detected non-integer timestamps in timestamp file."
+                    " We will floor to the nearest integer for the SFT name"
+                    " and let MFDv5 to properly deal with it."
```
dbkeitel, line 353:
```suggestion
                    " We will floor start and end times to the nearest integer"
                    " for the SFT name,"
                    " and let lalpulsar_Makefakedata_v5 handle the rest."
```

Rodrigo-Tenorio (approved):

PR #494, opened 2022-11-04
update Writer timestamps documentation

dbkeitel:
@rodrigo-tenorio here's the documentation update I had forgotten to commit in #492

### DIFF
--- a/pyfstat/make_sfts.py
+++ b/pyfstat/make_sfts.py
@@ -185,13 +185,16 @@ class Writer(BaseSearchClass):
             for reproducibility.
         timestamps: str or dict
             Dictionary of timestamps (each key must refer to a detector),
-            list of timestamps (`detectors` should be set),
+            a single list of timestamps
+            (will be replicated for all detectors; `detectors` must be set),
             or comma-separated list of per-detector timestamps files
             (simple text files,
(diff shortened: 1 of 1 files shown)


pyfstat/make_sfts.py
```diff
@@ -185,13 +185,16 @@ def __init__(
...
             or comma-separated list of per-detector timestamps files
             (simple text files,
             comments must use `%`,
             the first column is interpreted as SFT start times
-            and additional columns are ignored)
-            WARNING: In that last case, order must match that of `detectors`!
+            and additional columns are ignored;
+            `detectors` must be set,
+            and the length and order must match).
+            Timestamps must be integers.
```
Rodrigo-Tenorio, line 197:
```suggestion
            Timestamps must be integers;
            otherwise, will be implicitly cast by this method and MFDv5.
```

Rodrigo-Tenorio: Justification: It's going to work with floats anyways.

Rodrigo-Tenorio (changes requested):
It works with non-integers, so doesn't "Must".

Rodrigo-Tenorio (approved):

PR #495, opened 2022-11-04
pin lalsuite==7.11

dbkeitel:
Lalsuite 7.11 (including lalpulsar 5.2.0) came out today, and we'll likely put in at least one feature before our next release requiring functions added in that release. Next, we expect the backwards-incompatible release (with bump to lalpulsar 6.0.0 and probably (but not sure) to lalsuite 8.0) which we do _not_ want to have our CI or users pull in as upgrades automatically, so we want to pin <= that. (see #481)

We could try `lalsuite>=7.11,<8.0` but since we're not entirely sure of the next version's number, we can just go with `==7.11` for now and, if another backwards-compatible release happens first, rebump in a minor release from our side.

Conda to be done separately, as usual.

### DIFF
--- a/etc/pyfstat-dev.yml
+++ b/etc/pyfstat-dev.yml
@@ -15,7 +15,7 @@ dependencies:
   - attrs
   - corner
   - dill
-  - lalpulsar >=5.0.0
+  - lalpulsar >=5.0.0,<6.0.0
   - matplotlib-base >=2.1
   - numpy
   - pathos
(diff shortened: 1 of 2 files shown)


dbkeitel:
Also did [python-]lalpulsar >=5.2.0,<6.0 in pyfstat-dev.yml - will use that as a test case in advance of the actual conda recipe, which is maintained post-release over at the feedstock.

This branch not to be merged until after we actually make use of the 5.2.0 features, of course. If we end up bumping those to a later release, we can do max-pins only instead.

dbkeitel:
> until after we actually make use of the 5.2.0 features, of course. If we end up bumping those to a later release, we can do max-pins only instead.

Features (specifically XLAL[Copy/Create]MultiNoiseWeights) not used yet, so only added max-pins for upcoming release.

Rodrigo-Tenorio (approved):

PR #496, opened 2022-11-07
add sphinx_autodoc_defaultargs and update to sphinx 5.3.0

dbkeitel:
A feature I've been subconsciously missing for a long time, and only now found out there's a module for it.
Screenshot_20221107_092503

The `rst_prolog` stuff in the config comes from [their "Usage" recommendation](https://github.com/zwang123/sphinx-autodoc-defaultargs/blob/e3e4a123380ca54b7052571d0b2ac2f27fed2af3/README.md#usage) and seems to be called a "substitution rule" in sphinx lingo.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -38,6 +38,7 @@ extensions = [
     "sphinx_rtd_theme",
     "sphinx.ext.autodoc",
     "sphinx.ext.napoleon",  # https://github.com/tox-dev/sphinx-autodoc-typehints/issues/15
+    "sphinx_autodoc_defaultargs",  # should be loaded after napoleon
     "sphinx_autodoc_typehints",  # Must import napoleon before typehints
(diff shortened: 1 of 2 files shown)


Rodrigo-Tenorio:
See the following unexpected side effect: Those red labels weren't there at all before this change.

Off-line communication suggests that maybe we need to use sub-modules.

Screenshot from 2022-11-07 11-51-08

dbkeitel:
> See the following unexpected side effect: Those red labels weren't there at all before this change.

New sphinx 5.2.0 default behaviour. It seems there's no global way to disable it, and in general it's a nice feature. [This](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html) describes how one can disable it per item (class/function/...).

Rodrigo-Tenorio (approved):

PR #498, opened 2022-11-07
officially support python 3.11

dbkeitel:
LALSuite on PyPI is available for it since 7.11 last week, so let's see if we pass.

PR #503, opened 2022-11-21
test_cli.py/Only actually test that INFO messages match

Rodrigo-Tenorio:
Previous version of the test case assumed that every single message would be INFO, which is generally not the case as exemplified in #498 .

This PR fixes that problem by comparing solely on `INFO` lines, which is the main test case of this test.


dbkeitel:
Thanks. I've rebased #498 on this and will merge both if it succeeds. The codecov uploader failed here, but never mind.

dbkeitel (approved):

PR #506, opened 2022-11-21
write_atoms_to_file: change header comments to %%

dbkeitel:
- match lalpulsar.write_MultiFstatAtoms_to_fp columns header line
 - so that e.g. np.genfromtxt() and hence pyfstat.utils.read_txt_file_with_header() work on these files without getting confused by inconsistent comments markers

@itsluanayall did you never run into that problem (numpy.genfromtxt complaining about inconsistent line lengths because it doesn't understand our own header lines having `#` and the one from lalpulsar having `%%`? Or asked the other way round, any scripts of yours that this change would break?

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -1649,7 +1649,7 @@ class ComputeFstat(BaseSearchClass):
             plt.close()
         return ax
 
-    def write_atoms_to_file(self, fnamebase=""):
+    def write_atoms_to_file(self, fnamebase="", comments="%%"):
         """Save F-statistic atoms (time-dependent quantities) for a given parameter-space point.
 
         Parameters
@@ -1659,6 +1659,13 @@ class ComputeFstat(BaseSearchClass):
             `{fnamebase}_Fstatatoms_{dopplerName}.dat`
             where `dopplerName` is a canonical lalpulsar formatting of the
             'Doppler' parameter space point (frequency-evolution parameters).
+        comments: str
+            Comments marker character(s) to be prepended to header lines.
+            Note that the column headers line
+            (last line of the header before the atoms data)
+            is printed by lalpulsar, with `%%` as comments marker,
+            so (different from most other PyFstat functions)
+            the default here is `%%` too.
         """
         multiFatoms = getattr(self.FstatResults, "multiFatoms", None)
         if multiFatoms and multiFatoms[0]:
@@ -1667,7 +1674,7 @@ class ComputeFstat(BaseSearchClass):
             fnameAtoms = f"{fnamebase}_Fstatatoms_{dopplerName}.dat"
             fo = lal.FileOpen(fnameAtoms, "w")
             for hline in self.output_file_header:
-                lal.FilePuts(f"# {hline}\n", fo)
+                lal.FilePuts(f"{comments} {hline}\n", fo)
             lalpulsar.write_MultiFstatAtoms_to_fp(fo, multiFatoms[0])
             del fo  # instead of lal.FileClose() which is not SWIG-exported
         else:
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -624,6 +624,7 @@ class TestComputeFstat(BaseForTestsWithData):
 
 @pytest.fixture
 def CFS_default_params():
+    # FIXME: `default_Writer_params` should be yet another fixture
     return {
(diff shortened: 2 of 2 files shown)


pyfstat/core.py
```diff
@@ -1649,7 +1649,7 @@ def plot_twoF_cumulative(
             plt.close()
         return ax
 
-    def write_atoms_to_file(self, fnamebase=""):
+    def write_atoms_to_file(self, fnamebase="", comments="%%"):
```
Rodrigo-Tenorio, line 1652: - [ ] Check whether empty strings are a problem

Rodrigo-Tenorio (changes requested):

itsluanayall:
Hi,

this change wouldn't break any of my scripts since I am using a fork of PyFstat in which I implemented the option to  `run_output_h5` (alternative to `run`), i.e. put the atoms in hdf5 format instead of text. In this method I don't use `write_atoms_to_file`.

In any case, from this change I suppose the only fix in the scripts one would need to do would be to use `comments="%"` instead of `comments=#` in `numpy.genfromtxt`.

dbkeitel:
Thanks @itsluanayall  and yes scripts reading the output txt files manually should use `comments="%"` after this patch, while they should actually have failed before.

@Rodrigo-Tenorio added a test, with shared fixture together with test_context_finalizer, which can later be expanded to cover more of core. Also uses the `tmp_path` fixture, which we could later use to get rid of most of our clunky test class inheritances. Sanity/style check please.

pyfstat/core.py
dbkeitel: As discussed: they're immutable so they shouldn't.

tests/test_core.py
```diff
@@ -682,6 +692,36 @@ def test_context_finalizer(tCWFstatMapVersion, transientWindowType, cleanup):
...
+    )
+    search.get_fullycoherent_detstat(**lambda_params)
+    atoms_orig = pyfstat.tcw_fstat_map_funcs.reshape_FstatAtomsVector(
+        search.FstatResults.multiFatoms[0].data[0]
+    )
+    atomsdir = tmp_path
+    search.write_atoms_to_file(os.path.join(atomsdir, "CFS"))
+    atomsfiles = [
+        os.path.join(atomsdir, f) for f in os.listdir(atomsdir) if "Fstatatoms" in f
+    ]
```
Rodrigo-Tenorio, line 708:
```suggestion
    atomsdir = str(tmp_path)
    search.write_atoms_to_file(os.path.join(atomsdir, "CFS"))
    atomsfiles = [
        os.path.join(atomsdir, f) for f in os.listdir(atomsdir) if "Fstatatoms" in f
    ]
```
`tmp_path` is a `Path` object, so you may want to cast it into string to avoid a headache.

Rodrigo-Tenorio: Otherwise, fully embrace `Path` and see if it works:
```suggestion
    atomsdir = tmp_path
    search.write_atoms_to_file(atomsdir ( "CFS")
    atomsfiles = [
        (atomsdir / f) for f in os.listdir(atomsdir) if "Fstatatoms" in f
    ]
```

Rodrigo-Tenorio (changes requested):
Separate review: `tmp_path` is not a string.

tests/test_core.py
dbkeitel: It still confuses me that they made this non-portable-looking syntax the portable default, but sure, let's go for it.

Fixed one `(` to `/`.

tests/test_core.py
```diff
@@ -638,14 +634,28 @@ def test_context_finalizer(tCWFstatMapVersion, transientWindowType, cleanup):
         "minCoverFreq": default_signal_params["F0"] - 0.1,
         "maxCoverFreq": default_signal_params["F0"] + 0.1,
     }
-    lambda_params = {
+
+
+@pytest.fixture
+def lambda_params():
```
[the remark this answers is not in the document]
dbkeitel: let's keep that for future test-suite cleanup

tests/test_core.py
Rodrigo-Tenorio: Careful: I'm not saying this will work, I'm saying this *may* work if our code doesn't require too much "stringness" from `Path` :wink:

PR #517, opened 2022-12-20
init_transient_fstat_map_features: upgrade cuda/device info from DEBUG to INFO

dbkeitel:
Lesson recently learned: why would one ever not want to know which GPU one was running on...


Rodrigo-Tenorio (approved):
Agree

PR #518, opened 2022-12-21
SFT specification v3 compatibility

### LINKED ISSUE #481: SFT-spec v3 compatibility

dbkeitel:
work in progress, will fix #481 when ready


dbkeitel:
This now passes both the test suite and examples locally with lalsuite-7.11.1.dev20230209. A proper lalsuite release should happen next week and then I can rerun the CI actions here.

@Rodrigo-Tenorio you could start having a look over the diff in the meantime.

dbkeitel:
Added error checks on Writer labels: now we fail ourselves if they're non-alphanumeric or too long.

dbkeitel:
Readthedocs fails with `Could not find a version that satisfies the requirement lalsuite>=7.13` while all other tests pass, maybe it uses some cached version of PyPI and will only recover later.

ready for review @Rodrigo-Tenorio

pyfstat/core.py
```diff
@@ -520,6 +520,8 @@ def _get_SFTCatalog(self):
             " with constraints {}.".format(self.sftfilepattern, ", ".join(constr_str))
         )
         self.SFTCatalog = lalpulsar.SFTdataFind(self.sftfilepattern, constraints)
+        if self.SFTCatalog.length == 0:
+            raise IOError("No SFTs found.")
```
Rodrigo-Tenorio, line 524: I just checkted [XLALSFTdataFind](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_s_f_tcatalog_8c_source.html#l00068).
 
Is this here because XLALSFTdataFind does not check whether the final set of SFTs is non-empty? (It checks many other things, but not that one I'd say).

pyfstat/make_sfts.py
```diff
@@ -484,6 +509,11 @@ def _basic_setup(self):
         if self.tref is None:
             self.tref = self.tstart
 
+        if getattr(self, "SFTWindowBeta", None):
+            logger.warning(
+                "Option 'SFTWindowBeta' is defunct, please use 'SFTWindowParam'."
+            )
```
Rodrigo-Tenorio, line 515: Should this raise an exception? Right now setting `SFTWindowBeta` does *nothing*, doesn't it?

It should either set `SFTWindowParam` or raise exception.

pyfstat/make_sfts.py
```diff
@@ -770,13 +800,7 @@ def _build_MFD_command_line(self):
         cl_mfd.append('--outSFTdir="{}"'.format(self.outdir))
         cl_mfd.append('--outLabel="{}"'.format(self.label))
 
-        if self.noiseSFTs is not None and self.SFTWindowType is None:
-            raise ValueError(
-                "SFTWindowType is required when using noiseSFTs. "
-                "Please, make sure you understand the window function used "
-                "to produce noiseSFTs."
-            )
-        elif self.noiseSFTs is not None:
```
Rodrigo-Tenorio, line 779: This is removed because MFD does it for you, correct?

pyfstat/make_sfts.py
```diff
@@ -99,7 +99,8 @@ def __init__(
         sqrtSX=None,
         noiseSFTs=None,
         SFTWindowType=None,
-        SFTWindowBeta=0.0,
+        SFTWindowParam=0.0,
```
Rodrigo-Tenorio, line 102: Is this consistent with LALPulsar's default?

pyfstat/utils/sft.py
```diff
@@ -140,31 +147,30 @@ def get_official_sft_filename(
...
+        window_type=window_type,
+        privMisc=label,
+        pubObsKind=None,
+        pubChannel=None,
     )
+    spec.window_param = window_param or 0
+    spec.numSFTs = numSFTs
+    spec.SFTtimebase = Tsft
+    spec.gpsStart = tstart
+    # possible gotcha: duration may be different if nanoseconds of sft-epochs are non-zero
```
Rodrigo-Tenorio, line 174: Should we set a warning? LALPulsar codes don't work with nanoseconds in any case.

Rodrigo-Tenorio (changes requested):
I agree with the changes, just some minor comments so things are documented and a question.

pyfstat/core.py
dbkeitel: Yes, during testing I ran into some case where it was throwing a rather ugly error further down without this check.

pyfstat/make_sfts.py
dbkeitel: Good point, I think it might fail with a lalpulsar error further on and the warning is just for clarity, but I'll have to check to remind myself of what actually happens.

pyfstat/make_sfts.py
dbkeitel: No, because this strict requirement actually no longer exists. That was part of the v3 changes: the new SFT headers actually contain sufficient info to reconstruct the window and reuse it at injection time. MFD will only throw an error if a) the noiseSFTs are old ones without header window info / have an explicitly "unknown" window in the header or b) the user asks for a window that is mismatched with the one stored in the noiseSFTs header.

https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/2082

pyfstat/utils/sft.py
dbkeitel: I don't quite remember why I put this here; these are indeed UINT4 fields in https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/struct_s_f_t_filename_spec.html#a2462122a0833c0d5d2084dc543d0c1bf not GPS structs, so it should just fail when given a float...?

pyfstat/make_sfts.py
dbkeitel: Indeed it was not failing in any obvious way when passed the old name. Changed to an error in b0cd43b07e99d8ba5239770a60a1c162eb52207f

pyfstat/utils/sft.py
dbkeitel: Indeed the SFT spec struct has UINT4 fields, but the actual SFTs have full GPS info theoreticlaly allowing for nanoseconds. Comments was echoed from https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_s_f_tfile_i_o_8c_source.html#l00675 . Extended to refer to it in ac1f5a9 .

pyfstat/make_sfts.py
dbkeitel: This was a good catch indeed: MFD throws `Inconsistent Window-option: 'tukey' requires parameter = yes, but user supplied parameter = no` if trying to rely on the default (i.e. it's detecting "user did not make any conscious choice"). But as I had it implemented, we produced `window:      tukey(0)`.

Unfortunately going with `None` as our default triggers a rather ugly lal error `XLAL Error - XLALParseStringValueAsREAL8 (UserInputParse.c:168): strtod(): trailing garbage 'None' found after double-conversion of 'None'` so I now put in an explicit check via [XLALCheckNamedWindow](https://lscsoft.docs.ligo.org/lalsuite/lal/group___window__h.html#gaac5218a54390c9f427f8d359b598e9ec).

(954ddb8)

Rodrigo-Tenorio (approved):

dbkeitel:
@Rodrigo-Tenorio readthedocs was failing, so I updated its config and that seems to help. Not sure why: python 3.8 is definitely still supported by lalsuite 7.13 (I reproduced its environment on CIT and no problem), so maybe their default OS container had gotten terminally outdated, or changed to Windows or whatever.

Anyway, re-review last commit please, and then this should be finally ready to merge.

pyfstat/make_sfts.py
```diff
@@ -484,6 +510,22 @@ def _basic_setup(self):
...
+        if getattr(self, "SFTWindowType", None):
+            try:
+                lal.CheckNamedWindow(
+                    self.SFTWindowType, self.SFTWindowParam is not None
+                )
+            except RuntimeError:
+                raise ValueError(
+                    "XLAL error on checking SFT window options."
+                    f" Likely either SFTWindowType={self.SFTWindowType} is not a recognised window name,"
+                    " or it requires also setting a SFTWindowParam."
```
Rodrigo-Tenorio, line 526:
```suggestion
                    " or it requires also setting an SFTWindowParam."
```

Rodrigo-Tenorio (dismissed):
Just a (debatable) typo. Good to go otherwise

Rodrigo-Tenorio (approved):

PR #521, opened 2023-01-31
change lalsuite pin to <7.12

dbkeitel:
-  because current dev versions are 7.11.1.dev*
 -  so dev test jobs (e.g. [here](https://github.com/PyFstat/PyFstat/actions/runs/4055238044)) still only pull in the 7.11 release

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ requires = [
     "tqdm",
     "versioneer",
 ]
-lalsuite = "lalsuite>=7.7,<=7.11"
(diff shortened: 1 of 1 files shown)


dbkeitel:
[New test run](https://github.com/PyFstat/PyFstat/actions/runs/4055481154) now "fails the right way" by pulling in `lalsuite-7.11.1.dev20230131` and then failing due to the SFT filename function change.

dbkeitel:
Dropping this and just going for a v3 compatibility PR directly.

PR #529, opened 2023-02-23
PyCuda transient-Fstat kernels: add fallback for ill-conditioned antenna pattern matrix, and unit test

### LINKED ISSUE #527: pycuda F-stat maps: add condition number checks?
LALSuite has had a check on the condition number of the antenna pattern matrix since 2018 ([code link](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/group___l_a_l_compute_a_m__h.html#gad7ba2b601622946ba930216477c832b4); bonus historical reference, though probably not public-readable: https://git.ligo.org/CW/software/lalsuite/-/issues/13 ) and returns a default value of 2F=4 if that check fails. My pycuda implementation does not have that check and hence can in rare cases (especially for very short-duration single-detector calculations, but even then only for specific input parameter combinations) produce larger values than expected.

dbkeitel:
This adds the same antenna pattern matrix condition number check as in LALComputeAntennaPatternSqrtDeterminant() and estimateAntennaPatternConditionNumber(),  see https://git.ligo.org/CW/software/lalsuite/-/issues/13 and https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/122/  for the original LALSuite discussion. For now it's hardcoded to lalpulsar's standard 1e4 threshold and a fallback of 2F=4. It improves consistency between lalpulsar and pycuda results in rare corner cases, particularly when a transient window includes very few SFTs (typically 2) from a single detector, and even then it would only produce noticeable outliers for relatively rare Doppler parameter combinations.

While I revisited these files for the first time in years, I also added a basic test case that synths some atoms and recomputes them. If pycuda is not available, tests for that implementation are skipped, so e.g. in the CI the second half of the test shouldn't do anything. This skipping likely isn't entirely robust, e.g. when pycuda is available but no GPU can actually be accessed, I expect it will fail instead. But I don't think that's worth trying to work around for now, unless it turns out to cause problems for any developer. The test does also _not_ explicitly target the ill-conditioned situation, but we tested the fix separately on a set of parameters where spurious outliers had occurred.

cc @itsluanayall @Rodrigo-Tenorio 

closes #527

### DIFF
--- a/pyfstat/tcw_fstat_map_funcs.py
+++ b/pyfstat/tcw_fstat_map_funcs.py
@@ -266,6 +266,16 @@ def _optional_imports_pycuda():
     )
 
 
+def _get_transient_fstat_map_features():
+    """Helper function to check available features."""
+    features = {}
+    have_lal = _optional_import("lal")
+    have_lalpulsar = _optional_import("lalpulsar")
+    features["lal"] = have_lal and have_lalpulsar
+    features["pycuda"] = _optional_imports_pycuda()
+    return features
+
+
 def init_transient_fstat_map_features(feature="lal", cudaDeviceName=None):
     """Initialization of available modules (or 'features') for computing transient F-stat maps.
 
@@ -296,11 +306,7 @@ def init_transient_fstat_map_features(feature="lal", cudaDeviceName=None):
         A CUDA device context object, if assigned.
     """
 
-    features = {}
-    have_lal = _optional_import("lal")
-    have_lalpulsar = _optional_import("lalpulsar")
-    features["lal"] = have_lal and have_lalpulsar
-    features["pycuda"] = _optional_imports_pycuda()
+    features = _get_transient_fstat_map_features()
     logger.debug("Got the following features for transient F-stat maps:")
     logger.debug(features)
 
--- a/pyfstat/pyCUDAkernels/cudaTransientFstatExpWindow.cu
+++ b/pyfstat/pyCUDAkernels/cudaTransientFstatExpWindow.cu
@@ -101,14 +101,19 @@ __global__ void cudaTransientFstatExpWindow ( float *input,
 
     }
 
-    /* get determinant */
-    float Dd = ( Ad * Bd - Cd * Cd );
-    float DdInv = 0.0f;
-    /* safety catch as in XLALWeightMultiAMCoeffs():
-     * make it so that in the end F=0 instead of -nan
+    /* get inverse antenna pattern determinant,
+     * following safety checks from
+     * XLALComputeAntennaPatternSqrtDeterminant()
+     * and estimateAntennaPatternConditionNumber()
      */
-    if ( Dd > 0.0 ) {
-      DdInv  = 1.0 / Dd;
+    float sumAB  = Ad + Bd;
+    float diffAB = Ad - Bd;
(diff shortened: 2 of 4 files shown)


dbkeitel:
PS: full test with pycuda parts passes on ldas-pcdev8. On Artemisa, as expected a plain `pytest tests/test_tcw_fstat_map_funcs.py` fails, but with `gpurun` it passes.

pyfstat/tcw_fstat_map_funcs.py
```diff
@@ -266,6 +266,16 @@ def _optional_imports_pycuda():
     )
 
 
+def _get_transient_fstat_map_features():
+    """Helper function to check available features."""
+    features = {}
+    have_lal = _optional_import("lal")
+    have_lalpulsar = _optional_import("lalpulsar")
+    features["lal"] = have_lal and have_lalpulsar
```
Rodrigo-Tenorio, line 274: For the record, I confirm that `_optional_import` returns a `bool` and hence `features["lal"]` will *always* be a boolean as constructed here.

pyfstat/tcw_fstat_map_funcs.py
```diff
@@ -266,6 +266,16 @@ def _optional_imports_pycuda():
...
+    features["pycuda"] = _optional_imports_pycuda()
```
Rodrigo-Tenorio, line 275: `_optional_imports_pycuda` is a conjunction of `bool` values provided by `_optional_import`, hence the return value itself is *always* a `bool` as constructed here.

Rodrigo-Tenorio:
Ran the test case on my local machine:
```
(pyfstat-dev) rodrigo@pop-os:~/gravity/git/PyFstat$ pytest -k fstat_map_funcs
==================================== test session starts =====================================
platform linux -- Python 3.10.6, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/rodrigo/gravity/git/PyFstat, configfile: pytest.ini
plugins: flaky-3.7.0, anyio-3.6.2, cov-4.0.0, nbmake-1.3.4
collected 156 items / 148 deselected / 8 selected                                            

tests/test_tcw_fstat_map_funcs.py::test_compute_transient_fstat_map[lal-rect-0] PASSED [ 12%]
tests/test_tcw_fstat_map_funcs.py::test_compute_transient_fstat_map[lal-rect-10] PASSED [ 25%]
tests/test_tcw_fstat_map_funcs.py::test_compute_transient_fstat_map[lal-exp-0] PASSED  [ 37%]
tests/test_tcw_fstat_map_funcs.py::test_compute_transient_fstat_map[lal-exp-10] PASSED [ 50%]
tests/test_tcw_fstat_map_funcs.py::test_compute_transient_fstat_map[pycuda-rect-0] PASSED [ 62%]
tests/test_tcw_fstat_map_funcs.py::test_compute_transient_fstat_map[pycuda-rect-10] PASSED [ 75%]
tests/test_tcw_fstat_map_funcs.py::test_compute_transient_fstat_map[pycuda-exp-0] PASSED [ 87%]
tests/test_tcw_fstat_map_funcs.py::test_compute_transient_fstat_map[pycuda-exp-10] PASSED [100%]

============================= 8 passed, 148 deselected in 5.45s ==============================

```
:+1:

Rodrigo-Tenorio (approved):
Makes sense

PR #537, opened 2023-05-29
use lalsuite[lalpulsar] to pull in solar-system-ephemerides

### LINKED ISSUE #536: need to update lalsuite dependency for ephemerides
Installing from pip no longer gives the needed ephemerides files, as seen [here](https://github.com/PyFstat/PyFstat/actions/runs/5110635261/jobs/9186742099). This is because after https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/2156/diffs  we need to use `lalsuite[lalpulsar]` to pull in the new external dependency on https://pypi.org/project/solar-system-ephemerides/

Conda behaviour is so far not changed (https://anaconda.org/conda-forge/lalpulsar still contains the files), see the discussion on https://git.ligo.org/lscsoft/lalsuite/-/issues/675

dbkeitel:
- closes #536
 - LALSuite-dev test run: https://github.com/PyFstat/PyFstat/actions/runs/5112834724 (using `pip install --pre` for nightly lalsuite)

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,11 @@ NOTE: We require a recent LALSuite (>=7.13) / lalpulsar (>=6.0).
 If you need to work with older versions,
 the last PyFstat release supporting those was `1.19.1`.
 
-NOTE: Through pip, we install the dependency `lalsuite[lalpulsar]`
+NOTE: Through pip, we install the dependency `lalsuite[lalpulsar]`.
+The optional flag `[lalpulsar]` is ignored for `lalsuite <=  7.15`,
+as those releases carry their own ephemerides.
+For `lalsuite > 7.15`, `[lalpulsar]` installs `solar-system-ephemerides`
+and uses this auxiliary package to access the ephemerides.
 
 ### Optional dependencies
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,9 @@ A simple
 pip install pyfstat
 ```
 should give you the latest release version with all dependencies;
-recent releases now also include a sufficient minimal set of ephemerides files.
+recent releases now also include
+(or pull in, via dependency)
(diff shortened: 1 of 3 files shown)


dbkeitel:
Looks like this works @Rodrigo-Tenorio , did I forgot any special files?

I'll also update https://github.com/PyFstat/PyFstat/wiki/ephemerides-installation to explain the slight version chaos that we'll have for a transitionary period, but didn't feel like polluting the README with it.

README.md
```diff
@@ -174,6 +176,8 @@ NOTE: We require a recent LALSuite (>=7.13) / lalpulsar (>=6.0).
 If you need to work with older versions,
 the last PyFstat release supporting those was `1.19.1`.
 
+NOTE: Through pip, we install the dependency `lalsuite[lalpulsar]`
+
```
Rodrigo-Tenorio, line 180: Can we make this a bit more verbatim, since it's already trying to clarify a quite specific situation?
```suggestion
NOTE: Through pip, we install the dependency `lalsuite[lalpulsar]`. The optional flag `[lalpulsar]` is ignored for `lalsuite <=  7.15`, as those releases carry their own ephemerides. For `lalsuite > 17.5`, `[lalpulsar]` installs `solar-system-ephemerides` and uses this auxiliary package to access the ephemerides.

```

Rodrigo-Tenorio (changes requested):
@dbkeitel 

Just a suggestion to make it more verbatim.

README.md
dbkeitel: That's the level of detail I had thought of leaving to the wiki page, but no objection either. Fixed your suggestion above for version number typo and line breaks, committing now.

Rodrigo-Tenorio (approved):

PR #543, opened 2023-10-11
adapt integration.yml to using OIDC for PyPI

dbkeitel:
- follows new https://docs.pypi.org/trusted-publishers/ model
- split into separate jobs build, testpypi-publish, pypi-publish
- use environments
- use artifact upload/download


dbkeitel:
Turns out I can't meaningfully test the actual upload from a branch, need a tag, so once standard integration passes again, I'll merge and test with "rc" tags from master.

PR #547, opened 2024-06-28
Black 24.4.2

dbkeitel:
Set it as the version of reference for pre-commit, and make it happy with minor 2024 style changes.


dbkeitel:
This fixed the integration test. :tada: 

Codecov upload failed, will look into updating that action separately.

Issue #335, opened 2021-10-22
gridfile-based search


dbkeitel:
There are several use cases for reading grid files as a search setup:

- using CFSv2 or Weave to set up fancy grids without us having to directly wrap all that functionality
- as mentioned on #333: off-sourcing or any other re-running over MCMC output samples

Not high on the priority list for the moment, but would be "nice to have".

For future reference, I've had a quick look at [XLALLoadFullGridFile()](https://lscsoft.docs.ligo.org/lalsuite/lalpulsar/_doppler_full_scan_8c.html#a9eb2eb11ced883c03fa3b0d811a85f08) as called by CFSv2, but this returns a `DopplerFullScanState` which we don't use in our main search setup (only in `estimate_min_max_CoverFreq()`), and it also doesn't understand binary parameters.

So it'll be better to have our own simple wrapper, using pandas or whatever to read the file and then do a simple loop. (Not aiming for total optimisation, just getting something simple started, in keeping with the rest of the grid module.) That still leaves two sub-options:
1. adapt `GridSearch` to do this: would have to make `F0s` etc optional and fill them from the grid file, add binary options (-> #181)
2. add new `SearchOverGridFile` or similarly named class, inhereting either from `GridSearch` or base `ComputeFStat`.

To me 2 sounds cleaner.

Optional add-on feature that would make this super handy: read the other relevant parameters (SFTs, refTime, ...) from the file header (CFSv2's or PyFstat's).

PR #561, opened 2024-10-07
Allow core F-stat functions to take a dict for parameters

dbkeitel:
Some 2-year old edits I had lying around locally, rebased onto current master. This is a prerequisite for improvements such as #335.

`get_fullycoherent_detstat` and friends now take an optional `params` dictionary argument that can replace the old-style individual arguments for F0,... and the old style is already labeled as DEPRECATED and might be removed in the future. This will allow for including higher spindowns and non-standard parameters.

Along the way, I've also centralised the internal defaults for LaTeX labels into the core class.

### DIFF
--- a/pyfstat/core.py
+++ b/pyfstat/core.py
@@ -124,7 +124,7 @@ class BaseSearchClass:
         matches = [item for sublist in matches for item in sublist]
         if len(matches) > 0:
             return matches
-        else:
+        else:  # pragma: no cover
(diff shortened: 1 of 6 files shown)


dbkeitel:
After fixing a few more spurious cases, both the test suite and also the examples pass: https://github.com/PyFstat/PyFstat/actions/runs/11218806956

Users should not be facing a change, so I'll merge this already.

PR #576, opened 2025-01-28
Gridsearch BtSG support and logging

dbkeitel:
Fixes the missing `BtSG` support in `SearchOverGridFile` and improves logging.

- [x] @maferrerma Can you add an extra test that runs this class with transient options and that `BtSG` option, then checks that the output file actually contains the column headers `maxTwoF` and `lnBtSG`?


dbkeitel:
History cleaned up, merging now. Thanks for the test!

PR #577, opened 2025-02-18
Implementation of t0_MP and tau_MP computation for transient searches

maferrerma:
If BtSG is set to True, t0_MP and tau_MP are also returned in the 

Consistent with `XLALFindModeOfPDF1D()`, we return the center-value of the discrete ' containing the maximum of the distribution.


dbkeitel (approved):
This now looks consistent in itself. There are some follow-up issues with BSGL/single-F that I noticed when re-running examples with this, which I'll fix separately after merging, and there's some follow-up debugging and improving to be done for MCMCTransientSearch.

PR #579, opened 2025-02-21
Fix single IFO issues

dbkeitel:
This fixes several issues related to single-IFO F-stat outputs:

1. add explicit `singleFstats` options to the various grid based searches, in case a user wants twoFX, but not BSGL
2. fix some internal assignments of `twoFX` that were still coded for list, not dict style (forgotten in https://github.com/PyFstat/PyFstat/pull/561)
3. make sure these values are obtained and reported independent of transient options
4. ensure consistent internal ordering of detector names and hence single-IFO F-stat values, by using the sorted np.unique() rather than the unordered set() - so far, sometimes randomly H1<->L1 for example could get swapped for the `twoFX` output


dbkeitel:
succesful examples run: https://github.com/PyFstat/PyFstat/actions/runs/13461281519

maferrerma (approved):

PR #581, opened 2025-02-24
upgrade setuptools in integration.yml

dbkeitel:
- we need setuptools>=69.3.0 to support PEP 625
- to avoid PyPI deprecation warnings about sdist filename

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -24,7 +24,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install package and development dependencies
       run: |
-        python -m pip install --upgrade pip
+        # we might still need a newer pip to be able to pull in lalsuite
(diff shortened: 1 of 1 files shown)


dbkeitel:
PyPI had been complaining:

> This email is notifying you of an upcoming deprecation that we have determined may affect you as a result of your recent upload to 'PyFstat'.

> In the future, PyPI will require all newly uploaded source distribution filenames to comply with PEP 625 (https://peps.python.org/pep-0625/). Any source distributions already uploaded will remain in place as-is and do not need to be updated.

> Specifically, your recent upload of 'PyFstat-2.2.1.tar.gz' is incompatible with PEP 625 because it does not contain the normalized project name 'pyfstat'.

> In most cases, this can be resolved by upgrading the version of your build tooling to a later version that supports PEP 625 and produces compliant filenames. You do not need to remove the file.

With a more modern setuptools, this gives something complying: 'pyfstat-0+untagged.1.375a82e6.clean'

PR #582, opened 2025-05-13
unpin numpy (allow >=2)

dbkeitel:
I think the pin only remained in place after https://github.com/PyFstat/PyFstat/commit/6571123804f821c6a61d2bdd07607a8e06f80936 because of lalsuite (or maybe some other) dependencies, and since those seem to have all fixed their compatibility, I'm not getting any test failures unpinning.


dbkeitel:
examples run: https://github.com/PyFstat/PyFstat/actions/runs/15002882527

PR #584, opened 2025-05-28
enable python 3.13 support

dbkeitel:
Lalsuite has it now, so we can as well. No errors found in local testing.

Also remove the explicit `versioneer` dependence from `setup.py`. The local copy should be enough ("vendored mode") as far as I understand, and at least local testing was happy with it that way. (It also has never been a dependency in the conda recipes, see https://github.com/conda-forge/pyfstat-feedstock/blob/main/recipe/meta.yaml and discussion at https://github.com/conda-forge/pyfstat-feedstock/pull/34 ,)

### DIFF
--- a/.github/workflows/check_notebooks.yml
+++ b/.github/workflows/check_notebooks.yml
@@ -13,7 +13,7 @@ jobs:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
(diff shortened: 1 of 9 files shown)


dbkeitel:
Integration tests passed. I left the "run examples" github integration at 3.10 so it keeps testing that we don't use any "too new" features. (3.9 is going to be deprecated soon, so 3.10 is a good choice for now.) So I'm just doing a 3.13 all-examples run locally and will merge if that comes out clean.

PR #587, opened 2026-01-16
Solve overflow problem in plot_spectrogram.

maferrerma:
Spectogram example was giving this warning:
```
/home/maferrerma/miniforge3/envs/pyfstat/lib/python3.13/site-packages/pyfstat/utils/sft.py:365: RuntimeWarning: overflow encountered in cast
  q *= 2 / (Tsft * sqrtSX**2)
/home/maferrerma/miniforge3/envs/pyfstat/lib/python3.13/site-packages/pyfstat/utils/sft.py:365: RuntimeWarning: invalid value encountered in multiply
  q *= 2 / (Tsft * sqrtSX**2)
```
and giving a blank spectogram in some environments.

PR #588, opened 2026-01-16
more robust fix for overflows in plot_spectrogram()

dbkeitel:
@maferrerma https://github.com/PyFstat/PyFstat/pull/587 didn't work for me. Please check if this one still works for you.


maferrerma (approved):
This fix also works in my environment.

PR #604, opened 2026-08-04
Including glitch F1 for the cumulative 2F plot

martacarriothomas:
The F1 parameter was lost before so the plot didn't look complete. I fixed it.

### DIFF
--- a/pyfstat/mcmc_based_searches.py
+++ b/pyfstat/mcmc_based_searches.py
@@ -2686,12 +2686,20 @@ class MCMCGlitchSearch(MCMCSearch):
             delta_F0s.insert(self.theta0_idx, 0)
             delta_F0s = np.array(delta_F0s)
             delta_F0s[: self.theta0_idx] *= -1
+            delta_F1s = [d["delta_F1_{}".format(i)] for i in range(self.nglitch)]
+            delta_F1s.insert(self.theta0_idx, 0)
+            delta_F1s = np.array(delta_F1s)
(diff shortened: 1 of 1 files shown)


dbkeitel:
Thanks for your first PR @martacarriothomas !

The `codecov` check is failing because `MCMCGlitchSearch` isn't properly tested at all; this will need to be fixed later.

The `docs` check is also failing already for a while more generally, with the fix still pending.

I'll merge anyway, based on quick offline testing. Here are the cumulative plots from rerunning the standard  https://github.com/PyFstat/PyFstat/tree/master/examples/glitch_examples on this branch:

1. non-glitch-aware `MCMCSearch`:
PyFstatExampleStandardDirectedMCMCSearchOn1Glitch_twoFcumulative

2. `MCMCGlitchSearch` :
PyFstatExampleGlitchRobustDirectedMCMCSearchOn1Glitch_twoFcumulative

3. And with an actual `delta_F1 = -1e-11` included in the injection, but not in the prior:
PyFstatExampleGlitchRobustDirectedMCMCSearchOn1Glitch_twoFcumulative

4. And with prior `"F1": {"type": "unif", "lower": F1 - F1_width / 2.0, "upper": F1 + F1_width / 2.0}` it now works correctly:
PyFstatExampleGlitchRobustDirectedMCMCSearchOn1Glitch_twoFcumulative

dbkeitel (approved):

